PR #11, opened 2018-09-05
Add `prefixUrl` option

### LINKED ISSUE #1: Add `prefixUrl` option
Like in Got: https://github.com/sindresorhus/got#baseurl

sholladay:
Fixes #1

```js
ky('world', {
    baseUrl : 'https://mysite.com/hello/'
});
```

Note that my implementation is a little bit different than the way it works in `got`. We can change it, but I figured I'd show this other way first and hear your thoughts.

`input` is _resolved_ against `baseUrl`, which is in turn resolved against the document's [base URL](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base).

This is how I expect an option called `baseUrl` to behave, especially in the browser where such terminology is already built into the web platform. This isn't as much of a problem for `got` since there is no equivalent to the `<base>` tag in Node.js.

Would you rather:
 - Copy the `got` semantics and name, which is fine but potentially a bit confusing for people accustomed to how URLs resolve against base URLs in the browser
 - Copy the `got` semantics but name it something like `prefixUrl` to reflect the nature of what that does
 - Resolve the URLs as I've implemented it and let this module diverge a little bit from the semantics of `got`


readme.md
```diff
@@ -108,6 +108,15 @@ Shortcut for sending JSON. Use this instead of the `body` option. Accepts a plai
 
 Sets `options.method` to the method name and makes a request.
 
+#### baseUrl
```
sindresorhus: I know we went with `baseUrl` in Got, but hindsight, I think `baseURL` would have been better. Thoughts? // @szmarczak

szmarczak:
I'd prefer

> Copy the got semantics and name, which is fine but potentially a bit confusing for people accustomed to how URLs resolve against base URLs in the browser

Also please see https://github.com/sindresorhus/got/issues/562 and https://github.com/sindresorhus/got/pull/579

sindresorhus:
Would you be able to add an example to the `ky.extend()` docs showing usage of `baseUrl`? Like: https://github.com/sindresorhus/got#gotextendoptions

sindresorhus:
> This is how I expect an option called baseUrl to behave, especially in the browser where such terminology is already built into the web platform.

I agree. It should behave like people expect in the browser.

sindresorhus:
> I'd prefer
>> Copy the got semantics and name, which is fine but potentially a bit confusing for people accustomed to how URLs resolve against base URLs in the browser

@szmarczak Can you elaborate on why? The browser is different from Node.js. We can't make all the same decisions as with Got.

readme.md
szmarczak: I'm comfortable with both solutions :)

test.js
```diff
@@ -115,6 +119,19 @@ test('POST JSON', async t => {
...
 
+test('baseUrl option', async t => {
+	const server = await createTestServer();
+	server.get('/api/unicorn', (request, response) => {
+		response.end('rainbow');
+	});
+
+	t.is(await ky('unicorn', {
+		baseUrl: server.url + '/api/'
+	}).text(), 'rainbow');
```
sindresorhus: Can you add a test case for using an absolute URL?

sindresorhus:
We also need to decide what should happen when the base URL is missing a trailing `/`. Context: https://github.com/sindresorhus/got/issues/562#issuecomment-412796684

szmarczak:
@sindresorhus The same reason as for Got:
When baseUrl is `https://example.com/v1` and input is `/path` people expect `https://example.com/v1/path`

> We can't make all the same decisions as with Got.

I'm not saying that everything should be as Got, 'cuz we already have Got. But it wouldn't be bad to keep both features in the same way either.

readme.md
sholladay: I kind of like capitalizing letters in acronyms, however there are some [good reasons not to](https://stackoverflow.com/questions/15526107/acronyms-in-camelcase).

> Difficulty in distinguishing multiple acronyms:
>    i.e. `USID` vs `usId` (or `parseDBMXML` ...

sindresorhus: > i.e. USID vs usId

Yup, but that doesn't apply in this case. Also, it would be `usID`, not `USID`.

sholladay: Right, the first example is bad, but the point is names with multiple all-caps acronyms next to each other make it unclear that there are actually two individual acronyms in there. Not a huge deal if you already happen to know the acronym. And of course, multi-acronym names are not super common and most people seeing this module will already know what a URL is, anyway. Still, I'd say `baseUrl` is a pretty strongly held conventional in JavaScript for the above reasons.

That said, let's be rebels. I will make it `baseURL`. :)

sindresorhus:
@szmarczak I've been thinking about this a lot today and I agree.

So to be clear, it should work like this:

```js
ky('hello', {baseURL: 'https://example.com/v1'});
//=> 'https://example.com/v1/hello'

ky('/hello', {baseURL: 'https://example.com/v1/'});
//=> 'https://example.com/v1/hello'

ky('/hello', {baseURL: 'https://example.com/v1'});
//=> 'https://example.com/v1/hello'
```

Some open questions:

1. Should `baseURL` default to `document.baseURI`?

```js
document.baseURI //=> 'https://example.com/foo'

ky('/hello');
//=> 'https://example.com/foo/hello'
```

2. Should we allow a relative `baseURL` that is resolved against `document.baseURI`?

```js
document.baseURI //=> 'https://example.com'

ky('/hello', {baseURL: '/v1'});
//=> 'https://example.com/v1/hello'
```

@sholladay Thoughts on all this?

readme.md
sindresorhus: Hmm, good points. I've changed my mind. If we add `baseURL`, we should really make `throwHttpErrors` be `throwHTTPErrors`, but then we have the problem of not having clear separation. Probably easier to just always stay with camelCase no matter what. So let it be `baseUrl`.

readme.md
```diff
@@ -108,6 +108,15 @@ Shortcut for sending JSON. Use this instead of the `body` option. Accepts a plai
...
+
+Type: `string` [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL)
+
+When specified, `url` will be prepended by `baseUrl`.<br>
+If you specify an absolute URL, it will skip the `baseUrl`.
```
sindresorhus: Can you show examples of valid uses, like done in the Got docs: https://github.com/sindresorhus/got#baseurl You can just adapt those.

test.js
sindresorhus: You can borrow all of these tests: https://github.com/sindresorhus/got/pull/579/files#diff-bbc23533bfd4986af19324ccf2e7c4baR201 Just to ensure we cover it all.

szmarczak:
> Should baseURL default to document.baseURI?

> Should we allow a relative baseURL that is resolved against document.baseURI?

I agree.

millette:
If `baseUrl` contains a path, should I be able to use '../other' to bypass it or should the fetched URL always be prefixed with the `baseUrl`?

sindresorhus:
> If baseUrl contains a path, should I be able to use '../other' to bypass it or should the fetched URL always be prefixed with the baseUrl?

We are not really resolving the URL, but rather just prefixing `baseUrl`, so I don't think we want to support that.

sindresorhus:
Hmm, or instead of `document.baseURI`, should we default `baseUrl` to `location.origin`? That way the default `baseUrl` would be stable no mattter what pathname it's being run on.

```js
location.origin //=> https://example.com
document.baseURI //=> 'https://example.com/foo/bar/baz'

ky('/hello');
//=> 'https://example.com/hello'

ky('/hello', {baseURL: '/v1'});
//=> 'https://example.com/v1/hello'
```

That feels more correct to me.

felixfbecker:
I like the approach of just resolving the URL relative to the base URL a lot. In actually implemented my own clients like this in the past. It enables you to do other requests too, but have a "default" (which can be a powerful feature, but maybe also is unwanted in some cases).

One should just be aware that this means you cannot prefix your relative URL with `/` (or it will resolve from the root path), and the baseUrl usually needs to be suffixed with `/` (or `baseUrl: "foo.com/api", url: "users"` will resolve to `"foo.com/users"`).

sholladay:
> When baseUrl is https://example.com/v1 and input is /path people expect https://example.com/v1/path

I sympathize with this expectation, but I think it comes from a lack of understanding how URL resolution and base URLs work. The concept of a base URL is pervasive throughout the web platform and it works the same way everywhere. All `href` and `src` attributes on `<a>`, `` etc. are resolved against `document.baseURI` which is derived from the [`<base>`](https://html.spec.whatwg.org/multipage/semantics.html#the-base-element) tag and `<base>` in turn is resolved against `document.URL`. The resolved URL inherently depends on the presence of a trailing slash in the base URL. Even Node's non-standard, legacy API `url.resolve('https://example.com/v1', '/path')` returns `'https://example.com/path'`. And in the browser and Node, which both have the modern WHATWG `URL` API now, you'll see the same result. It's everywhere and I am not aware of any exceptions to this.

If I use `fetch('/path')`, the resulting URL that is actually fetched depends upon `document.baseURI` and whether that URL has a trailing slash, since that's how URL resolution works. `XMLHttpRequest` does the same thing, as do other HTTP clients. So using `/path` as input to an HTTP client has the same potential to be confusing no matter which client you are using.

If you want to keep the `got` behavior of appending `/` to the end of the "base URL", I'm pretty fine with that, too. But in that case, I strongly suggest we rename this option to `prefixURL` or similar. That way it is clear that we are defining our own semantics and the name is a little more intuitive, since it indicates that we are _appending_ `input` to it, as opposed to resolving against it.

readme.md
sholladay: I just noticed that `ky.HTTPError` is using the all-caps style. Want to rename it? (Separate PR, of course.)

sindresorhus:
@sholladay Thanks for your explanation. I thought I knew exactly how URL resolution worked, but I learned some new things now. I had no idea about the special trailing slash behavior.

While the URL resolution makes sense for a lot of things, I feel it just makes the common use-cases for `ky` harder to use and understand. So let's go with the Got behavior, but call it `prefixUrl` (notice it's not `prefixURL`).

sindresorhus:
@sholladay Any thoughts on https://github.com/sindresorhus/ky/pull/11#issuecomment-419213270 now that we go with `prefixUrl`?

sholladay:
Defaulting to a prefix of `location.origin` seems problematic to me, because then we have this:

```js
// Current URL: https://my-site.com/dir/page.html
fetch('hello.txt');  //=> https://my-site.com/dir/hello.txt
ky('hello.txt');  //=> https://my-site.com/hello.txt
```

As a new `ky` user, if I encountered that, I would personally get upset, because I'm expecting the migration to `ky` to be easy and that seems like a weird discrepancy.

I suggest defaulting to `document.baseURI`, regardless of whether we resolve or prefix it.

If we have a default prefix of `document.baseURI` (or, to be accurate, the parent directory thereof), then origin-relative URLs won't work the same as `fetch()`...

```js
// Current URL: https://my-site.com/dir/page.html
fetch('/hello.txt');  //=> https://my-site.com/hello.txt
ky('/hello.txt');  //=> https://my-site.com/dir/hello.txt
```

... but at least we could say it's a feature when people get tripped up by it.

sholladay:
Been kicking this around in my head for a while. Here is a proposal based on my latest thoughts.
 - No default value, to maintain maximum compatibility with `fetch`. All different kinds of relative `input` URLs should work as normal when this option is not used (`./foo`, `../foo`, `/foo`, `//foo.com/foo`)
 - When `prefixUrl` is enabled, `input` is appended as-is without transforming slashes or checking whether it is absolute. The implementation is simpler and it is highly predictable for users. You can always disable `prefixUrl` when you intend to fetch a non-prefixed resource.

This solves the common case of wanting to use this option for an `/api` prefix very well. It will work exactly as you would expect, URLs will be prefixed with `/api`.

```js
// On https://example.com/
ky('/users', { prefixUrl : '/api' });
//=> https://example.com/api/users
ky('users', { prefixUrl : '/api/' });
//=> https://example.com/api/users
```

The one (possibly) controversial part then becomes what to do about absolute `input` URLs. I think it would be bad to _usually_ append the `input` but _sometimes_ not do that - the distinction seems too arbitrary to me. Instead, you can always explicitly bypass the prefix when needed: just set `prefixUrl` to any falsey value if you don't want to use it. This would work if you are using a preconfigured instanced of `ky` that had `prefixUrl` enabled.

Below is an example of when you might want absolute `input` URLs to be prefixed, as well as how to disable that behavior when needed.

```js
const invertImage = ky.extend({ prefixUrl : 'https://image-proxy.com/invert?page=' });
invertImage('https://project.com');
//=> https://image-proxy.com/invert?page=https://project.com
invertImage('https://project.com/status', { prefixUrl : false });
//=> https://project.com/status
```

What do you guys think? Should be very straightforward and easy to use while still pretty powerful and flexible. But is it convenient enough when you want to make a request to a different domain from a preconfigured `ky` instance?

sindresorhus:
I've been thinking a lot about this too and I fully agree with your proposal. This will give us the best of both worlds. Let's go with this.

sholladay:
Ready for review again. :)

 - `baseUrl` -> `prefixUrl`, with the new semantics described [here](https://github.com/sindresorhus/ky/pull/11#issuecomment-422642912).
 - Added examples to the documentation.
 - Added more test cases. Let me know if you want them as separate tests with their own title. I also wanted to implement tests for relative URLs, but `node-fetch` does not support them. Happy to do this when we have browser testing!

sindresorhus:
It should also handle double and no slashes when joining the `prefixUrl` and the `input` (Just like `path.join()`). So these should be valid:

```js
await ky('unicorn', {prefixUrl: 'https://cats.com'});
await ky('unicorn', {prefixUrl: 'https://cats.com/'});
await ky('/unicorn', {prefixUrl: 'https://cats.com'});
await ky('/unicorn', {prefixUrl: 'https://cats.com/'});
```

sholladay:
Deduplicating slashes when `prefixUrl` ends with one and `input` begins with one is reasonable and should be fine.

I'm not sure how I feel about adding a slash when none are present. Seems a little more intrusive, with only a marginal benefit. That said, I went ahead and implemented it. 👍

sindresorhus:
> I'm not sure how I feel about adding a slash when none are present. Seems a little more intrusive, with only a marginal benefit.

When using `prefixUrl`, do you think we should enforce the `input` argument to either start with a `/` (`/unicorn`) or enforce it to not (`unicorn`)? If so, which one should we enforce? I kinda like the idea of consistency.

sholladay:
If we were to add validation like that, I would prefer to enforce that `input` has to begin with a slash _or_ `prefixUrl` has to end with one. That way people can choose which style they prefer. In my own projects, I would go with the latter as I dislike the extra slashes at every call site. But someone else may prefer it the other way and that's acceptable to me, too.

szmarczak:
I'd go for the latter

sindresorhus:
I don't like that you have to know whether `prefixUrl` ends with a slash or not. You would not even usually see the `prefixUrl` at the callsite as you would use it with `.extend()`. I would go with `input` being required to not start with a slash when used with `prefixUrl` as, like you said, it's shorter and it also makes it clear it's not an absolute path reference, but I would let `prefixUrl` end with a slash or not.

So, these are allowed:

```js
await ky('unicorn', {prefixUrl: 'https://cats.com'});
await ky('unicorn', {prefixUrl: 'https://cats.com/'});
```

But not:

```js
await ky('/unicorn', {prefixUrl: 'https://cats.com'});
await ky('/unicorn', {prefixUrl: 'https://cats.com/'});
```

sholladay:
> but I would let prefixUrl end with a slash or not

To confirm, when neither of them have a slash, you still want to join them with a slash, correct? Or do we apply the prefix as-is?

sindresorhus:
> when neither of them have a slash, you still want to join them with a slash, correct?

Yes

sholladay:
Alright, done. PTAL. :)

sindresorhus:
This looks great! Thank you so much for the great conversation, @sholladay. I think we ended up with a good compromise :)

PR #14, opened 2018-09-06
Add `throwHttpErrors` option

### LINKED ISSUE #9: `throwHttpErrors` option
It's always true. Let's make it an option so we can turn that off.

killynathan:
Fixes #9 

adds `throwHttpErrors` option defaulted to `true` along with updated docs and tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -98,11 +98,15 @@ class Ky {
 			credentials: 'same-origin', // TODO: This can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
 			retry: 3,
 			timeout: 10000,
+			throwHttpErrors: true,
 			...options
 		};
 
 		this._timeout = this._options.timeout;
 		delete this._options.timeout;
+
+		this._throwHttpErrors = this._options.throwHttpErrors;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -98,12 +98,16 @@ class Ky {
...
+			throwHttpErrors: true,
 			...options
 		};
 
 		this._timeout = this._options.timeout;
 		delete this._options.timeout;
 
+		this._throwHttpErrors = this._options.throwHttpErrors;
+		delete this._options.throwHttpErrors;
+
```
szmarczak: Remove these lines.

index.js
```diff
@@ -153,7 +157,9 @@ class Ky {
 					return retry();
 				}
 
-				throw error;
+				if (this._throwHttpErrors) {
```
```
if (this._options.throwHttpErrors) {
```

szmarczak:
I'd follow https://github.com/sindresorhus/ky/issues/15

index.js
killynathan: Got it. I thought I should delete it because like timeout, it's not something we want to pass to fetch.

szmarczak:
LGTM.

PR #16, opened 2018-09-06
Refactor TimeoutError

vivekimsit:


index.js
```diff
@@ -79,14 +79,10 @@ class TimeoutError extends Error {
...
-
-	(async () => {
-		await delay(ms);
-		reject(new TimeoutError());
-	})();
-});
+const timeout = (promise, ms) => Promise.race([
+	promise,
+	delay(ms).then(() => Promise.reject(new TimeoutError())) // eslint-disable-line promise/prefer-await-to-then
+]);
```
sindresorhus: Clever idea. Finally a good use for `Promise.race()`.

I think I would prefer using async function though, like this:

```js
const timeout = (promise, ms) => Promise.race([
	promise,
	(async () => {
		await delay(ms);
		throw new TimeoutError();
	})()
]);
```

sindresorhus: I prefer using `await` whenever possible. I personally find `await` more readable, even with the IIFE-wrapper.

vivekimsit: Done.

sindresorhus:
Thanks :)

PR #23, opened 2018-09-08
Feat: convert cjs -> es6 modules

dangdennis:
Use the now standardized es6 export syntax. You can export the defined default, classes, and any extra function as usual.


dangdennis:
#24

PR #34, opened 2018-09-22
Add `onDownloadProgress` option

### LINKED ISSUE #3: Stream API

### LINKED ISSUE #6: Progress events
We could use streams for that :)

dangdennis:
Here's the new PR for the potential stream API. 

Basic example for reading an image stream:
```js
function streamProgressCallback(percentage, bytes) {
     console.log('percentage:: ' + percentage)
     console.log('bytes:: ' + bytes)
}

const res = await ky.get(someImageURL, { stream: true, streamProgressCallback }).blob();

const imgObject = URL.createObjectURL(res);
const img = document.createElement('img');
img.src = img1;
document.body.appendChild(img)
```

Not sure how to test streams.

---

Fixes #3
Fixes #6

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,12 +2,6 @@
 
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
-type JSONObject = {[key: string]: JSONValue};
-interface JSONArray extends Array<JSONValue> {}
-export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
-
-export type JSONStringifyable = string | number | boolean | null | object;
-
 export type BeforeRequestHook = (options: Options) => void | Promise<void>;
 
 export type AfterResponseHook = (response: Response) => Response | void | Promise<Response | void>;
@@ -39,7 +33,7 @@ export interface Options extends RequestInit {
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 	*/
-	json?: JSONStringifyable;
+	json?: unknown;
 
 	/**
 	Search parameters to include in the request URL.
@@ -107,7 +101,7 @@ export interface ResponsePromise extends Promise<Response> {
 	formData(): Promise<FormData>;
 
 	// TODO: Use `json<T extends JSONValue>(): Promise<T>;` when it's fixed in TS.
-	// See https://github.com/sindresorhus/ky/pull/80
+	// See https://github.com/microsoft/TypeScript/issues/15300 and https://github.com/sindresorhus/ky/pull/80
 	/**
 	Get the response body as JSON.
 
@@ -129,7 +123,7 @@ export interface ResponsePromise extends Promise<Response> {
 	const result = await ky(…).json<Result>();
 	```
 	*/
-	json<T = JSONValue>(): Promise<T>;
+	json<T = unknown>(): Promise<T>;
 
 	text(): Promise<string>;
 }
@@ -219,7 +213,16 @@ declare const ky: {
 	delete(url: Request | URL | string, options?: Options): ResponsePromise;
 
 	/**
+	Create a new Ky instance with complete new defaults.
+
+	@returns A new Ky instance.
+	*/
+	create(defaultOptions: Options): typeof ky;
+
+	/**
 	Create a new Ky instance with some defaults overridden with your own.
+
+	In contrast to `ky.create()`, `ky.extend()` inherits defaults from its parent.
 
 	@returns A new Ky instance.
 	*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,7 +33,7 @@ export interface Hooks {
 }
 
 /**
-Options are the same as fetch, with some exceptions.
+Options are the same as `window.fetch`, with some exceptions.
 */
 export interface Options extends RequestInit {
 	/**
@@ -141,7 +141,7 @@ export class TimeoutError extends Error {
 	constructor();
 }
 
-export interface Ky {
+declare const ky: {
 	/**
 	Fetch the given `url`.
 
@@ -215,9 +215,7 @@ export interface Ky {
 
 	@returns A new Ky instance.
 	*/
-	extend(defaultOptions: Options): Ky;
+	extend(defaultOptions: Options): typeof ky;
 }
-
-declare const ky: Ky;
 
 export default ky;
--- a/index.d.ts
+++ b/index.d.ts
@@ -70,8 +70,8 @@ export interface Options extends RequestInit {
 
 	/**
 	Download progress event handler.
-	The function takes `percent`, `transferredBytes`, `totalBytes` and `chunk` arguments.
-	If it's not possible to retrieve the body size, total will be `0`.
+
+	If it's not possible to retrieve the body size, `totalBytes` will be `0`.
 	*/
 	onDownloadProgress?: (percent: number, transferredBytes: number, totalBytes: number, chunk?: Uint8Array) => void;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -68,6 +68,13 @@ export interface Options extends RequestInit {
 	timeout?: number;
 
 	/**
+	Download progress event handler.
+	The function takes `percent`, `transferredBytes`, `totalBytes` and `chunk` arguments.
+	If it's not possible to retrieve the body size, total will be `0`.
+	*/
+	onDownloadProgress?: (percent: number, transferredBytes: number, totalBytes: number, chunk?: Uint8Array) => void;
+
+	/**
 	Hooks allow modifications during the request lifecycle. Hook functions may be async and are run serially.
 	*/
 	hooks?: Hooks;
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,6 +9,10 @@ export type AfterResponseHook = (response: Response) => Response | void | Promis
 export interface DownloadProgress {
 	percent: number;
 	transferredBytes: number;
+	
+	/**
+	If it's not possible to retrieve the body size, it will be `0`.
+	*/
 	totalBytes: number;
 }
 
@@ -82,8 +86,6 @@ export interface Options extends RequestInit {
 
 	/**
 	Download progress event handler.
-
-	If it's not possible to retrieve the body size, `progress.totalBytes` will be `0`.
 	*/
 	onDownloadProgress?: (progress: DownloadProgress, chunk: Uint8Array) => void;
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,6 +6,12 @@ export type BeforeRequestHook = (options: Options) => void | Promise<void>;
 
 export type AfterResponseHook = (response: Response) => Response | void | Promise<Response | void>;
 
+export interface DownloadProgress {
+	percent: number;
+	transferredBytes: number;
+	totalBytes: number;
+}
+
 export interface Hooks {
 	/**
 	Before the request is sent.
@@ -67,7 +73,7 @@ export interface Options extends RequestInit {
 
 	If it's not possible to retrieve the body size, `totalBytes` will be `0`.
 	*/
-	onDownloadProgress?: (percent: number, transferredBytes: number, totalBytes: number, chunk?: Uint8Array) => void;
+	onDownloadProgress?: (progress: DownloadProgress, chunk?: Uint8Array) => void;
 
 	/**
 	Hooks allow modifications during the request lifecycle. Hook functions may be async and are run serially.
--- a/index.d.ts
+++ b/index.d.ts
@@ -61,11 +61,12 @@ export interface Options extends RequestInit {
 	retry?: number;
 
 	/**
-	Timeout in milliseconds for getting a response.
+	Timeout in milliseconds for getting a response. Can not be greater than 2147483647.
+	If set to `false`, there will be no timeout.
 
 	@default 10000
 	*/
-	timeout?: number;
+	timeout?: number | false;
 
 	/**
 	Download progress event handler.
--- a/index.d.ts
+++ b/index.d.ts
@@ -86,6 +86,8 @@ export interface Options extends RequestInit {
 
 	/**
 	Download progress event handler.
+	
+	@param chunk - Note: It's empty for the first call.
 	*/
 	onDownloadProgress?: (progress: DownloadProgress, chunk: Uint8Array) => void;
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,9 +6,9 @@ export type JSONValue = string | number | boolean | null | JSONObject | JSONArra
 
 export type JSONStringifyable = string | number | boolean | null | object;
 
-export type BeforeRequestHook = (options: Options) => void;
+export type BeforeRequestHook = (options: Options) => void | Promise<void>;
 
-export type AfterResponseHook = (response: Response) => Response | void;
+export type AfterResponseHook = (response: Response) => Response | void | Promise<Response | void>;
 
 export interface Hooks {
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+/// <reference lib="dom"/>
+
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
 type JSONObject = {[key: string]: JSONValue};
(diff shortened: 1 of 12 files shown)


sindresorhus:
I don't think `node-fetch` supports Stream, so we need to test in a real browser.

What's the best tool for browser testing these days? Preferably something using Puppeteer. Probably https://github.com/direct-adv-interfaces/mocha-headless-chrome or https://github.com/mantoni/mochify.js // @sholladay

sholladay:
I am aware of three high quality solutions for browser testing:
 - Puppeteer, which you already mentioned. My team recently used this to debug problems with our signup system. Turns out we had a cookie with an expiration time that was off by a few zeros (seconds vs milliseconds), which meant you had to fill out a form relatively quickly or it would fail. Puppeteer was very helpful to whip up a quick script to reliably reproduce the issue and narrow it down to a timing issue and eventually identify the root cause. I liked it, it was nice. I would not use it full time, though, because it's proprietary and only works with Chrome.
 - [Cypress](https://cypress.io), which is up and coming. They are trying to get around some of the limitations of WebDriver and to improve performance. It also does a cool thing where it records the state history so you can go backwards in time, a la the Redux DevTools. Unfortunately, it also only works with Chromium based browsers at the moment. They supposedly have plans to support other browsers, but I doubt it will happen any time soon (very hard problem to solve correctly).
 - [Intern](https://theintern.io/), which is what I use at companies. It wraps WebDriver in an easy to use package and handles a lot of bugs and edge cases automatically. It is the only one that is based on open web standards and is completely cross-browser. It is configuration-centric with good defaults. It has a great community and team behind it, along with a solid feature set (I paid SitePen in the past to develop some of them). At times it feels a bit verbose, but overall it is my favorite.

SitePen wrote [a post about Puppeteer](https://www.sitepen.com/blog/2017/10/04/browser-automation-with-puppeteer/) that is worth reading.

For `ky`, a lot of it will depend on your goals. Do you want to be able to test in Firefox, Safari, or Edge?  If so, I would go with Intern. Otherwise, I think Cypress is the most appropriate (it was designed with CI in mind). But there are still lots of tools in the Puppeteer ecosystem I have not tried yet, so maybe there is some awesome test runner integration out there. I see some `ava` based boilerplates (e.g. [ava-puppeteer](https://github.com/jakwuh/ava-puppeteer) and [ava-puppeteer-example](https://github.com/StephenDavidson/ava-puppeteer-example), but I am not sure how well they work, though.

sindresorhus:
Sounds like Cypress would be the best fit then. I only care about testing it with Chromium.

@dangdennis Would you be up for trying to integrate some Cypress tests?

dangdennis:
Thanks for the extensive thoughts @sholladay. 

I'll take this - I'll set up Cypress and a minimal test process.

sindresorhus:
> I'll take this - I'll set up Cypress and a minimal test process.

Awesome! Can you do it as a separate PR? Since there are other PRs that need it too, so it's not blocked by this PR.

ianwalter:
I'm a fan and use Cypress every day but I think it is overkill and not super well-suited for this.

For unit tests like these, I would recommend [Karmatic](https://github.com/developit/karmatic#readme) which wraps Puppeteer, Karma, and Jasmine.

sholladay:
Hmm I've used Karma but never heard of Karmatic. It's emblematic of the problem that Karma has, though... it needs a layer on top of it to be manageable and useful. I do like the developer experience of a full Karma environment when it's properly set up, but this seems like it's just hiding the complexity.

Some specific problems I see:
 - Karmatic downloads an entire Chromium installation. It's enormous! The install size is 273MB. This will slow down CI runs and ties us down to whatever version it feels like downloading. Any decent CI system comes with a built-in browser, which is usually pre-configured with certain settings for debugging, and it would be better to use that, both for performance reasons and because we're in control of it.
 - Karmatic has a peer dependency on Webpack, which I highly doubt Sindre will want to add to ky. But we'll see.

dangdennis:
Sorry for the delay folks. This is PR is still hanging on the other branch Cypress PR to fulfill. That's being delayed by a bug within Cypress and Typescript (or something, I forget). 

Once that PR is merged, we can continue working on a method for end-to-end testing for the stream API.

ianwalter:
@sholladay If you think that is big 😄 
image

But I am thankful that Cypress now uses a global cache.

sindresorhus:
We're now using Cypress, so this PR can be continued.

sindresorhus:
@dangdennis I don't think we even need a `stream` option yet. Let's go with the minimal thing to expose, a `onProgress` option.

```js
const onProgress = stats => {
	console.log(stats.percent, stats.bytesRead, stats.bytesTotal);
};

const res = await ky(someImageURL, {onProgress}).blob();
```

sindresorhus:
Can you also fix the merge conflict and rebase from master branch?

dangdennis:
I just fought with Git for an hour trying to squash my commits. So glad that's over with.

If onProgress is passed, Ky will automatically stream the response.

@sindresorhus I still can't get the build to pass.

sindresorhus:
> Unfortunately, @sindresorhus I still can't get the build to pass.

Probably because of https://github.com/sindresorhus/ky/pull/34/files#diff-cd87bb3712c5cc8ef9426451d8b99768L2

sindresorhus:
Also look at the diff, there are still many leftovers from the incorrect merge conflict resolving that needs to be fixed, like https://github.com/sindresorhus/ky/pull/34/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9L121

sholladay:
I'm willing to help on this. If I can get write access to the fork, I'll work on tests and see what I can do about the diff and commit history.

dangdennis:
@sholladay That'd be help. Is that something @sindresorhus can enable?

sholladay:
Nope, since it's your fork, you have control over it.

You can add me here: https://github.com/dangdennis/ky/settings/collaboration

dangdennis:
@sholladay So i cleaned up the commits. I'll attempt to write the integration test now.

dangdennis:
Update: Haven't had time to write this test yet. Hoping to get to it and the documentation this weekend or next weekend.

sindresorhus:
@dangdennis Still interested in finishing this? :)

dangdennis:
I am! I had so much trouble getting cypress to do what i want. But if you or someone is available to finish this up, I will greatly appreciate.

szmarczak:
> If you or someone is available to finish this up, I will greatly appreciate.

Sure, I'll do. I'll try tomorrow.

stramel:
@dangdennis With the merge of #84 hopefully testing should be easier. Looking forward to getting this in! 👍

index.js
```diff
@@ -292,6 +301,50 @@ class Ky {
...
+
+								if (percent !== 1) {
+									onProgress(percent, transferred, total);
+								}
+							}
+
+							controller.enqueue(value);
+							read();
+						} catch (error) {
+							console.log(error);
```
sindresorhus: Leftover?

sindresorhus:
I just hit me, this is for download progress only, right? Not upload? If so, I think we should make that clear by naming the function `onDownloadProgress` instead of `onProgress`.

sindresorhus:
Also, this PR introduces some loss in code coverage (see above).

index.js
```diff
@@ -292,6 +301,50 @@ class Ky {
 
 		return timeout(fetch(this._input, this._options), this._timeout, this.abortController);
 	}
+
+	_stream(response, onProgress) {
+		const total = Number(response.headers.get('content-length')) || 0;
+		let transferred = 0;
+
+		return new Response(
+			new ReadableStream({
```
sindresorhus: These globals should use our globals thing, so it's possible to polyfill them.

readme.md
```diff
@@ -195,6 +195,13 @@ Default: `10000`
 
 Timeout in milliseconds for getting a response.
 
+##### onProgress
+
+Type: `Function`<br>
+Default: `10000`
```
sindresorhus: Huh?

readme.md
```diff
@@ -195,6 +195,13 @@ Default: `10000`
...
+
+Progress event handler. The function takes `percent`, `transferred` and `total` arguments. If it's not possible to retrieve the body size, total will be `0`.
```
sindresorhus: Maybe show a short example here that logs the progress?

test/browser.js
```diff
@@ -61,3 +61,55 @@ test('aborting a request', withPage, async (t, page) => {
 
 	await server.close();
 });
+
+test('onProgress works', withPage, async (t, page) => {
+	const server = await createTestServer();
+
+	server.get('/', (request, response) => {
+		response.end('meow');
```
sindresorhus: This doesn't really test streaming as it's only one chunk. I'm not sure how it splits up chunks, but we should ensure it uses at least two chunks.

sindresorhus:
Note to self: We might be able to use https://github.com/mpotra/whatwg-streams in `ky-universal` to get Node.js support.

readme.md
szmarczak: Oh, just forgot to remove that part :P (quick copy & paste)

test/browser.js
szmarczak: Will do.

szmarczak:
> Also, this PR introduces some loss in code coverage (see above).

That's because nyc doesn't check tests ran in browser...

sindresorhus:
> That's because nyc doesn't check tests ran in browser...

Ah, makes sense. Doesn't matter. We can just use the nyc ignore comments.

sindresorhus:
@szmarczak Would you be able to finish this up?

szmarczak:
@sindresorhus Done :D

sindresorhus:
You need to add it to the TS definition too.

sindresorhus:
It's difficult to correctly polyfill `ReadableStream`, so I think we should be graceful there and only require it if the `onProgress` option is used, and document that fact.

index.js
```diff
@@ -292,6 +303,46 @@ class Ky {
 
 		return timeout(fetch(this._input, this._options), this._timeout, this.abortController);
 	}
+
+	/* istanbul ignore next */
+	_stream(response, onProgress) {
+		const total = Number(response.headers.get('content-length')) || 0;
+		let transferred = 0;
```
sindresorhus: "Transferred" what? Maybe call it `transferredBytes`.

index.js
```diff
@@ -292,6 +303,46 @@ class Ky {
...
+		return new Response(
+			new ReadableStream({
+				start(controller) {
+					const reader = response.body.getReader();
+
+					if (onProgress) {
+						onProgress(0, 0, total);
+					}
+
+					read();
```
sindresorhus: It's weird to have this before the definition. Can you move it to below the function?

index.js
```diff
@@ -292,6 +303,46 @@ class Ky {
 
 		return timeout(fetch(this._input, this._options), this._timeout, this.abortController);
 	}
+
+	/* istanbul ignore next */
+	_stream(response, onProgress) {
+		const total = Number(response.headers.get('content-length')) || 0;
```
sindresorhus: `totalBytes`

readme.md
```diff
@@ -195,6 +195,24 @@ Default: `10000`
...
+##### onProgress
+
+Type: `Function`
+
+Progress event handler. The function takes `percent`, `transferred` and `total` arguments. If it's not possible to retrieve the body size, total will be `0`.
+
+
+```js
+await ky('https://example.com', {
+	onProgress: (percent, transferred, total) => {
```
sindresorhus: I think it might be nicer to make this an object instead. Could be called `progress`.

`transferred` => `transferredBytes`
`total` => `totalBytes`

sindresorhus:
Also note: https://github.com/sindresorhus/ky/pull/34#issuecomment-466477985

> I just hit me, this is for download progress only, right? Not upload? If so, I think we should make that clear by naming the function onDownloadProgress instead of onProgress.

szmarczak:
> It's difficult to correctly polyfill ReadableStream, so I think we should be graceful there and only require it if the onProgress option is used, and document that fact.

Does that change anything?

sindresorhus:
> Does that change anything?

Yes, like I commented, we need to document what needs to be polyfilled in the docs for the option. We need to throw if the option is used and ReadableStream is not available, and we need to ensure Ky works fine without ReadableStream if the `onProgress` option is not used.

szmarczak:
> We need to throw if the option is used and ReadableStream is not available, and we need to ensure Ky works fine without ReadableStream if the onProgress option is not used.

Ah, now I get it. Thanks for clarifying, my mind doesn't work well at nights :P

index.js
```diff
@@ -292,6 +303,46 @@ class Ky {
...
+							controller.close();
+							return;
+						}
+
+						if (onProgress) {
+							transferred += value.byteLength;
+							const percent = total === 0 ? 0 : transferred / total;
+
+							if (percent !== 1) {
+								onProgress(percent, transferred, total);
```
szmarczak: I think we should also give the chunk here.

szmarczak: Also I think we should remove `setInterval` in Got and emit progress events every new chunk like it's done here.

sindresorhus: > I think we should also give the chunk here.

Sure. Could be useful until we implement a proper stream API.

> Also I think we should remove setInterval in Got and emit progress events every new chunk like it's done here.

👍 If it's possible. There's must be a good reason we did the `setInterval` thing in the first place.

szmarczak: > There's must be a good reason we did the setInterval thing in the first place, though.

The only reason I can think of is to not spam the progress events.

szmarczak:
I'll have to make a workaround to check `undefined` chunk.

```
const puppeteer = require('puppeteer');

(async () => {
    const browser = await puppeteer.launch();
    const page = await browser.newPage();

    await page.goto('about:blank');

    const result = await page.evaluate(async url => {
        return [undefined];
    }, 'about:blank');

    console.log(result);

     await page.close();
     await browser.close();
})();
```

```
$ nano demo.js
[ null ]
```

dangdennis:
@szmarczak Thank you for taking this to completion - 6 month PR!

szmarczak:
Yes, it was sent 6 months ago. But it needed just a few days to complete :P

sindresorhus:
@szmarczak https://github.com/sindresorhus/ky/pull/34#discussion_r273663506 is not done.

test/browser.js
```diff
@@ -88,15 +88,16 @@ test('onDownloadProgress works', withPage, async (t, page) => {
 
 		const text = await window.ky(url, {
 			onDownloadProgress: (percent, transferredBytes, totalBytes, chunk) => {
-				data.push([percent, transferredBytes, totalBytes, chunk instanceof Uint8Array ? decodeUTF8(chunk) : chunk]);
+				const stringifiedChunk = chunk instanceof Uint8Array ? decodeUTF8(chunk) : String(chunk);
```
sindresorhus: We need to document in the readme and typings in what situations the `chunk` would be `undefined`.

test/browser.js
```diff
@@ -86,3 +88,90 @@ test('throws TimeoutError even though it does not support AbortController', with
...
+	});
+
+	await page.goto(server.url);
+	await page.addScriptTag({path: './umd.js'});
+
+	const result = await page.evaluate(async url => {
+		window.ky = window.ky.default;
+
+		// `new TextDecoder('utf-8').decode` hangs up?
+		const decodeUTF8 = array => String.fromCharCode.apply(null, array);
```
sindresorhus:
```suggestion
		const decodeUTF8 = array => String.fromCharCode(...array);
```

test/browser.js
szmarczak: It's `undefined` only for the first call.

szmarczak:
> @szmarczak [#34 (comment)](https://github.com/sindresorhus/ky/pull/34#discussion_r273663506) is not done.

Oh, forgot about that. Will change.

sindresorhus:
@szmarczak You also need to fix the merge conflict.

index.d.ts
```diff
@@ -62,6 +68,13 @@ export interface Options extends RequestInit {
 	*/
 	timeout?: number | false;
 
+	/**
+	Download progress event handler.
+
+	If it's not possible to retrieve the body size, `totalBytes` will be `0`.
```
sindresorhus: This should be documented on the actual `totalBytes` property.

index.d.ts
```diff
@@ -62,6 +68,13 @@ export interface Options extends RequestInit {
...
+	*/
+	onDownloadProgress?: (progress: DownloadProgress, chunk?: Uint8Array) => void;
```
sindresorhus: I think this should be placed last.

index.js
```diff
@@ -234,6 +236,20 @@ class Ky {
 				throw new HTTPError(response);
 			}
 
+			// If `onDownloadProgress` is passed, use stream API internally
```
sindresorhus:
```suggestion
			// If `onDownloadProgress` is passed it uses the API internally
```

readme.md
```diff
@@ -196,6 +196,27 @@ Default: `10000`
...
 If set to `false`, there will be no timeout.
 
+##### onDownloadProgress
+
+Type: `Function`
+
+Download progress event handler. The function takes `progress` and `chunk` arguments.<br>
+The `progress` object contains following elements: `percent`, `transferredBytes` and `totalBytes`. If it's not possible to retrieve the body size, total will be `0`.
+
+For the first call, `chunk` will be always `undefined`.
```
sindresorhus: Instead of `undefined`, wouldn't it be better to just emit an empty chunk (empty `Uint8Array`)? Then people would not need to implement conditional handling and the types would be simpler.

readme.md
```diff
@@ -196,6 +196,27 @@ Default: `10000`
...
+
+
+```js
```
sindresorhus:
```suggestion
```js
import ky from 'ky';
```

readme.md
```diff
@@ -196,6 +196,27 @@ Default: `10000`
 Timeout in milliseconds for getting a response. Can not be greater than 2147483647.
 If set to `false`, there will be no timeout.
 
+##### onDownloadProgress
+
+Type: `Function`
+
+Download progress event handler. The function takes `progress` and `chunk` arguments.<br>
```
sindresorhus:
```suggestion
Download progress event handler. The function receives a `progress` and `chunk` argument.<br>
```

readme.md
```diff
@@ -196,6 +196,27 @@ Default: `10000`
...
+The `progress` object contains following elements: `percent`, `transferredBytes` and `totalBytes`. If it's not possible to retrieve the body size, total will be `0`.
```
sindresorhus:
```suggestion
The `progress` object contains the following elements: `percent`, `transferredBytes` and `totalBytes`. If it's not possible to retrieve the body size, `totalBytes` will be `0`.
```

readme.md
szmarczak: Will do. Great idea.

sindresorhus:
Thanks all! 🙌

PR #46, opened 2018-10-05
Include a ES5 build

### LINKED ISSUE #44: Include a UMD build
I'm having to transpile the module before usage and I'm not sure if that's intentional.

This is the line which caused me pain. 

https://github.com/sindresorhus/ky/blob/eb4b341b8dac088ca6ec436eedc536d101c4b02c/index.js#L192

Feel free to close this is if the use of export is intentional, I couldn't really tell from the docs.

rifler:
Fixes #44

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,11 +13,14 @@
 		"node": ">=8"
 	},
 	"scripts": {
-		"test": "xo && nyc ava && tsd-check"
+		"test": "xo && nyc ava && tsd-check",
+		"prepack": "cp index.js ky.js && babel ky.js --out-file es5.js && cp index.d.ts es5.d.ts"
 	},
 	"files": [
 		"index.js",
-		"index.d.ts"
+		"es5.js",
+		"index.d.ts",
+		"es5.d.ts"
 	],
 	"keywords": [
 		"fetch",
@@ -42,6 +45,9 @@
 		"node-fetch"
 	],
 	"devDependencies": {
+		"@babel/cli": "7.1.2",
+		"@babel/core": "7.1.2",
+		"@babel/preset-env": "7.1.0",
 		"ava": "1.0.0-beta.8",
 		"body": "^5.1.0",
 		"codecov": "^3.0.4",
@@ -52,6 +58,18 @@
 		"nyc": "^13.0.1",
 		"tsd-check": "^0.2.1",
 		"xo": "^0.23.0"
+	},
+	"babel": {
+		"presets": [
+			[
+				"@babel/preset-env",
+				{
+					"loose": true,
+					"modules": "umd",
+					"useBuiltIns": "usage"
+				}
+			]
+		]
 	},
 	"xo": {
 		"envs": [
--- a/readme.md
+++ b/readme.md
@@ -54,6 +54,24 @@ import ky from 'ky';
 })();
 ```
 
+<details>
+
+<summary>Also you can use it in es5 environment. Click here to see how</summary>
+
+```js
+const ky = require('ky/es5').default;
+
+ky
+	.post('https://some-api.com', {json: {foo: true}})
+	.json()
+	.then((json) => {
+		console.log(json);
+	});
+```
+
+</details>
+
+
 With plain `fetch`, it would be:
 
 ```js
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 node_modules
 yarn.lock
 .nyc_output
+# trash after es5 build
+/es5.js
+/es5.d.ts
+/ky.js


package.json
```diff
@@ -53,6 +59,16 @@
 		"tsd-check": "^0.2.1",
 		"xo": "^0.23.0"
 	},
+	"babel": {
+		"presets": [
+			[
+				"@babel/preset-env",
+				{
+					"loose": true
```
sindresorhus: Lets set `modules` to `umd` so it can even work without using a build-system.

sindresorhus:
Can you also mention it in the readme?

sindresorhus:
It seems you also need to include `@babel/polyfill`.

sindresorhus:
We should probably try this https://babeljs.io/docs/en/next/babel-preset-env.html#usebuiltins-usage-experimental option and see if it works.

package.json
sindresorhus: You also need to ensure `ky` is exported as a global with this config. I don't know how you do that in Babel.

rifler:
LGTM
image

I've updated readme too. Also make copy of index.js named ky.js, otherwise babel makes window.index instead of window.ky

readme.md
```diff
@@ -54,6 +54,24 @@ import ky from 'ky';
 })();
 ```
 
+<details>
+
+<summary>Also you can use it in es5 environment. Click here to see how</summary>
+
+```js
+const ky = require('ky').default;
```
sindresorhus: This should be `ky/es5`

readme.md
```diff
@@ -54,6 +54,24 @@ import ky from 'ky';
 })();
 ```
 
+<details>
+
+<summary>Also you can use it in es5 environment. Click here to see how</summary>
+
```
sindresorhus: This also needs to mention the requirement of the fetch polyfill.

sindresorhus:
> Also make copy of index.js named ky.js, otherwise babel makes window.index instead of window.ky

Is there really no better way of naming the default export? If not, you should open an issue on Babel about that. It's silly having to copy a file twice just to work around that.

sindresorhus:
You also need to test it on a ES5-only browser to ensure it works there, like IE11.

maple3142:
One way to do it is https://www.npmjs.com/package/babel-plugin-rename-umd-globals , or you can use a bundler suck as rollup, webpack along with babel.

readme.md
deadNightTiger: And a dependency on `@babel/polyfill` as well (or specifically `core-js@2` and `regenerator-runtime`).

rifler:
I suppose ky will become too heavy with all of this polyfills, regenerator-runtime etc.
The initial purpose was to transpile es6 exports to commonjs not to break ssr, cause node doesn't understand them well. But now it become too complex, overgrow with babel plugins, babel polyfills, corejs. Why not just create two files with `exports` and `module.exports` (other code will be the same) ?

deadNightTiger:
Another problem seems to be mixing named exports with `export default`, as there is no good way to represent this in CommonJS.
Usually you would transform `export { whatever };` to `exports.whatever = whatever;` and `export default whatever;` to `module.exports = whatever;`. But when both are present, you have to do `exports.default = whatever;` and then `require("module").default;` at the consuming side, which is not really idiomatic or convenient.
The best option might be just distributing the source and letting the consumer transpile/bundle it the way they want it.

sindresorhus:
> I suppose ky will become too heavy with all of this polyfills, regenerator-runtime etc. [...] But now it become too complex, overgrow with babel plugins, babel polyfills, corejs.

You have perfectly illustrated exactly why I usually say *no* to adding Babel to my packages. There's no way to solve the constraint of supporting all targets while still not being bloated.

sindresorhus:
> The initial purpose was to transpile es6 exports to commonjs not to break ssr, cause node doesn't understand them well.

Yes, let's just do that. Doesn't seem to be a good way to solve the ES5 problem without a total bloat-fest. CJS should be `index.js` and the source should be `module.js` and use a `"module"` key in package.json, see https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages

sindresorhus:
@rifler Still interested in finishing this up? :)

prichodko:
What about using https://github.com/developit/microbundle?

sindresorhus:
@prichodko If you're going to propose using something, I would recommend including arguments for what the benefits would be. `microbundle` doesn't solve any of the problems discussed in this PR, but introduces lots of other ones, since it uses Buble instead of Babel, which is super buggy and unmaintained.

prichodko:
@sindresorhus fair point. Sorry for that.

I think what was agreed on that ky is missing `cjs` build pointing to a `main` property in `package.json`, so I wanted to introduce a one-dependency solution for managing builds for multiple targets - primarily `cjs` (via main) and `es` (via module). I love the simplicity of `ky`, but it's a shame that it doesn't work out of the box with a framework like `Next.js`, which has a modern build setup. 

I've been using microbundle successfully in couple of mine libraries. More specifically - last commit to Buble was 11 days ago, [Rollup](https://github.com/rollup/rollup/blob/master/package.json#L72) is using it and @developit's recently released [htm](https://github.com/developit/htm) library is using microbundle as well, so I am not sure I follow.

sindresorhus:
Closing for lack of activity.

PR #47, opened 2018-10-08
update(readme): fix svg on npm

fouad:


sindresorhus:
No thanks. This should be fixed by npm instead. Open an issue here: https://github.com/npm/marky-markdown

fouad:
@sindresorhus I don't think that's something they can fix. The issue is that the [image hosted by GitHub](https://raw.githubusercontent.com/sindresorhus/ky/HEAD/media/logo.svg) doesn't have a content-type. If you're cool with a `png` instead of an `svg`, that'd work too.

sindresorhus:
Sorry, I'm not interested in doing anything special for npm.

sindresorhus:
> The issue is that the image hosted by GitHub doesn't have a content-type.

It does if you add `?sanitize=true`. https://raw.githubusercontent.com/sindresorhus/ky/HEAD/media/logo.svg?sanitize=true

fouad:
Ahh gotcha! didn't know that — will file an issue in `marky-markdown` with that query param as expected behavior for svg's

PR #50, opened 2018-10-09
Add type assertion for `.extend()` argument

### LINKED ISSUE #40: Maybe some assertions for `defaults` type in `extend`?
The `defaults` param will be used to `deepMerge` with other options and the `deepMerge` method will flat items for array argument.
Everything goes on normal and users may not know what happen if they pass in an array until they look into code. But there is something changed for options for `Ky` constructor actually.
Maybe there could be some assertions for the `defaults` param type as it should be an object? :)

poppinlp:
Fix #40

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -193,7 +193,10 @@ test('ky.extend() throws when given non-object argument', t => {
 	for (const value of nonObjectValues) {
 		t.throws(() => {
 			ky.extend(value);
-		}, TypeError, 'The `defaultOptions` argument must be an object');
+		}, {
+			instanceOf: TypeError,
+			message: 'The `defaultOptions` argument must be an object'
+		});
 	}
 });
 
--- a/test/main.js
+++ b/test/main.js
@@ -191,11 +191,9 @@ test('ky.extend() throws when given non-object argument', t => {
 	];
 
 	for (const value of nonObjectValues) {
-		const error = t.throws(() => {
+		t.throws(() => {
 			ky.extend(value);
-		}, TypeError);
-
-		t.is(error.message, 'The `defaultOptions` argument must be an object');
+		}, TypeError, 'The `defaultOptions` argument must be an object');
 	}
 });
 
--- a/test/main.js
+++ b/test/main.js
@@ -180,7 +180,7 @@ test('ky.extend()', async t => {
 });
 
 test('ky.extend() throws when given non-object argument', t => {
-	[
+	const nonObjectValues = [
 		true,
 		666,
 		'hello',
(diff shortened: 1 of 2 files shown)


test/main.js
```diff
@@ -179,6 +179,22 @@ test('ky.extend()', async t => {
...
+		666,
+		'hello',
+		[],
+		null,
+		() => {},
+		Symbol('ky')
+	].forEach(value => {
+		t.throws(() => {
+			ky.extend(value);
+		});
```
sindresorhus: You should also assert that it throws the correct message and the type of the error. Use the object form of the `t.throws` assertions: https://github.com/avajs/ava#throwsasyncthrower-expected-message

test/main.js
```diff
@@ -179,6 +179,22 @@ test('ky.extend()', async t => {
...
+test('ky.extend() throws when given non-object argument', t => {
+	[
+		true,
+		666,
+		'hello',
+		[],
+		null,
+		() => {},
+		Symbol('ky')
+	].forEach(value => {
```
sindresorhus: Make the list of types a separate variable and use a `for-of` loop instead. (Just my preference. Makes it more readable)

poppinlp:
@sindresorhus Hi. I've pushed a commit for that. Thanks for your aborative review.

sindresorhus:
Like I commented in https://github.com/sindresorhus/ky/pull/50#discussion_r223930968, you can use the object form of the throws assertion so you don't have to use `t.is` to test the error message.

test/main.js
```diff
@@ -179,6 +179,24 @@ test('ky.extend()', async t => {
...
+		[],
+		null,
+		() => {},
+		Symbol('ky')
+	];
+
+	for (const value of nonObjectValues) {
+		t.throws(() => {
+			ky.extend(value);
+		}, TypeError, 'The `defaultOptions` argument must be an object');
```
sindresorhus: Still not the **object** form. Should be:

```js
t.throws(
	() => {
		ky.extend(value);
	}
	{
		instanceOf: TypeError,
		message: 'The `defaultOptions` argument must be an object'
	}
);
```

poppinlp:
Sorry that's my fault. I've seen the `.throws(fn, [expected, [message]])` and have not looked into the description of `expected` argument.

poppinlp:
By the way, maybe there could be an error helper which includes our all static errors. We could require it in code and using to throw errors in lib and do assertions in test cases. This will prevent the same error message appearing in two places which we must keep them same right now. And this may be helpful for i18n(not the key point).

sindresorhus:
Could make sense in the future if we end up with a lot of error messages, but for now, it feels like a lot of overhead for little gain. The messages won't change much anyway, so I think it's fine how it is.

sindresorhus:
Thanks :)

PR #52, opened 2018-10-10
TypeScript: Add prefixUrl to Options interface

prichodko:


prichodko:
Should've looked into opened PRs. 😅

PR #55, opened 2018-10-18
Respect the `Retry-After` header when calculating retry delay

### LINKED ISSUE #45: Respect Retry-After response header
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

komcal:
Fix #45

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,7 +68,6 @@ const retryAfterStatusCodes = new Set([
 class HTTPError extends Error {
 	constructor(response) {
 		super(response.statusText);
-		Object.setPrototypeOf(this, HTTPError.prototype);
 		this.name = 'HTTPError';
 		this.response = response;
 	}
--- a/index.js
+++ b/index.js
@@ -151,25 +151,34 @@ class Ky {
 			} catch (error) {
 				const shouldRetryStatusCode = error instanceof HTTPError ? retryStatusCodes.has(error.response.status) : true;
 				const retryAfterHeader = error.response && error.response.headers.get('Retry-After');
-				if (!(error instanceof TimeoutError) && shouldRetryStatusCode && this._retryCount < this._options.retry) {
-					this._retryCount++;
-					const BACKOFF_FACTOR = 0.3;
-					let delaySeconds = BACKOFF_FACTOR * (2 ** (this._retryCount - 1)) * 1000;
-					if (retryAfterHeader) {
-						const after = Number(retryAfterHeader);
-						if (Number.isNaN(after)) {
-							delaySeconds = Date.parse(retryAfterHeader) - Date.now();
-						} else {
-							delaySeconds = after * 1000;
-						}
+				if ((error instanceof TimeoutError) || this._retryCount >= this._options.retry || !shouldRetryStatusCode) {
+					if (this._throwHttpErrors) {
+						throw error;
+					}
+					return 0;
+				}
+				this._retryCount++;
+				let delaySeconds;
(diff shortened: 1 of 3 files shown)


sindresorhus:
Don't forget the special handling of status code 413 as commented in https://github.com/sindresorhus/got/pull/508#issuecomment-404196389.

komcal:
@sindresorhus I already handle special case of status code 413.

index.js
```diff
@@ -150,17 +150,35 @@ class Ky {
...
+				if (error.response.status === 413) {
+					if (this._throwHttpErrors) {
+						throw error;
+					}
+					return 0;
 				}
+				const BACKOFF_FACTOR = 0.3;
+				delaySeconds = BACKOFF_FACTOR * (2 ** (this._retryCount - 1)) * 1000;
+				await delay(delaySeconds);
+				return retry();
```
```js
const retryAfterStatusCodes = new Set([413, 429, 503]);
...

		const calculateRetryDelay = error => {
			if (this._retryCount < this._options.retry && !(error instanceof TimeoutError)) {
				if (error instanceof HTTPError) {
					if (!retryStatusCodes.has(error.response.status)) {
						return 0;
					}

					const retryAfter = error.response.headers.get('retry-after');
					if (retryAfter && retryAfterStatusCodes.has(error.response.status)) {
						let after = Number(retryAfter);
						if (Number.isNaN(after)) {
							after = Date.parse(retryAfter) - Date.now();
						} else {
							after *= 1000;
						}

						// TODO: maxRetryAfter

						return after;
					}

					if (error.response.status === 413) {
						return 0;
					}
				}

				const BACKOFF_FACTOR = 0.3;
				return 1000 * BACKOFF_FACTOR * (2 ** (this._retryCount - 1));
			}

			return 0;
		};

		const retry = async () => {
			try {
				return await fn();
			} catch (error) {
				const ms = calculateRetryDelay(error);
				if (ms !== 0) {
					await delay(ms);
					return retry();
				}

				if (this._throwHttpErrors) {
					throw error;
				}
			}
		};
```

szmarczak: It's more beautiful :)

index.js
```diff
@@ -59,10 +59,16 @@ const retryStatusCodes = new Set([
...
+const retryAfterStatusCodes = new Set([
+	413,
+	429,
+	503
+]);
 
 class HTTPError extends Error {
 	constructor(response) {
 		super(response.statusText);
+		Object.setPrototypeOf(this, HTTPError.prototype);
```
szmarczak: This line is useless

komcal: @szmarczak fixed. I forgot to remove it.

test/retry.js
```diff
@@ -4,6 +4,8 @@ import ky from '..';
 
 const fixture = 'fixture';
 const defaultRetryCount = 2;
+const retryAfterOn413 = 2;
+const lastTried413access = Date.now();
```
sindresorhus: This one is only used in one test, so should be defined inside that test.

komcal: @sindresorhus `retryAfterOn413` and `lastTried413access` used in two tests
https://github.com/sindresorhus/ky/blob/cec5ab882a5ce1c60edb8ed504de48c77c8b09d5/test/retry.js#L86-L107

https://github.com/sindresorhus/ky/blob/cec5ab882a5ce1c60edb8ed504de48c77c8b09d5/test/retry.js#L109-L131

sindresorhus:
I noticed you live in Thailand too. Let me know if you're ever in Bangkok and want to meet up for coffee or something.

PR #57, opened 2018-10-26
Add `searchParams` option

### LINKED ISSUE #54: Add simple query/search parameters option?
I've added a very simple wrapper around fetch myself and added the ability to passing in a `qs` object, which adds some parameters to the url:
```js
const uri = new URL(url)
if (qs) {
	Object.keys(qs).forEach((key) => uri.searchParams.append(key, qs[key]))
}
```

it's a nicer api then just adding these to the url by hand.
I could make a PR if interested.

itaisteinherz:
Fixes #54


index.js
```diff
@@ -86,13 +86,14 @@ const timeout = (promise, ms) => Promise.race([
 ]);
 
 class Ky {
-	constructor(input, {timeout = 10000, hooks = {beforeRequest: []}, throwHttpErrors = true, json, ...otherOptions}) {
+	constructor(input, {timeout = 10000, hooks = {beforeRequest: []}, throwHttpErrors = true, searchParams = '', json, ...otherOptions}) {
```
szmarczak: `searchParams,` (no empty string)

index.js
```diff
@@ -105,7 +106,12 @@ class Ky {
 			this._options.prefixUrl += '/';
 		}
 
-		this._input = this._options.prefixUrl + this._input;
+		const url = new URL(this._options.prefixUrl + this._input);
+		if (searchParams) {
+			url.search = new URLSearchParams(searchParams).toString();
```
szmarczak: Create a new URLSearchParams instance only if searchParams is an object or an array. Don't parse strings when you want a string.

index.js
```diff
@@ -86,13 +86,14 @@ const timeout = (promise, ms) => Promise.race([
...
 class Ky {
-	constructor(input, {timeout = 10000, hooks = {beforeRequest: []}, throwHttpErrors = true, json, ...otherOptions}) {
+	constructor(input, {timeout = 10000, hooks = {beforeRequest: []}, throwHttpErrors = true, searchParams = '', json, ...otherOptions}) {
 		this._retryCount = 0;
 
 		this._options = {
 			method: 'get',
 			credentials: 'same-origin', // TODO: This can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
 			retry: 2,
+			searchParams: new URLSearchParams(searchParams),
```
szmarczak: This is useless

itaisteinherz:
Thanks for the detailed review @szmarczak 👍🏻
I'm still not sure why the build fails, since `URLSearchParams` should be available on Node.js v8.

index.js
```diff
@@ -105,7 +105,12 @@ class Ky {
 			this._options.prefixUrl += '/';
 		}
 
-		this._input = this._options.prefixUrl + this._input;
+		const url = new URL(this._options.prefixUrl + this._input);
+		if (searchParams) {
+			url.search = searchParams instanceof URLSearchParams ? searchParams : new URLSearchParams(searchParams).toString();
```
szmarczak: What about:

```js
if (typeof searchParams === 'string') {
    url.search = searchParams;
} else {
    url.search = new URLSearchParams(searchParams).toString();
}
```

itaisteinherz: Your suggestion looks good to me, except for the fact that if `searchParams` is an instance of `URLSearchParams` you end up  initiating a clone of it for no reason.
We could do this:
```js
if (typeof searchParams === 'string' || searchParams instanceof URLSearchParams) {
    url.search = searchParams;
} else {
    url.search = new URLSearchParams(searchParams).toString();
}
```
Which both doesn't parse strings nor `URLSearchParams` objects.

szmarczak: Lemme fix your code:

```js
if (typeof searchParams === 'string') {
    url.search = searchParams;
} else if (searchParams instanceof URLSearchParams) {
    url.search = searchParams.toString();
} else {
    url.search = new URLSearchParams(searchParams).toString();
}
```

I hope it works as expected

itaisteinherz: We could do that, but from my testing in both Node and my browser (Chrome), ` url.search = searchParams` works.

This is how I tested it:

screenshot

itaisteinherz: In a flash ;)

itaisteinherz: I ended up doing:
```js
if (typeof searchParams === 'string' || searchParams instanceof URLSearchParams) {
	url.search = searchParams;
} else if (searchParams) {
	url.search = new URLSearchParams(searchParams).toString();
}
```

Since if the `searchParams` option isn't defined, the test `prefix-url › prefixUrl option` fails (which makes sense).

itaisteinherz:
@szmarczak Can you mark [#57 (comment)](https://github.com/sindresorhus/ky/pull/57/files#r228672170) as resolved? We've already addressed the issue in the discussion on https://github.com/sindresorhus/ky/pull/57#discussion_r228680986 and in 6bb65d0f6befabb8c1c46ae07035c041e2c2d549.

maple3142:
How about handling this one?

```js
ky('https://example.com/?a=1&b=2', { searchParams: { a: 2 } })
```

itaisteinherz:
@maple3142 Currently, all query parameters get ignored when the `searchParams` option is supplied, and so in your example the final URL which will be requested is https://example.com/?a=2.
This behavior was derived from https://github.com/sindresorhus/ky/issues/54#issuecomment-430855425.

sindresorhus:
And do anyone disagree that it should replace the search params already in the URL instead of merging them in?

index.js
```diff
@@ -108,7 +108,7 @@ class Ky {
 		const url = new URL(this._options.prefixUrl + this._input);
 		if (typeof searchParams === 'string' || searchParams instanceof URLSearchParams) {
 			url.search = searchParams;
-		} else if (searchParams) {
+		} else if (searchParams && Object.values(searchParams).every(param => typeof param === 'number' || typeof param === 'string')) {
 			url.search = new URLSearchParams(searchParams).toString();
 		}
```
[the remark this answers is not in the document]
itaisteinherz: Done ✅ Let me know if you'd like me to change anything else.

felixfbecker:
It should probably mirror what `new URLSearchParams(object)` does.

sindresorhus:
Can you also fix the merge conflict?

test/main.js
```diff
@@ -139,6 +139,26 @@ test('timeout option', async t => {
...
+
+	const stringParams = '?pass=true';
+	const objectParams = {pass: 'true'};
+	const invalidObjectParams = {pass: ['true', 'false']};
+	const searchParams = new URLSearchParams(stringParams);
+
+	t.is(await ky.get(server.url, {searchParams: stringParams}).text(), stringParams);
+	t.is(await ky.get(server.url, {searchParams: objectParams}).text(), stringParams);
+	t.is(await ky.get(server.url, {searchParams}).text(), stringParams);
+	t.is(await ky.get(server.url, {searchParams: invalidObjectParams}).text(), '');
```
sindresorhus: This should use `t.throws`

readme.md
```diff
@@ -172,6 +172,13 @@ Throw a `HTTPError` for error responses (non-2xx status codes).
 
 Setting this to `false` may be useful if you are checking for resource availability and are expecting error responses.
 
+##### searchParams
+
+Type: `string` `Object<string, string|number>` `URLSearchParams`<br>
+Default: `''`
+
+Search parameters to use in the HTTP request (setting this option will override any search params defined in the URL itself).
```
sindresorhus: > Search parameters to include in the request URL. Setting this will override all search parameters in the URL.

itaisteinherz:
> The build failure should be easy to fix now that the Cypress tests are up and running. :)

@sholladay I've never used Cypress before, and so I'm not sure how to fix the build failure.
Also, the build fails on Node.js v8 because of `URLSearchParams` not being available when testing using AVA. Am I supposed to port all AVA tests to Cypress, or just the `searchParams`-related ones (or something else)?

index.js
```diff
@@ -134,7 +135,16 @@ class Ky {
 			this._options.prefixUrl += '/';
 		}
 
-		this._input = this._options.prefixUrl + this._input;
+		const url = new URL(this._options.prefixUrl + this._input);
```
sindresorhus: `new URL` => `new _globalThis.URL`

sindresorhus:
```suggestion
		const url = new _globalThis.URL(this._options.prefixUrl + this._input);
```

index.js
```diff
@@ -134,7 +135,16 @@ class Ky {
...
+		if (typeof searchParams === 'string' || searchParams instanceof URLSearchParams) {
+			url.search = searchParams;
+		} else if (searchParams && Object.values(searchParams).every(param => typeof param === 'number' || typeof param === 'string')) {
+			url.search = new URLSearchParams(searchParams).toString();
```
sindresorhus:
```suggestion
			url.search = new _globalThis.URLSearchParams(searchParams).toString();
```

test/main.js
```diff
@@ -139,6 +139,26 @@ test('timeout option', async t => {
...
+	const server = await createTestServer();
+
+	server.get('/', (request, response) => {
+		response.end(request.url.slice(1));
+	});
+
+	const stringParams = '?pass=true';
+	const objectParams = {pass: 'true'};
+	const invalidObjectParams = {pass: ['true', 'false']};
+	const searchParams = new URLSearchParams(stringParams);
```
sindresorhus:
```suggestion
	const searchParams = new global.URLSearchParams(stringParams);
```

sindresorhus:
@itaisteinherz No, we'll only use Cypress when it's impossible to use Node.js testing. Accept my code suggestion above and then rebase from master branch and the test should pass ;)

itaisteinherz:
> Accept my code suggestion above and then rebase from master branch and the test should pass ;)

Done, but I still had to merge from master because I couldn't figure out how to rebase properly 😓

szmarczak:
LGTM.

sindresorhus:
This still needs TypeScript typings, like mentioned in https://github.com/sindresorhus/ky/pull/57#issuecomment-434877577. Otherwise the code changes are looking good.

itaisteinherz:
@sindresorhus I'll finish the TypeScript typing in an hour or so.

sindresorhus:
You missed https://github.com/sindresorhus/ky/pull/57#issuecomment-434927306, but I can fix that for you.

sindresorhus:
Thank you for working on this :)

Issue #58, opened 2018-11-04
TypeError relative URL after upgrading to 0.5.0


peterbe:
I've got this code:

```javascript
  fetchHomeCards = async () => {
    this.setState({ loading: true });
    const response = await ky("/api/cards/");
    if (response.ok) {
...
```
it used to work just fine in 0.4.1, but busted now. 

In Firefox I get `TypeError: /api/cards/ is not a valid URL.` on this line:
```javascript
var url = new _globalThis.URL(this._options.prefixUrl + this._input);
```

In Chrome I get a lot more helpful error:
```
index.js:181 Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at new Ky (index.js:181)
    at ky (index.js:518)
    at CardsContainer._callee$ (State.js:15)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:288)
    at Generator.prototype.(:3000/anonymous function) [as next] (http://localhost:3000/static/js/1.chunk.js:941:21)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
    at asyncToGenerator.js:32
    at new Promise (<anonymous>)
    at CardsContainer.fetchHomeCards (asyncToGenerator.js:21)
    at Home._callee$ (Home.js:9)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:288)
    at Generator.prototype.(:3000/anonymous function) [as next] (http://localhost:3000/static/js/1.chunk.js:941:21)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
    at asyncToGenerator.js:32
    at new Promise (<anonymous>)
    at Home.<anonymous> (asyncToGenerator.js:21)
    at Home.componentDidMount (Home.js:12)
    at commitLifeCycles (react-dom.development.js:15986)
    at commitAllLifeCycles (react-dom.development.js:17388)
    at HTMLUnknownElement.callCallback (react-dom.development.js:147)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:196)
    at invokeGuardedCallback (react-dom.development.js:250)
    at commitRoot (react-dom.development.js:17549)
    at completeRoot (react-dom.development.js:19002)
    at performWorkOnRoot (react-dom.development.js:18924)
    at performWork (react-dom.development.js:18826)
[...]

PR #59, opened 2018-11-06
Fix relative URL resolution

sholladay:
Fixes https://github.com/sindresorhus/ky/issues/58

There was a regression introduced in https://github.com/sindresorhus/ky/commit/a7c71b5728adb5efab59f4057caf83a82453ba4d which broke the ability for `input` to be a relative URL like `foo` or `/foo`. The root cause is that [`window.URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) unfortunately does not support relative URLs without an explicit and absolute `base` argument. This fixes that and adds a test for it. Now that we have Cypress set up, we can properly test anything related to relative URLs.

The goal is to follow the URL resolution rules the same way that `fetch` does. So we resolve `input` against [`document.baseURI`](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI#The_base_URL_of_a_document).

### DIFF
--- a/cypress/integration/index.js
+++ b/cypress/integration/index.js
@@ -6,6 +6,7 @@ describe('ky', () => {
 			ky('/foo', {prefixUrl: '/'});
 		}).to.throw(Error, /must not begin with a slash/);
 	});
+
 	it('resolves relative URLs for `input` and `prefixUrl`', async () => {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for fixing this ❤️

PR #63, opened 2018-11-21
Normalize `method` option value casing

Seldszar:
According to [Fetch specifications](https://fetch.spec.whatwg.org/#concept-method-normalize), method name normalization is currently applied to all but `PATCH`. I personally use this library but in this case, I must override the `method` option in order to fix the issue.

This proposal normalizes shorthand `method` option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,7 +120,7 @@ class Ky {
 		this._retryCount = 0;
 
 		this._options = {
-			method: 'get',
+			method: 'GET',
 			credentials: 'same-origin', // TODO: This can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
 			retry: 2,
 			...otherOptions
@@ -266,7 +266,9 @@ const createInstance = (defaults = {}) => {
 	const ky = (input, options) => new Ky(input, deepMerge({}, defaults, options));
 
 	for (const method of requestMethods) {
-		ky[method] = (input, options) => new Ky(input, deepMerge({}, defaults, options, {method}));
+		ky[method] = (input, options) => new Ky(input, deepMerge({}, defaults, options, {
+			method: method.toUpperCase()
+		}));
 	}
 
 	ky.extend = defaults => createInstance(defaults);
--- a/index.js
+++ b/index.js
@@ -108,13 +108,7 @@ const timeout = (promise, ms) => Promise.race([
 	})()
 ]);
 
-const normalizeRequestMethod = input => {
-	if (requestMethods.includes(input)) {
-		return input.toUpperCase();
-	}
-
-	return input;
-};
+const normalizeRequestMethod = input => requestMethods.includes(input) ? input.toUpperCase() : input;
 
 class Ky {
 	constructor(input, {
--- a/index.js
+++ b/index.js
@@ -108,6 +108,14 @@ const timeout = (promise, ms) => Promise.race([
 	})()
 ]);
 
+const normalizeRequestMethod = input => {
+	if (requestMethods.includes(input)) {
+		return input.toUpperCase();
+	}
+
+	return input;
+};
+
 class Ky {
 	constructor(input, {
 		timeout = 10000,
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -266,7 +266,9 @@ const createInstance = (defaults = {}) => {
 	const ky = (input, options) => new Ky(input, deepMerge({}, defaults, options));
 
 	for (const method of requestMethods) {
-		ky[method] = (input, options) => new Ky(input, deepMerge({}, defaults, options, {method}));
+		ky[method] = (input, options) => new Ky(input, deepMerge({}, defaults, options, {
+			method: method.toUpperCase()
+		}));
```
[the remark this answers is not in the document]
Seldszar: I can, but I was thinking about the case where the method was a custom & lowercase one. But if you prefer this way, I can change it.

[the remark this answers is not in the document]
sindresorhus: I'm not sure I understand. Having it in the constructor should handle all cases from what I can see, although I might be missing something. Would you be able to provide a code example of what you're thinking or even better a test showing the use-case?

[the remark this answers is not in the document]
Seldszar: Oh, I think I misunderstood, my bad. I'll add  I'll do the change right away and include the tests.

sindresorhus:
Can you also document the behavior? It's useful for users to know that they can just use lowercase values for the `method` option.

Seldszar:
Is this explanation good enough? I'm not really good at this to be honest.

sindresorhus:
Tests are failing: https://travis-ci.com/sindresorhus/ky/jobs/164824107

readme.md
```diff
@@ -266,6 +266,10 @@ setTimeout(() => controller.abort(), 5000);
 })();
 ```
 
+### Method name normalization
+
+Ky normalizes the common used method names (`GET`, `POST`, `PATCH`, etc.) in order to avoid server errors due to case sensitivity.
```
sindresorhus: I was thinking just documenting the `method` option here: https://github.com/sindresorhus/ky#options And mentioning the normalization we do.

readme.md
```diff
@@ -104,6 +104,15 @@ Sets `options.method` to the method name and makes a request.
 
 Type: `Object`
 
+##### method
+
+Type: `String`
+Default: `'GET'`
+
+Sets the method name and makes the request.
```
sindresorhus: This is not correct. It doesn't make the request. And "sets the method name" is not really more descriptive than the option name.

It could say:

> Set the HTTP request method.

readme.md
```diff
@@ -104,6 +104,15 @@ Sets `options.method` to the method name and makes a request.
 
 Type: `Object`
 
+##### method
+
+Type: `String`
+Default: `'GET'`
```
sindresorhus: Since we normalize anyway, we can use `get` here to show users clearly that they can use lowercase.

readme.md
```diff
@@ -104,6 +104,15 @@ Sets `options.method` to the method name and makes a request.
 
 Type: `Object`
 
+##### method
+
+Type: `String`
```
sindresorhus: `String` => `string`

readme.md
szmarczak: Also I think it should be: `'GET'` => `GET`

index.js
```diff
@@ -108,6 +108,14 @@ const timeout = (promise, ms) => Promise.race([
 	})()
 ]);
 
+const normalizeRequestMethod = input => {
+	if (requestMethods.includes(input)) {
+		return input.toUpperCase();
+	}
+
+	return input;
+};
```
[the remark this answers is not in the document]
szmarczak: Shouldn't it be

```js
const normalizeRequestMethod = input => input.toUpperCase();
```

sindresorhus:
> Default: GET

https://github.com/sindresorhus/ky/pull/63#discussion_r242546976

> Ky transforms the method to uppercase to keep things consistent.

This is slightly misleading as it only normalizes the standard HTTP methods. It leaves custom ones alone.

PR #67, opened 2018-11-29
ky.extend prefixUrl usage error fixed

### LINKED ISSUE #66: ky.extend prefixUrl usage error
[](https://postimg.cc/625Z24mp)

with this usage I get the above error

```js
const SERVICE_API = 'http://localhost:4000';

const api = ky.extend({prefixUrl: SERVICE_API});

const json = await api.post(`/connect`, {json: params}).json();
```

indatawetrust:
#66

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -262,3 +262,25 @@ test('ky.extend() with deep array', async t => {
 
 	await server.close();
 });
+
+test('ky.extend() with prefix url', async t => {
+	const server = await createTestServer();
+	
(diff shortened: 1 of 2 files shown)


test/main.js
```diff
@@ -262,3 +262,25 @@ test('ky.extend() with deep array', async t => {
 
 	await server.close();
 });
+
+test('ky.extend() with prefix url', async t => {
```
sindresorhus: This test passes even without your changes.

PR #69, opened 2018-12-17
Handle non-DOM environments better

### LINKED ISSUE #64: React native reference error in 0.5.1
I have been using this library on version 0.4.1 for some time and that works in react-native when running on the simulator with a debugger attached. Running on device results in Reference Error: self is not defined. This should be fixed in issue #53 (With commit f107e7b92d90e4b0287846a49ba3f30d9b9ca648) which is available from version 0.5.0. After upgrading ky to 0.5.0 or 0.5.1 Running on a simulator with a debugger attached or a device results in Reference Error: document is not defined.

I import ky and export an instance as follows:
```
import ky from "ky";
export const api = ky.extend({prefixUrl: BASE_URL});
```

The error shown in the console is as follows:
```
Possible Unhandled Promise Rejection (id: 0):
ReferenceError: document is not defined
ReferenceError: document is not defined
    at new Ky (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:111478:76)
    at Function.ky.(anonymous function) [as post] (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:111772:16)
    at _callee3$ (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:115504:24)
    at tryCatch (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:2081:19)
    at Generator.invoke [as _invoke] (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:2256:24)
    at Generator.prototype.(anonymous function) [as next] (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:2124:23)
    at tryCatch (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:2081:19)
    at invoke (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:2157:22)
    at blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:2167:15
    at tryCallOne (blob:file:///58581b4b-dc99-4126-b9a9-cb3f918e33f0:6115:14)
```

felixjung:
I've tried to address the problem described in the issue. I left some comments in the diff, which I'm eager to get feedback on.

Fixes #64

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,10 +110,13 @@ const timeout = (promise, ms) => Promise.race([
 
 const normalizeRequestMethod = input => requestMethods.includes(input) ? input.toUpperCase() : input;
 
-const isAbsoluteUrl = str => /^[a-z]+:\/\//.test(str);
+const isAbsoluteUrl = input => {
+	const url = typeof input === 'string' ? input : input.toString();
+	return /^[a-z]+:\/\//.test(url);
+};
 
 const createRequestUrl = (prefixUrl, input) => {
-	if (!prefixUrl) {
+	if (!prefixUrl || isAbsoluteUrl(input)) {
 		return input;
 	}
 
--- a/index.js
+++ b/index.js
@@ -124,7 +124,7 @@ const createRequestUrl = (prefixUrl, input) => {
 		throw new Error('`input` must not begin with a slash when using `prefixUrl`');
 	}
 
-	const separator = /\/$/.test(prefixUrl) ? '' : '/';
+	const separator = prefixUrl.endsWith('/') ? '' : '/';
 
 	return `${prefixUrl}${separator}${input}`;
 };
--- a/index.js
+++ b/index.js
@@ -110,6 +110,22 @@ const timeout = (promise, ms) => Promise.race([
 
 const normalizeRequestMethod = input => requestMethods.includes(input) ? input.toUpperCase() : input;
 
+const isAbsoluteUrl = str => /^[a-z]+:\/\//.test(str);
+
+const createRequestUrl = (prefixUrl, input) => {
+	if (!prefixUrl) {
+		return input;
+	}
+
+	if (prefixUrl && input.startsWith('/')) {
+		throw new Error('`input` must not begin with a slash when using `prefixUrl`');
+	}
+
+	const separator = /\/$/.test(prefixUrl) ? '' : '/';
+
+	return `${prefixUrl}${separator}${input}`;
+};
+
 class Ky {
 	constructor(input, {
 		timeout = 10000,
@@ -131,14 +147,17 @@ class Ky {
 		this._options.prefixUrl = String(this._options.prefixUrl || '');
 		this._input = String(input || '');
 
-		if (this._options.prefixUrl && this._input.startsWith('/')) {
-			throw new Error('`input` must not begin with a slash when using `prefixUrl`');
-		}
-		if (this._options.prefixUrl && !this._options.prefixUrl.endsWith('/')) {
-			this._options.prefixUrl += '/';
+		const isFullUrlInput = isAbsoluteUrl(this._input);
+		const isFullUrlPrefix = isAbsoluteUrl(this._options.prefixUrl);
+		const documentBaseUri = _globalThis.document && _globalThis.document.baseURI;
+
+		if (!isFullUrlInput && !isFullUrlPrefix && !documentBaseUri) {
+			throw new Error('The `prefixUrl` option must be set to a full URL, incl. protocol, when using a relative `input` URL inside a non-browser environment');
 		}
 
-		const url = new _globalThis.URL(this._options.prefixUrl + this._input, document.baseURI);
+		const urlString = createRequestUrl(this._options.prefixUrl, this._input);
+		const url = new _globalThis.URL(urlString, documentBaseUri);
+
 		if (typeof searchParams === 'string' || searchParams instanceof _globalThis.URLSearchParams) {
 			url.search = searchParams;
 		} else if (searchParams && Object.values(searchParams).every(param => typeof param === 'number' || typeof param === 'string')) {
--- a/readme.md
+++ b/readme.md
@@ -130,7 +130,7 @@ Search parameters to include in the request URL. Setting this will override all 
 
 Type: `string` [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL)
 
-When specified, `prefixUrl` will be prepended to `input`. The prefix can be any valid URL, either relative or absolute. A trailing slash `/` is optional, one will be added automatically, if needed, when joining `prefixUrl` and `input`. The `input` argument cannot start with a `/` when using this option.
+When specified, `prefixUrl` will be prepended to `input`. The prefix can be any valid URL, either relative or absolute. A trailing slash `/` is optional, one will be added automatically, if needed, when joining `prefixUrl` and `input`. The `input` argument cannot start with a `/` when using this option. If the `input` argument is an absolute URL (i.e., https://github.com), the `prefixUrl` is ignored.
 
 Useful when used with [`ky.extend()`](#kyextenddefaultoptions) to create niche-specific Ky-instances.
 
--- a/readme.md
+++ b/readme.md
@@ -290,7 +290,7 @@ See my answer [here](https://twitter.com/sindresorhus/status/1037763588826398720
 
 See my answer in [#10](https://github.com/sindresorhus/ky/issues/10).
 
-#### Using ky in non-browser environments
+#### Using `ky` in non-browser environments
 
 If you plan to use ky in non-browser environments, for example when rendering React on the server, you need to polyfill certain things on the global object, specifically, `fetch` and the `Headers` constructor. Below, you find an example of an `http` module used throughout an app. It relies on the `node-fetch` module for polyfills.
 
--- a/readme.md
+++ b/readme.md
@@ -294,19 +294,19 @@ See my answer in [#10](https://github.com/sindresorhus/ky/issues/10).
 
 If you plan to use ky in non-browser environments, for example when rendering React on the server, you need to polyfill certain things on the global object, specifically, `fetch` and the `Headers` constructor. Below, you find an example of an `http` module used throughout an app. It relies on the `node-fetch` module for polyfills.
 
-```javascript
-import ky from 'ky'
-import fetch, { Headers } from 'node-fetch'
+```js
+import ky from 'ky';
+import fetch, {Headers} from 'node-fetch';
 
 if (typeof window === 'undefined') {
-  global.fetch = fetch
-  global.Headers = Headers
+	global.fetch = fetch;
+	global.Headers = Headers;
 }
 
-// Here you could also extend ky with a default prefixUrl for your app
-const http = ky
+// Here you could also extend Ky with a default `prefixUrl` for your app
+const http = ky;
 
-export default ky
+export default ky;
 ```
 
 #### What does `ky` mean?
--- a/readme.md
+++ b/readme.md
@@ -290,6 +290,25 @@ See my answer [here](https://twitter.com/sindresorhus/status/1037763588826398720
 
 See my answer in [#10](https://github.com/sindresorhus/ky/issues/10).
 
+#### Using ky in non-browser environments
+
+If you plan to use ky in non-browser environments, for example when rendering React on the server, you need to polyfill certain things on the global object, specifically, `fetch` and the `Headers` constructor. Below, you find an example of an `http` module used throughout an app. It relies on the `node-fetch` module for polyfills.
+
+```javascript
+import ky from 'ky'
+import fetch, { Headers } from 'node-fetch'
+
+if (typeof window === 'undefined') {
+  global.fetch = fetch
+  global.Headers = Headers
+}
+
+// Here you could also extend ky with a default prefixUrl for your app
+const http = ky
+
+export default ky
+```
+
 #### What does `ky` mean?
 
 It's just a random short npm package name I managed to get. It does, however, have a meaning in Japanese:
--- a/test/_require.js
+++ b/test/_require.js
@@ -6,6 +6,3 @@ global.Headers = Headers;
 global.Response = Response;
 global.URL = URL;
 global.URLSearchParams = URLSearchParams;
-global.document = {
-	baseURI: 'https://example.com'
-};
--- a/test/prefix-url.js
+++ b/test/prefix-url.js
@@ -5,7 +5,7 @@ import ky from '..';
 const createServer = async () => {
 	const server = await createTestServer();
 	server.get('/', (request, response) => {
-		if (request.query.page === '/https://cat.com/') {
+		if (request.query.page === '/cat') {
 			response.end('meow');
 			return;
 		}
@@ -52,7 +52,7 @@ test('prefixUrl can be the full request URL', async t => {
 });
 
 test('prefixUrl can be a URL object', async t => {
-	t.is(await ky('https://cat.com/', {prefixUrl: new URL(`${t.context.server.url}/?page=`)}).text(), 'meow');
+	t.is(await ky('cat', {prefixUrl: new URL(`${t.context.server.url}/?page=`)}).text(), 'meow');
 });
 
 test('when prefixUrl is specified, `input` cannot have a leading`/`', t => {
@@ -65,7 +65,11 @@ test('when prefixUrl is specified, `input` cannot have a leading`/`', t => {
 	}, '`input` must not begin with a slash when using `prefixUrl`');
 });
 
-test('prefixUrl cannot be relative when `input` is relative and run on a server', t => {
+test('when `input` is an absolute URL, it takes precedence over prefixUrl', async t => {
+	t.is(await ky(`${t.context.server.url}/api/unicorn`, {prefixUrl: `${t.context.server.url}/api`}).text(), 'rainbow');
+});
+
+test('prefixUrl cannot be relative when `input` is relative and run in non-browser environments', t => {
 	delete global.document;
 
 	t.throws(() => {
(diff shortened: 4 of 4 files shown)


index.js
```diff
@@ -108,6 +108,22 @@ const timeout = (promise, ms) => Promise.race([
...
 ]);
 
+const isAbsoluteUrl = str => /^[a-z]+:\/\//.test(str);
+
+const createRequestUrl = (prefixUrl, input) => {
+	if (!prefixUrl) {
+		return input;
+	}
+
+	if (prefixUrl && input.startsWith('/')) {
```
felixjung: I somewhat feel that this exception is a bit arbitrary and tied to a very specific use case of passing a URL as query parameter. I also found this in the tests as something like

```
key.get('https://bar.com', { prefix: 'https://foo.com?redirect=' })
```

If a user decides to use `ky` in such a way, I don't see why the library should handle exceptions for this. I'd lean on the other end and allow users to do any of the following:

```
key.get('users', { prefix: 'https://foo.com/api' })
key.get('/users', { prefix: 'https://foo.com/api' })
key.get('/users', { prefix: 'https://foo.com/api/' })
```

These all seem like more relevant convenience use cases than preventing users from breaking the above-mentioned query case example.

index.js
```diff
@@ -128,14 +144,17 @@ class Ky {
 		this._options.prefixUrl = String(this._options.prefixUrl || '');
 		this._input = String(input || '');
 
-		if (this._options.prefixUrl && this._input.startsWith('/')) {
-			throw new Error('`input` must not begin with a slash when using `prefixUrl`');
-		}
-		if (this._options.prefixUrl && !this._options.prefixUrl.endsWith('/')) {
-			this._options.prefixUrl += '/';
+		const isFullUrlInput = isAbsoluteUrl(this._input);
```
felixjung: I found this whole part a bit tricky. When do you consider the input relative? Some examples:

- `input` starts with `/`. This is relative.
- `input` starts with `/+`, what now? I guess it's relative and could be accepted.
- `input` starts with `[a-z]`. At this point you have to start thinking about the `input` being a full url including a protocol. How to handle this?

index.js
```diff
@@ -128,14 +144,17 @@ class Ky {
...
+		const isFullUrlPrefix = isAbsoluteUrl(this._options.prefixUrl);
+		const documentBaseUri = _globalThis.document && _globalThis.document.baseURI;
+
+		if (!isFullUrlInput && !isFullUrlPrefix && !documentBaseUri) {
```
felixjung: I think we have to check all three here.

- If `input` is a full URL (i.e., `https://foo.com`), we don't need a prefix and we don't need `document.baseURI`.
- If `input` is relative and `prefixUrl` is a full URL (i.e., `https://foo.com`), we also don't need `document.baseURI`.
- If neither `input` nor `prefixUrl` are full URLs, then we need `document.baseURI` to be present. If it's not present, we throw an error.

Here I was wondering how to handle the case where a user passes full URLs for both `input` and `prefixUrl`. I was considering throwing an error in this case, but then saw the example of using `input` to set a query parameter in a URL.

test/_require.js
```diff
@@ -6,6 +6,3 @@ global.Headers = Headers;
 global.Response = Response;
 global.URL = URL;
 global.URLSearchParams = URLSearchParams;
-global.document = {
```
felixjung: Removed the mock, because `ky` can handle this now.

index.js
```diff
@@ -108,6 +108,22 @@ const timeout = (promise, ms) => Promise.race([
 	})()
 ]);
 
+const isAbsoluteUrl = str => /^[a-z]+:\/\//.test(str);
```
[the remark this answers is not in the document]
szmarczak: I think it's OK.

@sindresorhus I just wonder what we should do about `//example.com` and `example.com`.

index.js
szmarczak: I agree, but many people would be confused.

`/` means root of the tree. But in that case the root is different, it's hard to choose.

index.js
szmarczak: Maybe there should be an option to pass custom `baseURI`?

index.js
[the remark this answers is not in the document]
sindresorhus: @szmarczak In contrast to Got, Ky doesn't support URL without a scheme, like `example.com`.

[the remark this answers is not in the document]
szmarczak: IKR. Anyway, try `<a href="//example.com">//example.com</a>` - it will use the current protocol. IMO it'd be very useful.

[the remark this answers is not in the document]
sindresorhus: @szmarczak While I agree we need to decide on how to handle it, you should not use it yourself. It's an anti-pattern: https://www.paulirish.com/2010/the-protocol-relative-url/

sindresorhus:
> I'd like to also add something regarding SSR to the README. You'll still have to mock the Headers constructor together with fetch for ky to work. Should this go in the global README or be moved somewhere to the docs?

We can add it to the readme for now. I'll move it around in the future if the readme becomes too large.

index.js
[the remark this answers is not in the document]
szmarczak: Good point. Let's keep it that way :)

index.js
felixjung: Cool.

felixjung:
Thanks for the feedback. I'll try to take some time this week to review and act where needed -- also about the README.

index.js
```diff
@@ -108,6 +108,22 @@ const timeout = (promise, ms) => Promise.race([
...
+const createRequestUrl = (prefixUrl, input) => {
+	if (!prefixUrl) {
+		return input;
+	}
+
+	if (prefixUrl && input.startsWith('/')) {
+		throw new Error('`input` must not begin with a slash when using `prefixUrl`');
+	}
+
+	const separator = /\/$/.test(prefixUrl) ? '' : '/';
```
sholladay: Why not keep `.endsWith()`? Much more readable IMO.

index.js
sholladay: > If input is absolute, it should take priority over prefixUrl

I mentioned this briefly in #11. I tried to characterize it as _partially implemented_ URL resolution, which I find to be somewhat confusing behavior. It means the option neither strictly behaves like a prefix, nor like any other link or base URL on the web platform. But of course, it can be documented as such. I just like the simplicity of the prefix behavior. It's easy to reason about.

> I was considering throwing an error in this case, but then saw the example of using input to set a query parameter in a URL.

I'd be fine with that, personally. My intention with those tests that have `prefixUrl` ending in a query string was to make it obvious that `ky` isn't just naively joining the URLs with a `/`. I doubt many people are actually using a query string prefix. The point is that semantically, they could, if they needed to for some reason. This would be a breaking change, at least in edge cases like that.

A bigger issue, in my mind is that this feels surprising:

```js
ky('http://foo.com', {prefixUrl: '/api'});  // prefixUrl is silently ignored
ky('http://foo.com', {prefixUrl: 'http://bar.com'});  // Error is thrown
```

I don't think we should do that. I want it to either throw in both cases, or in neither.

test/prefix-url.js
```diff
@@ -11,24 +11,64 @@ test('prefixUrl option', async t => {
 		}
 		response.end('zebra');
 	});
+
 	server.get('/api/unicorn', (request, response) => {
 		response.end('rainbow');
 	});
 
-	t.is(await ky(`${server.url}/api/unicorn`, {prefixUrl: false}).text(), 'rainbow');
```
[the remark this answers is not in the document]
felixjung: I agree and added the tests back.

> Could you explain which assertions were hard to satisfy? That would give me some insight into what might break in the wild and whether it matters.

I didn't find them hard to satisfy. What I meant was that, while refactoring, I couldn't figure out which `t.is` actually failed. Does that answer your question? Maybe I didn't get it. 😅

felixjung:
Hey all. I've rebased with master and I think I addressed all your comments. Sorry for the delay. Could you give it another look? Thanks!

LoicMahieu:
Hi!
Thanks for your works on this topic!

- Require to manual patch `global` object is annoying. If you use `ky` in a library, you should warn your users to do this patch. Also it could introduce unnecessary complexity in tools like `jest` (as library consumer). 
If we use `isomorphic-fetch`, it solves the problem naturally. Is that could be a solution ?

- `ky` is published as ESM. Is there a way to easily use it in "vanilla" nodejs ? I come across several issues to use it in `jest`. In order to make it working, I should transform it with babel (in a TS project... 🙄)

- Small issue, the minimal supported nodejs version is `v10.0.0` since it depends on `global.URL`. It could be useful to specify it in the `engine` field of the package.json.

sholladay:
@felixjung I just opened a PR that is sort of an alternative to this, based on my own vision of how libraries should handle globals. You can check it out in #74.

I am hoping that we can merge our efforts somehow and keep the best of each. Would love to hear your thoughts.

sindresorhus:
> Require to manual patch global object is annoying. If you use ky in a library, you should warn your users to do this patch. Also it could introduce unnecessary complexity in tools like jest (as library consumer).

We require users to patch globals themselves as libraries should not modify the global environment. We could consider adding a `globals` option or something where you could pass it in.

Maybe:

```js
import ky from 'ky';
import fetch, {Headers, Response} from 'node-fetch';

(async () => {
	const json = await ky('https://example.com', {globals: {fetch, Headers, Response}}).json();
})();
```

@felixjung @sholladay @LoicMahieu Thoughts?

> ky is published as ESM. Is there a way to easily use it in "vanilla" nodejs ? I come across several issues to use it in jest. In order to make it working, I should transform it with babel (in a TS project... 🙄)

Just needs a PR. See: https://github.com/sindresorhus/ky/pull/46#issuecomment-432546233

> Small issue, the minimal supported nodejs version is v10.0.0 since it depends on global.URL. It could be useful to specify it in the engine field of the package.json.

Fixed.

sholladay:
I'm lightly in favor of a `globals` option. But I think we should try to address these issues automatically first and see how that goes. The `globals` option may turn out to not be very necessary. I don't see much harm in adding it, either, though, other than the usual caveats of adding another option.

sindresorhus:
> But I think we should try to address these issues automatically first and see how that goes. The globals option may turn out to not be very necessary.

That would definitely be my preference.

readme.md
```diff
@@ -290,6 +290,25 @@ See my answer [here](https://twitter.com/sindresorhus/status/1037763588826398720
...
+import ky from 'ky';
+import fetch, {Headers} from 'node-fetch';
+
+if (typeof window === 'undefined') {
+	global.fetch = fetch;
+	global.Headers = Headers;
+}
+
+// Here you could also extend Ky with a default `prefixUrl` for your app
+const http = ky;
```
sindresorhus: This is confusing when you're not actually using the `http` variable or showing off usage of `prefixUrl`.

readme.md
```diff
@@ -290,6 +290,25 @@ See my answer [here](https://twitter.com/sindresorhus/status/1037763588826398720
 
 See my answer in [#10](https://github.com/sindresorhus/ky/issues/10).
 
+#### Using `ky` in non-browser environments
+
+If you plan to use ky in non-browser environments, for example when rendering React on the server, you need to polyfill certain things on the global object, specifically, `fetch` and the `Headers` constructor. Below, you find an example of an `http` module used throughout an app. It relies on the `node-fetch` module for polyfills.
+
+```js
+import ky from 'ky';
+import fetch, {Headers} from 'node-fetch';
```
sindresorhus: You're missing the `Response` global.

sholladay:
> import 'node-fetch/polyfill';

I can see that being convenient. 👍 

As long it doesn't overwrite APIs that are already defined, I might use it. Though, I do like explicitly polyfilling only what I need.

> ky.polyfillFetchIfNeeded();

Is the goal of this to help use ky or more generally use things that might need fetch? If the former, I don't see how it's an improvement over internal references within ky. If the latter, how would this work without shipping ky with a fetch implementation? Also, would this only be for fetch? Presumably you mean all globals that ky needs?

sindresorhus:
> Is the goal of this to help use ky or more generally use things that might need fetch? If the former, I don't see how it's an improvement over internal references within ky. If the latter, how would this work without shipping ky with a fetch implementation? Also, would this only be for fetch? Presumably you mean all globals that ky needs?

This could indeed use internal references. When `ky.polyfillFetchIfNeeded();` is called, it could try to import `node-fetch` and use it. So the user would of course need a dependency on `node-fetch` too, so it would only really work when using Ky from npm, since we're taking advantage of require traversing upwards to find dependencies, which means we can use the `node-fetch` that the user has a direct dependency on.

sholladay:
Oh, I see. So a `peerDependency`, essentially - aside from whether we declare it in package.json or just in documentation for that one function.

I like the concept, generally. My main concern would be that, even if we inlined the `require()` within the polyfill function, it would take a smart bundler to avoid shipping `node-fetch` to the browser. For those whom treeshaking failed for any reason, `node-fetch` production install is 168KB _before_ browserifying its Node.js core dependencies: `https`, `zlib`, and others. This might be trading one problem for another one. I'm not sure how well treeshaking is working these days, last time I really looked into it, it was hit or miss in my own apps.

Would also be a bit strange to use `require()` inside of an ES module. Though, i suppose it's perfectly valid, technically.

readme.md
felixjung: Hm, interesting. I didn't find notice any runtime errors in my application when not polyfilling this.

felixjung:
When using the code from this PR in my app, it actually ended up not working. Somehow, requests made on the server ended up failing but worked just fine on the client. I ended up switching to Axios (time constraints), which worked out of the box. So, I guess there is some work to do.

As to your suggestions, I think being explicit about using a polyfill and not relying on side-effects of importing `node-fetch` is preferable. So I'd be in favor of this approach

```javascript
import ky from 'ky';
import fetch, {Headers, Response} from 'node-fetch';

(async () => {
	const json = await ky('https://example.com', {globals: {fetch, Headers, Response}}).json();
})();
```

combined with some solid documentation. The approach of providing `ky.polyfillFetchIfNeeded();` kind of has the disadvantage of forcing a specific fetch polyfill on the user, no? Unless, you pass in the polyfills you want to use. But that's like the `globals` option.

sindresorhus:
> So a peerDependency, essentially - aside from whether we declare it in package.json or just in documentation for that one function.

Yes, although more like an indirect dependency.

> I like the concept, generally. My main concern would be that, even if we inlined the require() within the polyfill function, it would take a smart bundler to avoid shipping node-fetch to the browser.

There are ways to trick bundlers to ignore the `require` call.

For Webpack-only, we could do:

```js
const originalRequire = typeof __non_webpack_require__ === 'function' ? __non_webpack_require__ : require;
const fetch = originalRequire('node-fetch');
```

Or more generally:

```js
const originalRequire = global['require'];
const fetch = originalRequire('node-fetch');
```

Although, I think the `node-fetch/polyfill` way would be much better.

sindresorhus:
> The approach of providing ky.polyfillFetchIfNeeded(); kind of has the disadvantage of forcing a specific fetch polyfill on the user, no? Unless, you pass in the polyfills you want to use.

Yes, which is why I think this belongs in the polyfill library itself instead: 
`import 'node-fetch/polyfill';`

sholladay:
@felixjung thoughts on how to proceed? Are you interested in doing a PR to `node-fetch`? Should we keep this open? I think our non-DOM story is in pretty good shape now, except for polyfills. Could definitely use some help there.

index.js
```diff
@@ -131,14 +150,17 @@ class Ky {
...
+		const isFullUrlInput = isAbsoluteUrl(this._input);
+		const isFullUrlPrefix = isAbsoluteUrl(this._options.prefixUrl);
+		const documentBaseUri = _globalThis.document && _globalThis.document.baseURI;
+
+		if (!isFullUrlInput && !isFullUrlPrefix && !documentBaseUri) {
+			throw new Error('The `prefixUrl` option must be set to a full URL, incl. protocol, when using a relative `input` URL inside a non-browser environment');
 		}
 
-		const url = new _globalThis.URL(this._options.prefixUrl + this._input, document.baseURI);
+		const urlString = createRequestUrl(this._options.prefixUrl, this._input);
```
szmarczak: Why cannot we just `new URL(this._input, this._options.prefixUrl)`?

sholladay: @szmarczak if we did, then `prefixUrl` would have to be an absolute URL, which is undesirable. I'm really quite annoyed that `new URL()` doesn't support relative URLs. I recommend avoiding `new URL()` when the desired behavior is to follow browser resolution rules. As of #74, we conditionally use it only if we need it.

szmarczak: > if we did, then prefixUrl would have to be an absolute URL, which is undesirable.

Umm... But we do: https://github.com/sindresorhus/ky/blob/d5d3eaa24f903294d1b86d96a7652262c61536eb/index.js#L145-L151

>  I recommend avoiding new URL() when the desired behavior is to follow browser resolution rules. As of #74, we conditionally use it only if we need it.

I mean `_globalThis.URL` :P

szmarczak: Is there any particular reason for `createRequestUrl`? Why we can't leave it as it is?

felixjung:
Hi, Sorry for the lack of communication. I'm quite busy at the moment.

> I think our non-DOM story is in pretty good shape now, except for polyfills.

Do you mean with this PR or on master (have not checked in a while). If the polyfill is what's missing, then I don't think I would open a PR to node-fetch, myself. Feel free to close my PR then.

If the PR is still needed to improve the SSR story of ky, I'd be happy to help get it over the finish line.

sindresorhus:
I think what's left is to open an issue/PR on `node-fetch`.

sholladay:
> Do you mean with this PR or on master

On master. I used this PR as inspiration and made some different fixes that should make it even easier to use Ky in Node, Deno, web workers, and more unusual environments. It's been released, so if you have a chance, try it out and let me know how well it works for you.

Closing, since we should pursue polyfills separately. I have a todo item to look at the tests again and maybe cherry pick some of those changes.

felixjung:
👌

sindresorhus:
I published a new package which should make it easy to use Ky in Node.js or isomorphically: https://github.com/sindresorhus/ky-universal

PR #74, opened 2019-01-03
Avoid use of globals where possible

### LINKED ISSUE #64: React native reference error in 0.5.1

sholladay:
Fixes #64 

This is similar to PR #69, however I've taken a different approach. Some of my goals include:
 - No regex testing of URLs.
 - Let `fetch()` handle URL resolution where possible.
 - Guard against `ReferenceError`s by not accessing globals when we don't need them.
 - Simplify testing and mocking by treating `global`, `window`, and `self` as not mutually exclusive.

This should make things significantly easier in non-browser environments. In my testing on Node 10+, I was able to remove everything except for `fetch` and `Response` in [test/_require.js](https://github.com/sindresorhus/ky/blob/master/test/_require.js) and the test suite passed.

Previously, Ky would find the global object and then assume everything lived on that object, which makes things cumbersome in non-DOM / no-browser environments. Now, we lookup each global we need in each namespace. So for example, in Node you don't have to implement all of `window` anymore. If you assign a fetch implementation to `window.fetch` but nothing else, Ky will still be able to find `URL` in modern versions of Node where it lives on `global` and you won't have to assign it manually - it will _just work_.

Further, in all environments, globals are conditionally accessed to avoid unnecessary `ReferenceError`s. Ky also no longer relies on the `Headers` class.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,6 @@ const document = getGlobal('document');
 const URL = getGlobal('URL');
 const URLSearchParams = getGlobal('URLSearchParams');
 const Response = getGlobal('Response');
-const Headers = getGlobal('Headers');
 const fetch = getGlobal('fetch');
 
 const isObject = value => value !== null && typeof value === 'object';
@@ -47,6 +46,13 @@ const deepMerge = (...sources) => {
 	}
 
 	return returnValue;
+};
+
+const objectFromEntries = entries => {
+	return entries.reduce((result, [key, value]) => {
+		result[key] = value;
+		return result;
+	}, {});
 };
 
 const requestMethods = [
@@ -146,15 +152,21 @@ class Ky {
 			this._options.prefixUrl += '/';
 		}
 
-		const url = new URL(this._options.prefixUrl + this._input, document && document.baseURI);
-		if (typeof searchParams === 'string' || searchParams instanceof URLSearchParams) {
-			url.search = searchParams;
-		} else if (searchParams && Object.values(searchParams).every(param => typeof param === 'number' || typeof param === 'string')) {
-			url.search = new URLSearchParams(searchParams).toString();
-		} else if (searchParams) {
-			throw new Error('The `searchParams` option must be either a string, `URLSearchParams` instance or an object with string and number values');
+		// Avoid using URL() if we can, as it does not support relative URLs
+		// and is yet another global to mock in a non-browser environment.
+		if (searchParams) {
+			const url = new URL(this._options.prefixUrl + this._input, document && document.baseURI);
+			if (typeof searchParams === 'string' || (URLSearchParams && searchParams instanceof URLSearchParams)) {
+				url.search = searchParams;
+			} else if (Object.values(searchParams).every(param => typeof param === 'number' || typeof param === 'string')) {
+				url.search = new URLSearchParams(searchParams).toString();
+			} else {
+				throw new Error('The `searchParams` option must be either a string, `URLSearchParams` instance or an object with string and number values');
+			}
+			this._input = url.toString();
+		} else {
+			this._input = this._options.prefixUrl + this._input;
 		}
-		this._input = url.toString();
 
 		this._timeout = timeout;
 		this._hooks = deepMerge({
@@ -163,14 +175,17 @@ class Ky {
 		}, hooks);
 		this._throwHttpErrors = throwHttpErrors;
 
-		const headers = new Headers(this._options.headers || {});
-
 		if (json) {
-			headers.set('content-type', 'application/json');
+			const headers = this._options.headers &&
+				typeof this._options.headers.entries === 'function' ?
+				objectFromEntries(this._options.headers.entries()) :
+				this._options.headers;
+			this._options.headers = {
+				...headers,
+				'content-type': 'application/json'
+			};
 			this._options.body = JSON.stringify(json);
 		}
-
-		this._options.headers = headers;
 
 		this._response = this._fetch();
 
--- a/index.js
+++ b/index.js
@@ -1,19 +1,26 @@
-// Polyfill for `globalThis`
-const _globalThis = (() => {
-	if (typeof self !== 'undefined') {
-		return self;
+const getGlobal = property => {
+	if (typeof self !== 'undefined' && self && property in self) {
+		return self[property];
 	}
 
 	/* istanbul ignore next */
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -1,19 +1,22 @@
...
+		return window[property];
 	}
-})();
+
+	return global[property];
+};
+
+const document = getGlobal('document');
+const URL = getGlobal('URL');
+const URLSearchParams = getGlobal('URLSearchParams');
```
sindresorhus: I'm targeting Node.js 10 now, so we can just use `URL` and `URLSearchParams` globals directly in the source. https://github.com/sindresorhus/ky/commit/c5771ee799a9dd8ad8c705262e028f6da19163ee

index.js
```diff
@@ -138,15 +148,21 @@ class Ky {
...
 
-		const url = new _globalThis.URL(this._options.prefixUrl + this._input, document.baseURI);
-		if (typeof searchParams === 'string' || searchParams instanceof _globalThis.URLSearchParams) {
-			url.search = searchParams;
-		} else if (searchParams && Object.values(searchParams).every(param => typeof param === 'number' || typeof param === 'string')) {
-			url.search = new _globalThis.URLSearchParams(searchParams).toString();
-		} else if (searchParams) {
-			throw new Error('The `searchParams` option must be either a string, `URLSearchParams` instance or an object with string and number values');
+		// Avoid using URL() if we can, as it does not support relative URLs
+		// and is yet another global to mock in a non-browser environment.
```
sindresorhus: This comment is confusing as right below it `URL` *is* actually used.

test/_require.js
```diff
@@ -6,6 +6,3 @@ global.Headers = Headers;
 global.Response = Response;
```
sindresorhus: `global.Headers` is no longer needed, I guess.

index.js
```diff
@@ -155,15 +171,18 @@ class Ky {
...
 		if (json) {
-			headers.set('content-type', 'application/json');
+			const headers = this._options.headers &&
+				typeof this._options.headers.entries === 'function' ?
+				objectFromEntries(this._options.headers.entries()) :
+				this._options.headers;
+			this._options.headers = {
+				...headers,
+				'content-type': 'application/json'
+			};
```
sindresorhus: Not a big fan of this. I don't really see the point in adding lots of complications here to save one global. Users needs to polyfill `Response` for Node.js anyway. So adding `Headers` too doesn't change much. The main target is the browser and we should optimize for that.

test/_require.js
```diff
@@ -6,6 +6,3 @@ global.Headers = Headers;
...
 global.URL = URL;
 global.URLSearchParams = URLSearchParams;
```
sindresorhus: Same as https://github.com/sindresorhus/ky/pull/74#discussion_r245502155

sindresorhus:
I like this approach, except for not using `Headers`.

index.js
sholladay: True, I agree. Without the comment, though, I feel the purpose is not super obvious. Will try to think of a better way to describe or code it. Suggestions welcome!

index.js
sholladay: Yeah, that's fair, I will revert this part. My initial intention was to just do this (without the complicated mess above):

```js
this._options.headers = {
    ...this._options.headers,
    'content-type': 'application/json'
};
```
But then I realized that ky supports custom `Headers` instances and those cannot be spread into an object. It would work with object literals, but not `Headers`.

So, fine, I thought, I'll get its entries and convert it. Seems simple enough. Can't use `Object.entries()` on these `Headers` instances, but `headers.entries()` is a thing. Oh, turns out `headers.entries()` doesn't return a normal entries array, but a `Headers Iterator`. Jeez! Well, at least I can spread that into an array and then use `Object.fromEntries()` to make an object out of it. Nope. `Object.fromEntries()` doesn't have good cross-platform support yet.

This shouldn't be so difficult...  😞

sholladay:
@sindresorhus I think this is ready to go, unless you notice anything else.

Note that I included an unrelated fix for the Travis failures. It was still running Node 8.

sindresorhus:
This is looking great! 👌

@sholladay Would you be interested in joining the project as a maintainer? You have done some really good PRs and you seem interested in the project.

sholladay:
@sindresorhus Yes, would love to join. It's an awesome project. :)

sindresorhus:
Yay. Welcome to the project! 🎉

PR #77, opened 2019-01-06
Fix the `throwHttpErrors` option for POST requests

daixianceng:
Here is a bug:

When I set `throwHttpErrors` to `false` and then `POST` a request, if the response is not OK, it will still throw `HTTPError`. So I fixed it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -183,7 +183,7 @@ class Ky {
 					}
 				}
 
-				if (!response.ok) {
+				if (!response.ok && throwHttpErrors) {
 					throw new HTTPError(response);
 				}
 
--- a/index.js
+++ b/index.js
@@ -167,7 +167,7 @@ class Ky {
 		this._response = this._fetch();
 
 		for (const type of responseTypes) {
-			const isRetry = retryMethods.has(this._options.method.toLowerCase());
+			const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
 			const fn = async () => {
 				if (this._retryCount > 0) {
 					this._response = this._fetch();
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test?

daixianceng:
All tests passed

index.js
```diff
@@ -183,12 +184,14 @@ class Ky {
 					}
 				}
 
-				if (!response.ok) {
+				if (!response.ok && (isRetry || (!isRetry && throwHttpErrors))) {
```
jenbuzz: It works as is but using "this.throwHttpErrors" would be more in-sync with the rest of the code.

```suggestion
				if (!response.ok && (isRetry || (!isRetry && this.throwHttpErrors))) {
```

index.js
```diff
@@ -167,7 +167,8 @@ class Ky {
 		this._response = this._fetch();
 
 		for (const type of responseTypes) {
-			this._response[type] = this._retry(async () => {
+			const isRetry = retryMethods.has(this._options.method.toLowerCase());
```
jenbuzz: isRetry might be a little confusing since it's more like a "allowRetry" check

index.js
sholladay: Saying `isRetry || (!isRetry && throwHttpErrors)` is the same as just saying `isRetry || throwHttpErrors`

index.js
sholladay: Maybe `isRetriableMethod`?

sindresorhus:
Thanks :)

PR #80, opened 2019-01-07
Add generics type to json()

b-gyula:


sholladay:
Would something like this be better, perhaps?

https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-338650717

sindresorhus:
@b-gyula What's the benefit of this change?

b-gyula:
> @b-gyula What's the benefit of this change?
- It's shorter to write `const resp:MyObjectType = await ky.post(URL,params).json<MyObjectType>()` instead of  `const resp:MyObjectType = await ky.post(URL,params).json() as Promise<MyObjectType>` 
- it has possibility to do some generic type check in `json()` as it will know the expected type

sindresorhus:
> instead of `const resp:MyObjectType = await ky.post(URL,params).json() as Promise<MyObjectType>`

The `as` applies to the awaited result, so it should be `as MyObjectType`. You also don't need the `:MyObjectType`, so the correct version would be:

```js
const resp = await ky.post(URL,params).json() as MyObjectType
```

sindresorhus:
I think we should add the JSON type regardless of whether we decide to use generics or not.


```typescript
type JSONObject = {[key: string]: JSONValue};
interface JSONArray extends Array<JSONValue> {}
export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
```

sholladay:
@bthallion since you use TypeScript a lot, any tips or thoughts on this?

bthallion:
I'm not sure it should be the responsibility of .json() to coerce the type, it could confuse the reader into thinking the result is type safe. Adding the JSON return type is a good change

msokk:
How does the casting now work with #82 ?
I'm getting:
```
[ts]
Conversion of type 'JSONValue' to type 'EmployeeResponse' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'JSONArray' is missing the following properties from type 'EmployeeResponse': firstName, lastName, group [2352]
```

`unknown` could directly be castable like the examples referred above. 
Do we now have to cast twice?

bthallion:
You could try validating the type at runtime, and do the JSONValue -> any -> EmployeeResponse assertions that way. This typing does require two assertions, unless you write a deserialize fn to populate an EmployeeResponse class instance (I typically use classes that marshal out data from JSON rather than assert the type directly)

```
function isEmployeeResponse(obj: any): obj is EmployeeResponse {
  try {
    return typeof obj.lastName === 'string'; // Etc.
  } catch (err) {
    console.error('whoops');
  }
}

const resp = await ky.post(URL,params).json();

if (isEmployeeResponse(resp)) {
  const b = resp as EmployeeResponse;
}
```

msokk:
@bthallion thanks, I have totally overlooked user-defined type guards in Typescript.

sindresorhus:
I didn't realize you would have to double-cast it. That's indeed inconvenient.

How about making it:

```ts
json<T extends JSONValue>(): Promise<T>;
```

?

That way we ensure your custom type is a valid JSON value while still making it easy for you to specify it.

---

// @sholladay @7Pass

7Pass:
@sindresorhus you can try it using the following test code:
```
type JSONObject = {[key: string]: JSONValue};
interface JSONArray extends Array<JSONValue> {}
export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;

declare function json<T extends JSONValue>(): Promise<T>;

interface Result  {
    value: number;
}

console.log(json<Result>());
                ^^^^^^^ Index signature is missing in type 'Result'. [2344]
```

sindresorhus:
@7Pass Thanks for elaborating.

=> https://github.com/sindresorhus/ky/pull/98 <=

PR #81, opened 2019-01-16
Add a UMD build to support older environments

### LINKED ISSUE #44: Include a UMD build

sholladay:
Closes #44 

Following the discussions in #44 and #46, we decided to only convert module syntax. This does not transpile the code to ES5, so extremely old environments are still not supported out of the box. Rather, this makes it easy to import ky in non-ESM environments, such as Node.js.

I went with Rollup instead of Babel, because it is simpler to configure and its output is slightly better (easier to read 👀 and no hardcoded AMD ID).

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,13 +13,16 @@
 		"node": ">=10"
 	},
 	"scripts": {
+		"build": "rollup index.js --format umd --name ky --file umd.js",
+		"prepublishOnly": "npm run build",
 		"test": "xo && nyc ava && tsd-check && npm run cypress:run",
 		"cypress:open": "cypress open",
 		"cypress:run": "cypress run"
 	},
 	"files": [
 		"index.js",
-		"index.d.ts"
+		"index.d.ts",
+		"umd.js"
 	],
 	"keywords": [
 		"fetch",
@@ -54,6 +57,7 @@
 		"esm": "^3.0.84",
 		"node-fetch": "^2.3.0",
 		"nyc": "^13.1.0",
+		"rollup": "^1.1.0",
 		"tsd-check": "^0.3.0",
 		"xo": "^0.24.0"
 	},
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
 		"node": ">=10"
 	},
 	"scripts": {
-		"build": "rollup index.js --format umd --name ky --file umd.js",
+		"build": "rollup index.js --format=umd --name=ky --file=umd.js",
 		"prepublishOnly": "npm run build",
 		"test": "xo && nyc ava && tsd-check && npm run cypress:run",
 		"cypress:open": "cypress open",
--- a/readme.md
+++ b/readme.md
@@ -93,13 +93,13 @@ With plain `fetch`, it would be:
 })();
 ```
 
-In environments that do not support `import`, you can load `ky` in UMD format. For example, using `require()`:
+In environments that do not support `import`, you can load `ky` in [UMD format](https://medium.freecodecamp.org/anatomy-of-js-module-systems-and-building-libraries-fadcd8dbd0e). For example, using `require()`:
 
 ```js
 const ky = require('ky/umd').default;
 ```
 
-This also means `ky` can be used as an AMD module, and if no module system is detected, then `ky` will be available as a global.
+If no module system is detected, then `ky` will be available as a global.
 
 ## API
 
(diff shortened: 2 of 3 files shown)


sindresorhus:
What are your thoughts on using the `"module"` field in package.json for the ESM source and `"main"` for the CommonJS version?

Per https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages

package.json
```diff
@@ -13,13 +13,16 @@
 		"node": ">=10"
 	},
 	"scripts": {
+		"build": "rollup index.js --format umd --name ky --file umd.js",
```
sindresorhus:
```suggestion
		"build": "rollup index.js --format=umd --name=ky --file=umd.js",
```

readme.md
```diff
@@ -93,6 +93,13 @@ With plain `fetch`, it would be:
 })();
 ```
 
+In environments that do not support `import`, you can load `ky` in UMD format. For example, using `require()`:
```
sindresorhus: Can you linkify `UMD format` to somewhere that describes what it is?

readme.md
```diff
@@ -93,6 +93,13 @@ With plain `fetch`, it would be:
...
+
+```js
+const ky = require('ky/umd').default;
+```
+
+This also means `ky` can be used as an AMD module, and if no module system is detected, then `ky` will be available as a global.
```
sindresorhus: I don't want to advertise for AMD module compatibility.

readme.md
sholladay: Surprisingly, the UMD repository on GitHub does a pretty poor job of explaining what it is and its purpose. I couldn't quickly find anything succinct. I did find a good [article on module systems](https://medium.freecodecamp.org/anatomy-of-js-module-systems-and-building-libraries-fadcd8dbd0e) that covers UMD. I'll link that, but happy to update it if anyone finds something better.

readme.md
sholladay: Removed.

sholladay:
> What are your thoughts on using the "module" field in package.json ...

I thought about this for a bit and IMO, it is probably unnecessary. Here is the impact as I see it:

 - For bundlers: Seemingly no difference. I've been successfully bundling ky in my own apps with Rollup for months, with no special config. From what I read, Webpack and Parcel should also be able to handle the existing ESM `index.js` file just fine, without using `module`.
 - For browsers: They just download whatever file you specify and don't consult package.json.
 - For CDNs: Probably varies. IIUC, the `unpkg` CDN does pay attention to `main`, but not `module`. So in this case, having `main` / `index.js` be ESM is preferable, since using a CDN implies the environment is probably a browser, and all of our supported browsers support ESM.
 - For Node: This seems like the one area where `main` as CJS and `module` as ESM would benefit. It would allow people to `require('ky')` instead of `require('ky/umd')`.

So the choice appears to be between giving preferential treatment to ESM users on unpkg or CJS users on Node. The former seems like the obvious choice to me.

sholladay:
I added to the "without Webpack" FAQ, but not to the download links. Just thought it would be a bit cluttered otherwise, since the list is of alternative CDNs. I suppose we could reformat it as a table, but I just left it alone for now.

sindresorhus:
Agreed. Looks good now :)

FredKSchott:
In case you ever want to revisit this, it's fairly unexpected for `require("package-name")` to fail on Node.js, unless you specifically don't want to support Node.js. 

@sholladay you mention that the choice is between "giving preferential treatment to ESM users on unpkg or CJS users on Node". You could support everyone with:

```js
// Node.js loads umd
"main": "umd.js",
// Bundlers load esm
"module": "index.js",
// Unpkg loads esm
"unpkg": "index.js",
```

> From the unpkg docs:
> If you omit the file path (i.e. use a “bare” URL), unpkg will serve the file specified by the `"unpkg"` field in package.json, or fall back to `"main"`.

`"module"` is well supported at this point (by both tooling and the community): https://www.pikapkg.com/about/stats

Happy to submit a quick PR if there's interest.

sholladay:
Thanks for the suggestion. I'm not entirely opposed to that, although it would be much better if unpkg just respected `module` like the bundlers do. And I wonder what other tools this might affect. We definitely want to give preference to ESM without tweaking config for various tools.

Do you know if there any plans for that?

PR #82, opened 2019-01-16
Use more specific JSON types in TypeScript

sindresorhus:
Hopefully a JSON type will be included in TypeScript at some point: https://github.com/Microsoft/TypeScript/issues/1897 For now, we just define it manually.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,9 @@
+type JSONObject = {[key: string]: JSONValue};
+interface JSONArray extends Array<JSONValue> {}
+export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
+
+export type JSONStringifyable = object | number | string | boolean;
+
 export type BeforeRequestHook = (options: Options) => void;
 
 export type AfterResponseHook = (response: Response) => Response | void;
@@ -29,7 +35,7 @@ export interface Options extends RequestInit {
 	/**
 	 * Shortcut for sending JSON. Use this instead of the `body` option.
 	 */
-	json?: object;
+	json?: JSONStringifyable;
 
 	/**
 	* Search parameters to include in the request URL.
@@ -77,7 +83,7 @@ export interface ResponsePromise extends Promise<Response> {
 	arrayBuffer(): Promise<ArrayBuffer>;
 	blob(): Promise<Blob>;
 	formData(): Promise<FormData>;
-	json(): Promise<unknown>;
+	json(): Promise<JSONValue>;
 	text(): Promise<string>;
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,7 @@ type JSONObject = {[key: string]: JSONValue};
 interface JSONArray extends Array<JSONValue> {}
 export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
 
-export type JSONStringifyable = object | number | string | boolean;
+export type JSONStringifyable = string | number | boolean | null | object;
 
 export type BeforeRequestHook = (options: Options) => void;
 
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -58,12 +58,12 @@ ky(server.url, {searchParams: {foo: 1}});
 ky(server.url, {searchParams: new URLSearchParams({foo: 'bar'})});
 
 // `json` option
-ky(server.url, {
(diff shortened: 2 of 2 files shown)


index.d.ts
```diff
@@ -1,3 +1,9 @@
+type JSONObject = {[key: string]: JSONValue};
+interface JSONArray extends Array<JSONValue> {}
+export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
+
+export type JSONStringifyable = object | number | string | boolean;
```
sholladay: This seems wrong. `null` can be stringified, for example, and `ky.post(url, { json : null })` should work today.

Shouldn't the `json` option use `JSONValue` instead? And if I'm right, then this can be removed.

index.test-d.ts
```diff
@@ -56,3 +56,15 @@ ky(server.url, {searchParams: 'foo=bar'});
 ky(server.url, {searchParams: {foo: 'bar'}});
 ky(server.url, {searchParams: {foo: 1}});
 ky(server.url, {searchParams: new URLSearchParams({foo: 'bar'})});
+
+// `json` option
+ky(server.url, {
+	json: {
```
sholladay: nit: `GET` requests are not supposed to have a body, so while these tests serve their purpose, they are actually demonstrating improper usage.

szmarczak: We could automatically set `method` to `POST` if user provides a body with no method.

sholladay: Hmm my initial reaction is that feels too magical to me. I actually didn't realize `got` did that. Who says they want `POST` instead of `PUT` or `PATCH`? I suppose we should align their behavior, though.

szmarczak: > Who says they want POST instead of PUT or PATCH? 

If someone wants PUT or PATCH, then sets `method: 'PUT'` or `method: 'PATCH'` ;)

sholladay: Right, it's easy to be explicit about that, and personally, I would force the user to choose a method.

index.d.ts
sindresorhus: Ah, missed the `null`.

The reason I didn't go with `JSONValue` is that it might be inconvenient. What if you have an object with a property with a value of `new Date()` and you just want it to be stringified? `JSON.stringify()` usage is expected to coerce values and I'm not sure we should be super strict here. Thoughts?

index.test-d.ts
sindresorhus: Continued in https://github.com/sindresorhus/got/issues/639#issuecomment-455042713

sindresorhus: > GET requests are not supposed to have a body, so while these tests serve their purpose, they are actually demonstrating improper usage.

@sholladay Any ideas how we could enforce this in TS?

index.d.ts
sholladay: So the idea is essentially to be explicit and "strict" about _output_ of `.json()`, but flexible about input of the `json` option. That makes sense to me. 👍

index.test-d.ts
sholladay: My first thought is to use an overload, but there are some other approaches [detailed here](http://artsy.github.io/blog/2018/11/21/conditional-types-in-typescript/) that I haven't used and may be worth exploring.

sindresorhus: Continued in https://github.com/sindresorhus/ky/issues/85

index.d.ts
```diff
@@ -1,3 +1,9 @@
+type JSONObject = {[key: string]: JSONValue};
```
felixfbecker: I don't think this does what you intend it to. The way this is defined, no object is considered "JSON stringifyable" unless it has an index signature. For example, this would fail:

```ts
interface Foo { foo: number }
const foo: Foo = { foo: 123 }
const json: JSONObject = foo // Error: Foo is not assignable to JSONObject because it has no index signature
```

I don't think there is currently a way to express this constraint correctly.

Besides, JSON.stringify() actually doesn't care if an object has methods. It will simply omit them (unlike `postMessage()`, which throws an error).

sindresorhus: This is for the response though, not for being stringifyiable, for stringification we're quite lax: https://github.com/sindresorhus/ky/blob/b6ed9dfc17e1081b992b54f95e65969deb633c5a/index.d.ts#L5

PR #84, opened 2019-01-20
Replace Cypress with AVA + Puppeteer

ianwalter:
Cypress is great but when it comes to unit testing the developer experience is better with AVA. With this setup, I think I would be able to help complete #27.
- [x] Replace puippeteer-helper with in-repo helper
- [x] Move tests to browser-specific file


package.json
```diff
@@ -47,12 +47,13 @@
 		"node-fetch"
 	],
 	"devDependencies": {
+		"@ianwalter/dist": "^1.4.0",
+		"@ianwalter/puppeteer-helper": "^1.0.0",
```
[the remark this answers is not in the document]
ianwalter: Yes, I added `dist` before I pulled in the changes that added Rollup and the UMD build. I will just use that instead.

`puppeteer-helper` is based off of a recipe in the AVA docs so I think if they wanted that to be an official helper they would have done it already. I don't think extensibility is an issue since you can pass any `puppeteer` config through `puppeteer-helper` as the second argument. The only thing that I could see becoming an issue is that `puppeteer-helper` controls the `puppeteer` version but I can't think of any good alternatives to that.

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
+dist
```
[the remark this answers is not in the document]
ianwalter: I'm pretty sure you can't load ESM since whatever AVA uses will try to parse the import statement and throw an error. I will use the UMD build.

test/prefix-url.js
```diff
@@ -33,3 +41,25 @@ test('prefixUrl option', async t => {
 
 	await server.close();
 });
+
+test('prefixUrl option in browser', withPage, async (t, page) => {
```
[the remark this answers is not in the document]
ianwalter: I was just trying to keep the tests that were already in Cypress and provide a starting point for testing in the browser. I don't see the point in testing prefixUrl in node at all since how it works is affected by `document`, but I'm also not willing to refactor all of the tests right now.

package.json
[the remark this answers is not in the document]
sholladay: In my opinion, `puppeteer-helper` should import `puppeteer` as a `peerDependency`, so its users can use their own Puppeteer version.

test/prefix-url.js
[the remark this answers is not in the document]
sholladay: > I was just trying to keep the tests that were already in Cypress and provide a starting point for testing in the browser.

Right, I'm just asking to keep the browser tests in a separate file, as the Cypress tests were.

> I don't see the point in testing prefixUrl in node at all since how it works is affected by document

It's very important to test `prefixUrl` in Node, precisely because its behavior is affected by the presence of `document`. We support `prefixUrl` in Node even you do not define a `document`.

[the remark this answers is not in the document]
sholladay: I suppose I didn't word that very well. It is true that Node is not a target environment. But people are using Ky in Node anyway. They have complained when we've broken Node support in the past and thus far we have fixed it since it was easy. More importantly, we support Web Workers, which is a browser-based environment without a `dicument`. We also just added support for [Deno](https://github.com/denoland/deno}.

So, what I should have said is, it's very important to test Ky in a non-DOM environment. Doing that in Node happens to be a convenient way to do so that also aligns with things our users are doing. I don't see a good reason to remove the Node tests, but I guess it would be fine if we add some Web Worker tests, etc.

package.json
```diff
@@ -13,11 +13,9 @@
 		"node": ">=10"
 	},
 	"scripts": {
-		"build": "rollup index.js --format=umd --name=ky --file=umd.js",
-		"prepublishOnly": "npm run build",
-		"test": "xo && nyc ava && tsd-check && npm run cypress:run",
-		"cypress:open": "cypress open",
-		"cypress:run": "cypress run"
+		"pretest": "rollup index.js --format=umd --name=ky --file=umd.js",
```
sholladay: We can avoid this by using `page.addScriptTag({path: './index.js', type: 'module'})`. Is there a reason you didn't want to do that? Not a big deal, but loading the source directly feels a little more clean to me and it will speed up the tests a bit.

szmarczak: Why did you rename this?

package.json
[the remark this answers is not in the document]
szmarczak: IMO there's no need for that module. [10 additional lines of code](https://github.com/avajs/ava/blob/master/docs/recipes/puppeteer.md) is not much.

szmarczak:
@ianwalter Any update on this?

ianwalter:
@sholladay @szmarczak Thank you for your feedback. I should be able to address this sometime this week.

package.json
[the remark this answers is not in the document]
ianwalter: Fair enough, I find it useful, but I'll replace it with a in-repo helper.

test/prefix-url.js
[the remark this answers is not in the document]
ianwalter: Ok, gotcha. I will move these to a `browser.test.js` or something like that.

package.json
ianwalter: @sholladay Yea it would be nice if I could do that but I don't think it's practical because if you use a module script you have to import the module in the `evaluate` code right? But then if you have an import statement outside of the top level you get a parse error. I can think of a couple ways around this but they're not great.

@szmarczak Rename `build` to `pretest`? Out of habit initially, but really because the browser tests test against the UMD build so it's convenient to use the `pretest` npm convention since it automatically gets run before `test` runs without needing to explicitly run it. You can still run it manually (as is the case with prepublishOnly) if you want though.

ianwalter:
Ok, I believe I have addressed the feedback. I also added a browser test for aborting a request.

package.json
sholladay: > if you use a module script you have to import the module in the `evaluate` code right?

Yeah, I guess you're right. Just didn't think it all the way through. We could make a helper for that, but maybe it's not worth it at that point.

szmarczak:
@sindresorhus @sholladay LGTM.

test/browser.js
```diff
@@ -0,0 +1,63 @@
+import test from 'ava';
+import createTestServer from 'create-test-server';
+import withPage from './helpers/with-page';
+
+test('prefixUrl option', withPage, async (t, page) => {
+	const server = await createTestServer();
+	server.get('/', (request, response) => {
+		if (request.query.page === '/https://cat.com/') {
```
sholladay: Looks like this isn't used.

package.json
sindresorhus: > but really because the browser tests test against the UMD build so it's convenient to use the `pretest` npm convention since it automatically gets run before `test` runs without needing to explicitly run it. You can still run it manually (as is the case with prepublishOnly) if you want though.

I prefer being explicit. It's weird having to run `npm run pretest` to build. The script should be called `build`.

sindresorhus: > I can think of a couple ways around this but they're not great.

Just curious, what are the ways around it? Using `import` would be much nicer, so would be good to know the downsides.

sindresorhus:
Added a couple of comments in https://github.com/sindresorhus/ky/pull/84#discussion_r256294662

package.json
sholladay: I will take care of putting the `build` script back. I'm working on some test tweaks right now.

test/browser.js
```diff
@@ -0,0 +1,66 @@
...
+	await page.addScriptTag({path: './umd.js'});
+
+	const error = await page.evaluate(url => {
+		return new Promise(resolve => {
+			window.ky = window.ky.default;
+			const controller = new AbortController();
+			const req = window.ky(`${url}/test`, {signal: controller.signal}).text();
+			controller.abort();
+			req.then(resolve).catch(error => resolve(error.toString()));
+		});
```
sindresorhus: I don't think you need to use the `Promise` constructor here. You can just handle it with async/await and try/catch.

sholladay: I removed `new Promise()` but didn't bother to use `await` and `try` / `catch`. Just seemed more verbose than necessary here. Happy to change it, though, LMK.

sindresorhus: Looks good

package.json
ianwalter: @sindresorhus 
1. Pass a string to `page.evaluate` instead of a function. No parse error but writing JavaScript in a string would be annoying. I could see it being a viable option in the future though if there was a tagged template literal for JavaScript that editors could syntax highlight.
2. Extract the evaluation code to another file. This creates more files, essentially takes some of the test code out of the test file, and changes the evaluation interface a bit. I actually just switched over to this method personally despite its downsides. I updated [`puppeteer-helper`](https://github.com/ianwalter/puppeteer-helper) to make this easy, you just call `t.evaluate()` with a relative path to the evaluation script.

sindresorhus: Okey. Makes sense. Thanks for elaborating.

sindresorhus:
Thank you, @ianwalter :)

PR #86, opened 2019-01-20
Support running Ky in Deno

sholladay:
I wanted to run Ky in [Deno](https://github.com/denoland/deno). Turns out it works, with the only blocker being that Deno uses `globalThis` instead of `global`. Annoying, but Deno is just [following standards](https://github.com/tc39/proposal-global).

Should we officially advertise support for it? I don't know. But I am so excited about Deno. And since the code is super simple, I figured I'd at least propose it. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,11 +11,12 @@ const getGlobal = property => {
 		return window[property];
 	}
 
-	if (typeof global !== 'undefined' && global && property in global) {
-		return global[property];
+	/* istanbul ignore next */
+	if (typeof globalThis !== 'undefined' && globalThis && property in globalThis) {
+		return globalThis[property];
 	}
 
-	return globalThis[property];
+	return global[property];
 };
 
 const document = getGlobal('document');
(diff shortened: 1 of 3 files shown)


sindresorhus:
Sure, we can mention it in the readme. I like Deno too.

zhmushan:
Can `ky` provide a ts branch for deno specifically?
I prefer to get rid of the `.d.ts` file

sholladay:
> Can ky provide a ts branch for deno specifically?

A branch wouldn't be too hard to maintain at our current size, but I'm not sure long term, and it's a bit unusual. Worth considering, though. At least Deno can import modules directly from GitHub, so it's equally convenient from the user's perspective.

> I prefer to get rid of the .d.ts file

I can see the benefit, but unless we did decide to maintain a TS branch in git, it would also require an additional build step for JS environments.

Either way, seems like one step forward, one step back, to me.

I'm going to merge this, but we can continue the conversation here.

sindresorhus:
@sholladay It's out https://github.com/sindresorhus/ky/releases/tag/v0.7.1

PR #87, opened 2019-01-20
Improve type system for get and head methods

### LINKED ISSUE #85: TypeScript - Don't allow `body` option for GET and HEAD requests
Would be nice to not allow the `body` option when using `ky.get()` and `ky.head()`.

We could use a custom Omit type for this:

```typescript
type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
```

Would also be nice to handle `ky(…, {method: 'get', body: …})`, but not sure whether that is possible?

It would also be nice to disallow the `body` option when the `json` option is used.

More info: http://artsy.github.io/blog/2018/11/21/conditional-types-in-typescript/

doniyor2109:
Fixes #85

Todo:

 - [x] Do not allow the body option when using `ky.get()` and `ky.head()`
 - [x] Do not allow the body option when using `(…, {method: 'get', body: …})`
 - [x] Disallow the `body` option when the `json` option is used

### DIFF
--- a/index.js
+++ b/index.js
@@ -169,7 +169,7 @@ class Ky {
 
 		if (json) {
 			if (this._options.body) {
-				throw new TypeError('The `json` option cannot be used with the `body` option');
+				throw new Error('The `json` option cannot be used with the `body` option');
 			}
 
 			headers.set('content-type', 'application/json');
--- a/index.js
+++ b/index.js
@@ -168,6 +168,10 @@ class Ky {
 		const headers = new Headers(this._options.headers || {});
 
 		if (json) {
+			if (this._options.body) {
+				throw new TypeError('The `json` option cannot be used with the `body` option');
+			}
+
 			headers.set('content-type', 'application/json');
 			this._options.body = JSON.stringify(json);
 		}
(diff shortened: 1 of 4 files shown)


sholladay:
LGTM.

Feels a little verbose but I don't have any constructive suggestions.

sindresorhus:
@doniyor2109 Can you add some type tests for the various cases? => https://github.com/sindresorhus/ky/blob/master/index.test-d.ts

index.js
```diff
@@ -168,6 +168,10 @@ class Ky {
 		const headers = new Headers(this._options.headers || {});
 
 		if (json) {
+			if (this._options.body) {
+				throw new TypeError('The `json` option cannot be used with the `body` option');
```
sindresorhus: It's not really a TypeError. Use `Error`.

sindresorhus:
@szmarczak This is still missing the type tests I requested in https://github.com/sindresorhus/ky/pull/87#issuecomment-460209196.

szmarczak:
Ok, I'll try to do it :)

szmarczak:
@sindresorhus done

PR #89, opened 2019-01-25
Add `ky/umd` TypeScript definition

rifler:


umd.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'ky/umd' {
+	import ky from 'ky';
+
+	export default ky;
```
sindresorhus: You can just export it directly:

```js
export {default} from 'ky';
```

umd.d.ts
```diff
@@ -0,0 +1,6 @@
...
+	export * from 'ky';
```
sindresorhus: I'm not that familiar with export syntax, but are you sure this doesn't also export the default?

umd.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'ky/umd' {
```
sindresorhus: I don't think this is needed.

rifler: you are right

umd.d.ts
rifler: Only `export * from 'ky';` doesn't work for my `typescript@3.1.6`

PR #92, opened 2019-01-31
Add generic typing for ResponsePromise.json()

7Pass:
Instead of enforcing `JSONValue`, which very often forces additional casting, let the user choose which value type he wants the result to be.
JSONValue is specified as the default type, so there's no breaking change for code that already expects `json()` to return JSONValue.


szmarczak:
Duplicate of #80

PR #93, opened 2019-01-31
Add tip about afterResponse hooks not responding

7Pass:
It took me quite a while to realize why my `afterResponse` hooks are not working.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -358,6 +358,10 @@ Alternatively, you can use the [`umd.js`](umd.js) file with a traditional `<scri
 </script>
 ```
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
That sounds like a bug. Can you open an issue instead? The `afterResponse` hook should work even if you do `await ky(…)`.

sholladay:
Confirmed. From what I can see, the underlying problem also applies to our retry logic.

[This code](https://github.com/sindresorhus/ky/blob/d5d3eaa24f903294d1b86d96a7652262c61536eb/index.js#L187-L211) probably needs to be moved to a function and called both inside and outside of the loop.

PR #95, opened 2019-02-01
Fix `afterResponse` hook & retry not being called

### LINKED ISSUE #94: afterResponse hooks not fired if await ky directly
As discussed in Pr #93:
`afterResponse` hooks are not executed for the following code:

```js
const response = await ky.get('https://example.com', {
    hooks: {
        afterResponse: [
             response => {
                 // This won't get called
             }
         ]
     }
});

const data = await response.json();
```

I know this is not how `ky` is supposed to be used, but it's easy to fall into this and see that everything is working, except the hooks.

### LINKED ISSUE #96: Error not thrown without the json method
Hi,

I was just trying to catch errors in the front-end of my react/redux app when I spotted a really odd behavior.

- When doing a DELETE request on my API without ".json()" appended to my response, http errors are not thrown, even with throwHttpErrors forced true in the options.

```js
try {
    await api.delete(`roles/${id}`);
    dispatch(actions.deleteRole(id));
  } catch (error) {
    dispatch({
      type: types.DELETE_ROLE_ERROR,
      payload: {
        error,
      },
    });
  }
```

- When doing the same DELETE request with ".json()" appended to the response, errors are thrown but the request is fired twice so the second one always returns a 404.

```js
try {
    await api.delete(`roles/${id}`).json();
    dispatch(actions.deleteRole(id));
  } catch (error) {
    dispatch({
      type: types.DELETE_ROLE_ERROR,
      payload: {
        error,
      },
    });
  }
```

Any idea why this would happen?

7Pass:
Fixes #94
Fixes #96

~~`ky` will no longer return a promise directly. It now returns an object with Body methods, plus a `raw` method to retrieve a cloned `Response` directly, useful when there's a need to simply check status code.~~

Note from @szmarczak: I've edited this PR so it should work properly now.

#### TODO:

 - [x] Tests

### DIFF
--- a/index.js
+++ b/index.js
@@ -183,49 +183,35 @@ class Ky {
 
 		this._response = this._fetch();
 
-		const getResponse = async (type) => {
-			if (this._retryCount > 0) {
-				this._response = this._fetch();
-			}
-
-			let response = await this._response;
-
-			for (const hook of this._hooks.afterResponse) {
-				// eslint-disable-next-line no-await-in-loop
-				const modifiedResponse = await hook(response.clone());
-
-				if (modifiedResponse instanceof Response) {
-					response = modifiedResponse;
-				}
-			}
-
-			if (!response.ok && (isRetriableMethod || this._throwHttpErrors)) {
-				throw new HTTPError(response);
-			}
-
-			if (type) {
-				return response.clone()[type]();
-			}
-
-			// Return the raw response
-			return response.clone();
-		};
-
-		const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
-		const getBodyMethod = async (type) => {
-			const fn = async () => await getResponse(type);
-			return isRetriableMethod ? this._retry(fn) : fn;
-		};
-
-		const promise = {
-			raw: getBodyMethod(""),
-		};
-
 		for (const type of responseTypes) {
-			promise[type] = getBodyMethod(type);
+			const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
+			const fn = async () => {
+				if (this._retryCount > 0) {
+					this._response = this._fetch();
+				}
+
+				let response = await this._response;
+
+				for (const hook of this._hooks.afterResponse) {
+					// eslint-disable-next-line no-await-in-loop
+					const modifiedResponse = await hook(response.clone());
+
+					if (modifiedResponse instanceof Response) {
+						response = modifiedResponse;
+					}
+				}
+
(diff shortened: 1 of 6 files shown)


7Pass:
This fixes bug #94

7Pass:
Then we can return the `raw()` call and attach the body methods to that promise instead. I'm not sure the memory affect because the Response will be cloned twice: once for the returned object, then another time for the body object.

szmarczak:
I agree with @sholladay. The idea of `getResponse` function seems to be nice. I'll try to fix this PR.

szmarczak:
There's another bug: it will make another request when using method shortcuts...

index.js
```diff
@@ -181,34 +181,32 @@ class Ky {
...
+			}
 
-					if (modifiedResponse instanceof Response) {
-						response = modifiedResponse;
-					}
-				}
+			return response;
+		};
 
-				if (!response.ok && (isRetriableMethod || this._throwHttpErrors)) {
```
szmarczak: I've removed `isRetriableMethod` check here, I don't think it's needed here...

szmarczak:
TODO: tests

index.js
```diff
@@ -181,34 +181,32 @@ class Ky {
...
-					throw new HTTPError(response);
-				}
+		const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
+		this._response = isRetriableMethod ? this._retry(fn) : fn();
```
sholladay: Unrelated to this PR, but it's been bothering me that `this._response` is saved on the instance despite only being used in the scope of the constructor. I'm pretty sure we can just do `const result =` instead of `this._response =`.

sholladay:
Awesome! This is exactly what I was hoping for. 👏 

I don't see any obvious problems. I will play around with this in my app in the next day or two.

index.js
szmarczak: Sure, I saw this too but maybe it was done like so for some reason.

@sindresorhus What do you think?

sindresorhus: 👍👍

sindresorhus:
> In my opinion, we should remain as close to the fetch API as we reasonably can, so that Ky is intuitive and familiar. I'm 👎 on this particular design. I'm pretty sure we can fix the bug without changing our return type.

👍

index.js
szmarczak: `const result` or `this._response`?

sindresorhus: `const result`

szmarczak:
I'll write the tests tomorrow

PR #97, opened 2019-02-03
Use `.toString()` instead of `String(...)`

szmarczak:
http://jsben.ch/ykoFf

### DIFF
--- a/index.js
+++ b/index.js
@@ -139,8 +139,8 @@ class Ky {
 			...otherOptions
 		};
 		this._options.method = normalizeRequestMethod(this._options.method);
-		this._options.prefixUrl = String(this._options.prefixUrl || '');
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is based on a meaningless microbenchmark. It's not testing actual usage. It will have no effect on actual usage. Also, `String()` is faster in Safari, but again, meaningless.

sindresorhus:
screen shot 2019-02-03 at 21 39 51

sindresorhus:
I'm always vary (from experience) every time I see a link to a microbenchmark :P

Microbenchmarks don't even reflect how JS engines work and optimize. Engines works on context, so some code in isolation can be a lot slower than when it's in context of some other code.

Therefor, benchmarks should always test the actual API, not a small code-unit.

PR #101, opened 2019-02-13
Closes #96

### LINKED ISSUE #96: Error not thrown without the json method

NoChapter:
The retrial should be triggered at the begingning, not by calling explict BODY method following.

### DIFF
--- a/index.js
+++ b/index.js
@@ -200,37 +200,30 @@ class Ky {
 
 		this._options.headers = headers;
 
-		this._response = this._fetch();
+		const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
 
-		for (const type of responseTypes) {
-			const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
-			const fn = async () => {
-				if (this._retryCount > 0) {
+		// distinguish returned value from this._response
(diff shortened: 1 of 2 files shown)


sholladay:
Thanks, but there is a PR for this already: https://github.com/sindresorhus/ky/pull/95

I think we'll go with the other one instead, since it's further along in terms of reviews, etc.

PR #110, opened 2019-03-16
Make the `retry` option more powerful

### LINKED ISSUE #2: More comprehensive `retry` option
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/ky/issues/2)

In addition to accepting a number, it should accept an object with the ability to specify which methods and status codes to retry on. Same as Got: https://github.com/sindresorhus/got#retry

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/ky/issues/2).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/ky/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

whitecrownclown:
Fixes #2

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,21 +47,21 @@ export interface RetryOptions {
 	/**
  	The set of methods allowed to retry
 
- 	@default new Set(['get', 'put', 'head', 'delete', 'options', 'trace'])
+ 	@default ['get', 'put', 'head', 'delete', 'options', 'trace']
  	*/
 	methods?: string[];
 
 	/**
  	The set of statusCodes allowed to retry
 
- 	@default new Set([408, 413, 429, 500, 502, 503, 504])
+ 	@default [408, 413, 429, 500, 502, 503, 504]
  	*/
 	statusCodes?: number[];
 
 	/**
  	The set of statusCodes allowed to retry with Retry-After header
 
- 	@default new Set([413, 429, 503])
+ 	@default [413, 429, 503]
  	*/
 	afterStatusCodes?: number[];
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,6 +30,39 @@ export interface Hooks {
 	afterResponse?: AfterResponseHook[];
 }
 
+export interface RetryOptions {
+	/**
+	 * Number of retries to retry failed requests
+	 *
+	 * @default 2
+	 */
+	retries?: number,
+	/**
+	 * The set of methods allowed to retry
+	 *
+	 * @default new Set(['get', 'put', 'head', 'delete', 'options', 'trace'])
+	 */
+	methods?: Set<string>,
+	/**
+	 * The set of statusCodes allowed to retry
+	 *
+	 * @default new Set([408, 413, 429, 500, 502, 503, 504])
+	 */
+	statusCodes?: Set<number>,
+	/**
+	 * The set of statusCodes allowed to retry with Retry-After header
+	 *
+	 * @default new Set([413, 429, 503])
+	 */
+	afterStatusCodes?: Set<number>,
+	/**
+	 * If Retry-After header is greater than `maxRetryAfter`, the request will be canceled.
+	 *
+	 * @default undefined
+	 */
+	maxRetryAfter?: undefined | number
+}
+
 /**
  * Options are the same as fetch, with some exceptions.
  */
@@ -52,11 +85,9 @@ export interface Options extends RequestInit {
 	prefixUrl?: URL | string;
 
 	/**
-	 * Numer of times to retry failed requests.
-	 *
-	 * @default 2
+	 * RetryOptions or Number of times to retry failed requests
 	 */
-	retry?: number;
+	retry?: RetryOptions | number;
 
 	/**
 	 * Timeout in milliseconds for getting a response.
--- a/index.d.ts
+++ b/index.d.ts
@@ -38,35 +38,42 @@ export interface Hooks {
 
 export interface RetryOptions {
 	/**
-	 * Number of retries to retry failed requests
-	 *
-	 * @default 2
-	 */
-	retries?: number,
+ 	Number of retries to retry failed requests
+
+ 	@default 2
+
+ 	@example
+ 	```
+ 	import ky from 'ky';
+
+ 	const parsed = await ky('https://example.com', {retry: {retries: 10, methods: ['get'], statusCodes: [413]}}).json();
+ 	```
+ 	*/
+	retries?: number;
 	/**
-	 * The set of methods allowed to retry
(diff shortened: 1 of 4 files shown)


szmarczak:
Also this needs tests.

whitecrownclown:
Thanks for the feedback, i will add tests also.

index.js
```diff
@@ -135,14 +135,21 @@ class Ky {
...
 
 		this._options = {
 			method: 'get',
 			credentials: 'same-origin', // TODO: This can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
-			retry: 2,
+			retry: {
+				retries: 2,
+				methods: retryMethods,
+				statusCodes: retryStatusCodes,
+				afterStatusCodes: retryAfterStatusCodes,
```
sholladay: This doesn't seem to be documented. And I find the option names confusing because the difference between `retry.statusCodes` and `retry.afterStatusCodes` is not clear.

That said, I don't have any good suggestions, either.

szmarczak: @sindresorhus Should this be exposed?

sindresorhus:
You also need to update the TypeScript definition.

index.js
sindresorhus: > Should this be exposed?

No

sindresorhus:
@whitecrownclown Bump :)

whitecrownclown:
Update: I will continue working on this this following weekend

index.js
whitecrownclown: @sindresorhus Do you mean you shouldn't be able to overwrite `afterStatusCodes` and just use the defaults?

sindresorhus: Yes. We have not exposed it in Got either and no one has requested it yet.

index.d.ts
```diff
@@ -36,6 +36,46 @@ export interface Hooks {
...
+ 	const parsed = await ky('https://example.com', {retry: {retries: 10, methods: ['get'], statusCodes: [413]}}).json();
+ 	```
+ 	*/
+	retries?: number;
+	/**
+ 	The set of methods allowed to retry
+
+ 	@default new Set(['get', 'put', 'head', 'delete', 'options', 'trace'])
+ 	*/
+	methods?: Set<string> | Array<string>;
```
sindresorhus: Only array as input.

```suggestion
	methods?: string[];
```

index.d.ts
```diff
@@ -36,6 +36,46 @@ export interface Hooks {
...
+	/**
+ 	Number of retries to retry failed requests
+
+ 	@default 2
+
+ 	@example
+ 	```
+ 	import ky from 'ky';
+
+ 	const parsed = await ky('https://example.com', {retry: {retries: 10, methods: ['get'], statusCodes: [413]}}).json();
```
sindresorhus: Should be wrapped in an async IIFE `(async () => { })();`

It's also not very readable on just one line.

Also, this code example should be added to the option docs in the readme too. The readme and TS definition should be in sync.

index.d.ts
```diff
@@ -36,6 +36,46 @@ export interface Hooks {
 	afterResponse?: AfterResponseHook[];
 }
 
+export interface RetryOptions {
+	/**
+ 	Number of retries to retry failed requests
```
sindresorhus:
```suggestion
 	The number of times to retry failed requests.
```

index.d.ts
sindresorhus: This code example should be on line 102

sindresorhus:
I think the TS and readme docs could use some more work. Also ensure they're as in sync as possible.

index.d.ts
szmarczak: A pleonasm :joy:

sindresorhus:
https://github.com/sindresorhus/ky/pull/110#discussion_r297628398 is not reflected in the implementation and tests. We should only accept array as input, but use Set *internally*.

readme.md
```diff
@@ -172,15 +172,32 @@ import ky from 'ky';
...
 ##### retry
 
-Type: `number`<br>
-Default: `2`
+Type: `object | number`<br>
 
-Retry failed requests made with one of the below methods that result in a network error or one of the below status codes.
+Default:
+- retries: `2`
+- methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
```
sindresorhus: The tests and examples are lowercase, so these should probably be that too.

whitecrownclown:
@sindresorhus 	Should we throw an error when using `Set` as input? Right now it's documented as only using `Array` and the options are normalized into `Set` [internally](https://github.com/sindresorhus/ky/pull/110/files#diff-168726dbe96b3ce427e7fedce31bb0bcR159).

sindresorhus:
Yes, test it with `Array.isArray`.

whitecrownclown:
I'm not sure as to why the build fails on travis. The tests run alright locally.

`browser › prefixUrl option
  /home/travis/build/sindresorhus/ky/node_modules/puppeteer/lib/Launcher.js:351
  Rejected promise returned by test.`

`Reason:
  TimeoutError {
    message: 'Timed out after 30000 ms while trying to connect to Chrome! The only Chrome revision guaranteed to work is r672088',
  }`

readme.md
```diff
@@ -172,15 +172,32 @@ import ky from 'ky';
...
+
+Delays between retries counts with function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is attempt number (starts from 1).
+
+```js
+import ky from 'ky';
+
+(async () => {
+	const retryOptions = {retries: 10, methods: ['get'], statusCodes: [413]};
+
+  	const parsed = await ky('https://example.com', {retry: retryOptions}).json();
```
sindresorhus: There are lots of mixed whitespace all over this PR. In the future, turn on showing whitespace so you can clearly see this yourself.

sindresorhus:
I don't know why Travis is failing, but it's passing locally for me too.

index.js
```diff
@@ -139,21 +139,53 @@ const timeout = (promise, ms, abortController) => new Promise((resolve, reject)
...
+const normalizeRetryOptions = retry => {
+	if (typeof retry === 'number') {
+		return {
+			...defaultRetryOptions,
+			retries: retry
+		};
+	}
+
+	if (retry.methods && !Array.isArray(retry.methods)) {
+		throw new Error('retry.methods must be an array');
```
sindresorhus: I noticed now that this should not throw, since Ky returns a promise, it should **always** return a promise and never synchronously throw.

PR #111, opened 2019-03-25
Clarify usage of `.json()`

mesqueeb:
It took me a long time until I understood what that sentence actually meant, until I saw the `json()` used in other examples. It's VERY easy to overlook, so making this small modification goes a long way for newbies to understand this sentence much better!!


sholladay:
Thanks!

I made a small change so it flows better and doesn't accidentally imply that `.get()` is called _on_ the `Response`.

PR #113, opened 2019-03-28
Fix Error types

### LINKED ISSUE #112: HTTPError type incorrect
When trying to create a `new HTTPError` using a response, Typescript says that `HTTPError` only accepts a message as a `string | undefined`. The source shows that it should accept a `Response` object.

stramel:
Fixes #112 

Adds `constructor` definitions for both `HTTPError` and `TimeoutError`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -129,7 +129,9 @@ export class HTTPError extends Error {
 /**
  * The error thrown when the request times out.
  */
-export class TimeoutError extends Error {}
+export class TimeoutError extends Error {
+	constructor();
+}
 
 export interface Ky {
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -122,6 +122,7 @@ export interface ResponsePromise extends Promise<Response> {
  * The error has a response property with the `Response` object.
  */
 export class HTTPError extends Error {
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -122,13 +122,16 @@ export interface ResponsePromise extends Promise<Response> {
...
+	constructor(response: Response);
 	response: Response;
 }
 
 /**
  * The error thrown when the request times out.
  */
-export class TimeoutError extends Error {}
+export class TimeoutError extends Error {
+	constructor();
```
sholladay: Can TypeScript not figure this out on its own? Seems silly.

stramel: It was using the Error constructor which accepts a message.

sholladay: Yeah, I get the intention of the type definition. I'm just a bit surprised that TypeScript can't figure out that this function doesn't accept any arguments.

stramel: I believe that Typescript would be smart enough to infer the proper constructor if our implementation was written in Typescript.

sholladay: Yeah, perhaps. Seems like a good feature request, in any case.

Issue #100, opened 2019-02-13
request object in beforeRequest hook doesnt contain searchParams


dontwork:
We are using ky.extend ad have some hooks set up in our instance of ky.

Reading the docs for beforeRequest and it says `Ky will make no further changes to the request after this. ` based on that i expected to be able to access the searchParams that are present for that request but the object looks like this:

```
credentials: "same-origin"
headers: Headers {}
method: "GET"
prefixUrl: "URL"
retry: 2
```

is it possible to access searchParams in this hook?

sholladay:
Hi @dontwork. I looked into this and the problem is essentially that the `options` that Ky passes to your `beforeRequest` hook are _not_ all of the options that the end user passed to Ky, but rather all of the options that Ky will pass to [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch). You seem to expect the former and I can see how this is confusing because `prefixUrl` and `retry` are in there, but that seems like a mistake on our part, since `fetch()` won't know what to do with them and it could lead to problems in the future if new options are added to `fetch()`.

In a similar vein, `fetch()` won't understand `searchParams` and we are not currently passing it, so you don't see it in your `beforeRequest` hook.

We should make this clearer by being more deliberate about it. We want to pass to `fetch()` any options that Ky has normalized plus all unknown options, but probably not any options that are only for Ky itself. That's the set of options that `beforeRequest` should receive before we pass it to `fetch()`.

Once that is done, there's still the issue of accessing the `searchParams` somehow. One idea would be to add a new argument to `beforeRequest` hooks for Ky-specific options. Meh. Or maybe we should pass everything to the hook in one object and then use the hook's return value as the fetch options. There are some different approaches to consider...

Another thing to note is that by the time your hook runs, we've already processed the `searchParams` and modified the request URL accordingly. Your hook runs too late to modify it in any meaningful way even if we passed it to you.

Our documentation could be clearer on all of this, too. It mentions "normalized options" without defining which options and whether they can be modified in the hook. Some can, such as `headers`, while others wouldn't be even if we passed them, such as `searchParams`.

dontwork:
So, what you are saying is that if i wanted to perform some action that added or removed searchParams at a global level this cannot and will not be possible within this hook. The solution would likely be to have a wrapping function around ky have it in there?

sholladay:
[...]

PR #115, opened 2019-03-28
Support accessing to options in response hook

### LINKED ISSUE #114: Support accessing options in afterResponse hook
There is a case where I need to behave differently in afterResponse hook according to the parameters passed in options, so access options in afterResponse hoook is very useful.

It is an example, I extend ky, when error occurs, global messages can be display or not according to `silent` passed in `_ky.get()`.
```js
const _ky = ky.extend({
  hooks: {
    afterResponse: [
      (res, { silent }) => {
        if (!res.ok && !silent) {
          // display global messages...
        }
        return res
      }
    ]
  }
})

_ky.get('example.com/api/xx', { silent: false })  // if error occurs, global message will display
_ky.get('example.com/api/xx', { silent: true })   // if error occurs, global message will not display
```

kangkai124:
Fixes #114

### DIFF
--- a/index.js
+++ b/index.js
@@ -209,7 +209,7 @@ class Ky {
 
 			for (const hook of this._hooks.afterResponse) {
 				// eslint-disable-next-line no-await-in-loop
-				const modifiedResponse = await hook(response.clone(), this._options);
+				const modifiedResponse = await hook(this._input, this._options, response.clone());
 
 				if (modifiedResponse instanceof Response) {
 					response = modifiedResponse;
@@ -287,7 +287,7 @@ class Ky {
 	async _fetch() {
 		for (const hook of this._hooks.beforeRequest) {
 			// eslint-disable-next-line no-await-in-loop
-			await hook(this._options);
+			await hook(this._input, this._options);
 		}
 
 		return timeout(fetch(this._input, this._options), this._timeout, this.abortController);
--- a/index.js
+++ b/index.js
@@ -209,7 +209,7 @@ class Ky {
 
 			for (const hook of this._hooks.afterResponse) {
 				// eslint-disable-next-line no-await-in-loop
-				const modifiedResponse = await hook(response.clone());
+				const modifiedResponse = await hook(response.clone(), this._options);
 
 				if (modifiedResponse instanceof Response) {
 					response = modifiedResponse;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -209,7 +209,7 @@ class Ky {
 
 			for (const hook of this._hooks.afterResponse) {
 				// eslint-disable-next-line no-await-in-loop
-				const modifiedResponse = await hook(response.clone());
+				const modifiedResponse = await hook(response.clone(), this._options);
```
szmarczak: This should be documented.

sholladay:
I'm okay with this idea. However, before we do it, we should think about which options people should see in the options object. You might, for example, expect the options object to contain `timeout` and `throwHttpErrors`, but it won't, as implemented by this PR. I'd say that's likely to confuse users, though obviously we can just document it that way. Thoughts?

Also see https://github.com/sindresorhus/ky/issues/100, where I explain a similar issue.

kangkai124:
@sholladay The problem you pointed makes sense. May be we could tell people that options object in afterResponse is not all same with what they passed to ky. For example, we can not get `timeout` by default, and we pass `searchParams object` to ky, but get parsed `query string` in afterResponse. One solution is that we add those params, which useful in afterResponse, to a special object such as called `resOptions`, like this:
```
ky.get(url, {
  timeout: 2000,
  resOptions: {
    silent: false,
    duration: 800
  }
})
```
And if you agree, I am willing to write it to document.

index.js
kangkai124: Readme has been updated.

sholladay:
I've been thinking about this and I'm starting to lean towards having all options be passed to the hooks, so that we can have a simpler API and simpler TypeScript types. We can selectively remove properties before passing them to `fetch()` if it ever becomes necessary, but for now, it isn't.

The next problem is just making sure the documentation is clear about which options the hooks are capable of modifying

szmarczak:
I agree with @sholladay

sholladay:
I updated the code so the hooks now both have `input` and `options` as their first two arguments, just like Ky and `fetch()` have. The `afterResponse` hook has `response` as a third argument. So this is a breaking change.

We still need to update how the options are destructured so all of the options actually end up on that options object passed to the hooks. Will have to do that later.

sindresorhus:
The TypeScript definition also needs to be updated.

sindresorhus:
@kangkai124 Bump

sindresorhus:
Closing in favor of https://github.com/sindresorhus/ky/pull/136.

PR #118, opened 2019-04-03
Set accept header for ky shortcut methods

### LINKED ISSUE #83: Set the `accept` header when shortcut methods are called
Mentioned it in https://github.com/sindresorhus/got/pull/695#issuecomment-453865817

> One thing that got does, which I think it awesome, and which we don't have in ky, is setting the accept header when .json() is called.

@sholladay What's the benefit of sending this header? It works already without.

selrond:
closes #83 

I was first thinking about conditionally setting this inside `ky`s constructor, but it would mean more complexity, as there's no way to know whether the method is set by calling shortcut method or not. We get this for free in `createInstance` function, where shortcut methods are generated, and where I eventually decided to set the header.

### DIFF
--- a/index.js
+++ b/index.js
@@ -63,13 +63,13 @@ const requestMethods = [
 	'delete'
 ];
 
-const responseTypes = [
-	'json',
-	'text',
-	'formData',
-	'arrayBuffer',
-	'blob'
-];
+const responseTypes = {
+	json: 'application/json',
+	text: 'text/*',
+	formData: 'multipart/form-data',
+	arrayBuffer: '*/*',
+	blob: '*/*'
+};
 
 const retryMethods = new Set([
 	'get',
@@ -203,6 +203,7 @@ class Ky {
 		this._options.headers = headers;
 
 		const fn = async () => {
+			await delay(1);
 			let response = await this._fetch();
 
 			for (const hook of this._hooks.afterResponse) {
@@ -224,8 +225,9 @@ class Ky {
 		const isRetriableMethod = retryMethods.has(this._options.method.toLowerCase());
 		const result = isRetriableMethod ? this._retry(fn) : fn();
 
-		for (const type of responseTypes) {
+		for (const [type, mimeType] of Object.entries(responseTypes)) {
 			result[type] = async () => {
+				headers.set('accept', mimeType);
 				return (await result).clone()[type]();
 			};
 		}
--- a/readme.md
+++ b/readme.md
@@ -117,7 +117,7 @@ The `input` and `options` are the same as [`fetch`](https://developer.mozilla.or
 - The `credentials` option is `same-origin` by default, which is the default in the spec too, but not all browsers have caught up yet.
 - Adds some more options. See below.
 
-Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, correct `Accept` header will be set depending on body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range `200...299`.
+Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, proper `Accept` header will be set depending on body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range `200...299`.
 
 ### ky.get(input, [options])
 ### ky.post(input, [options])
--- a/readme.md
+++ b/readme.md
@@ -117,7 +117,7 @@ The `input` and `options` are the same as [`fetch`](https://developer.mozilla.or
 - The `credentials` option is `same-origin` by default, which is the default in the spec too, but not all browsers have caught up yet.
 - Adds some more options. See below.
 
-Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range `200...299`.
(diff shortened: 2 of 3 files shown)


test/methods.js
```diff
@@ -41,3 +41,21 @@ test('custom method remains identical', async t => {
 
 	await server.close();
 });
+
+test('accept header for method shortcuts is application/json', async t => {
+	t.plan(6);
```
szmarczak: No need to use this

index.js
```diff
@@ -302,7 +302,24 @@ const createInstance = (defaults = {}) => {
 	const ky = (input, options) => new Ky(input, deepMerge({}, defaults, options));
 
 	for (const method of requestMethods) {
-		ky[method] = (input, options) => new Ky(input, deepMerge({}, defaults, options, {method}));
+		ky[method] = (input, options = {}) => {
+			const {headers: providedHeaders, ...restOfOptions} = options;
+			const headers = new Headers(providedHeaders);
+			if (!headers.has('Accept')) {
+				headers.set('Accept', 'application/json');
```
szmarczak: This is inappropriate - it should be set only if someone does `ky(...).json()`.

selrond: @szmarczak I've explicitly asked about this [here](https://github.com/sindresorhus/ky/issues/83#issuecomment-478900902) and oyu confirmed it. 
Besides, i don't believe it's technically possible to set accept headers once `json()` is called on the return value from `ky(...)` - the request has already been made right?

sholladay:
Sorry I didn't comment on the issue about this earlier. The expected behavior is for the `Accept` header to be set based on the `responseType`, not the `requestMethod`.

In `got`, this is accomplished by having `.json()` modify the options object used by the main function.

index.js
szmarczak: I have missed this part:

> there needs to be Accept: application/json header set by default.

Anyway as you can read from the OP:

> One thing that got does, which I think it awesome, and which we don't have in ky, is setting the accept header when .json() is called.

So yeah, we need to iterate over every shortcut method and just replace `.json()` I think.

szmarczak:
Hmm... As @selrond did already notice replacing `.json()` won't have any effect because the request has been made.

One of the solutions would be to cancel the previous request and make a new one OR delay making the request.

sholladay:
I haven't looked much into how `got` deals with that part of it. Does it delay the request? If so, how long is it delayed for, one tick of the event loop?

szmarczak:
Yup. It delays the request using `setImmediate`.

sholladay:
Delaying the request seems fine to me, albeit it's a little strange.

Alternatively, we could change the structure of the API to be `await ky.get.json(input, options)` instead of `await ky.get(input, options).json()`. That way Ky knows how to parse the response before the request is sent.

selrond:
@sholladay Delaying the request for the sake of arbitrary API shape seems totally unnacceptable IMHO.

sholladay:
It wouldn't be delayed by any meaningful amount of time. We would just do a simple `setTimeout(fn, 0)`. I agree that is not ideal and it feels a bit icky, but I also don't think it poses any material problem for users. It seems that `got` has been getting by just fine with this approach.

That said, the delay wouldn't be necessary with my proposed API. So thoughts on `await ky.get.json(input, options)`?

szmarczak:
I vote for the `await ky(input, options).json()` approach.
This way we're staying familiar with the `fetch` API.

The delay would be `<=1ms`, not a significant change.

sindresorhus:
> I vote for the await ky(input, options).json() approach.

👍 Me too.

> The delay would be <=1ms, not a significant change.

Yup, it would be neglible.

sindresorhus:
This behavior should also be documented in the readme.

test/methods.js
selrond: eba3de96

selrond:
@szmarczak @sholladay I've been going through got source, but haven't found anything that'd suggest me it's setting accept header when `got().json()` is called.
The only thing I found is this: https://github.com/sindresorhus/got/blob/b5e443b2f9/source/request-as-event-emitter.ts#L350

which suggest, that only checks for option.responseType. Could you please point me to the right direction?

szmarczak:
Place `await delay(1);` before https://github.com/sindresorhus/ky/blob/7114ba310d31c8839668baa4be1225d3c3512c08/index.js#L206

Then I would replace https://github.com/sindresorhus/ky/blob/7114ba310d31c8839668baa4be1225d3c3512c08/index.js#L66-L72

with

```js
const responseTypes = {
	json: 'application/json',
	text: 'text/*',
	formData: 'multipart/form-data',
	arrayBuffer: '*/*',
	blob: '*/*'
};
```

and replace https://github.com/sindresorhus/ky/blob/7114ba310d31c8839668baa4be1225d3c3512c08/index.js#L227-L231

with

```js
 for (const [type, mimeType] of Object.entries(responseTypes)) { 
 	result[type] = async () => { 
                headers.set('accept', mimeType);
 		return (await result).clone()[type](); 
 	}; 
 } 
```

That's all. Remember to test this :)

selrond:
@szmarczak @sholladay pls review again

Concerning the simple test, all other methods behave exactly like `.text()`, but would require correct data to be sent back, which would in case of `multipart/form-data` require additional library like https://github.com/form-data/form-data, so I decided to leave all others out.

Besides, I'm a bit concerned with this type of usage:

```js

const tmp = await ky.get(..)
/* other stuff */
const result = await tmp.json() // headers are going to be pointlessly set here in `json()` call, as tmp already contains promise of result of request that's already been fired
```

The only way it'd work is when one uses `ky.get(..).json()`. Isn't it too hacky?

readme.md
```diff
@@ -117,7 +117,7 @@ The `input` and `options` are the same as [`fetch`](https://developer.mozilla.or
 - The `credentials` option is `same-origin` by default, which is the default in the spec too, but not all browsers have caught up yet.
 - Adds some more options. See below.
 
-Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range `200...299`.
+Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, correct `Accept` header will be set depending on body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range `200...299`.
```
szmarczak:
```suggestion
Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, proper `Accept` header will be set depending on body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range `200...299`.
```

PR #122, opened 2019-04-08
Fix problem with debugging timed-out requests

### LINKED ISSUE #91: `Promise.race()` breaks debugging
I generally develop frontend having "Pause on exceptions" functionality of chrome dev tools enabled. Here Ky doesn't work so good and forced me to disable that feature.
What happens is that determining if the fetch request goes well or it goes in timeout, Ky puts the actual request in race against a promise that on timeout (delay) throws an error. It's a nice looking piece of code and works fine, but when the fetch resolves correctly (in time) the other promise continues on its way and launches an uncontrolled exception that is caught by "Pause on exceptions".

A solution can be removing the `throw` statement form the timeout promise and chaining the result of the promises race with another function that checks which promise won (fetch or timeout), if the latter then throw the error.

arty-name:
Fixes #91

As discussed in #91 the `TimeoutError` is thrown unconditionally in at least two major browsers. This should only happen if the request hadn’t completed before the timeout. 

Note from @szmarczak: provided a shorter workaround.

### DIFF
--- a/index.js
+++ b/index.js
@@ -113,32 +113,14 @@ class TimeoutError extends Error {
 
 const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
 
-const timeout = (promise, ms, abortController) => {
-	let successful = false;
-	return Promise.race([
-		(async () => {
-			try {
-				return await promise;
-			} finally {
-				successful = true;
-			}
-		})(),
-		(async () => {
-			await delay(ms);
-
-			if (successful) {
-				return;
-			}
-
-			if (abortController) {
-				// Throw TimeoutError first
-				setTimeout(() => abortController.abort(), 1);
-			}
-
-			throw new TimeoutError();
-		})()
-	]);
-};
+// `Promise.race()` workaround (#91)
+const timeout = (promise, ms, abortController) => new Promise((resolve, reject) => {
+	promise.then(resolve).catch(reject);
+	delay(ms).then(() => {
+		reject(new TimeoutError());
+		abortController.abort();
+	});
+});
 
 const normalizeRequestMethod = input => requestMethods.includes(input) ? input.toUpperCase() : input;
 
--- a/index.js
+++ b/index.js
@@ -113,18 +113,32 @@ class TimeoutError extends Error {
 
 const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
 
-const timeout = (promise, ms, abortController) => Promise.race([
-	promise,
-	(async () => {
-		await delay(ms);
-		if (abortController) {
-			// Throw TimeoutError first
-			setTimeout(() => abortController.abort(), 1);
-		}
+const timeout = (promise, ms, abortController) => {
+	let successful = false;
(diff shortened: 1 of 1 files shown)


arty-name:
I did consider writing a test for this one but don’t really see a way to check if TimeoutError was instantiated.

index.js
```diff
@@ -113,18 +113,32 @@ class TimeoutError extends Error {
...
+
+			if (abortController) {
+				// Throw TimeoutError first
+				setTimeout(() => abortController.abort(), 1);
+			}
+
+			throw new TimeoutError();
+		})()
+	]);
+};
```
```js
// `Promise.race()` workaround (#91)
const timeout = (promise, ms, abortController) => new Promise((resolve, reject) => {
	promise.then(resolve).catch(reject);
	delay(ms).then(() => {
		reject(new TimeoutError());
		abortController.abort();
	});
});
```

This does the same as `Promise.race()`, but the debugger just omits the error as expected :)

arty-name: Smart and elegant! I’m quite impressed!

I guess with your proposed solution there’s no need for my PR and you can just apply the change by yourself.

szmarczak:
> This should only happen if the request hadn’t completed before the timeout.

I disagree. It "throws", but it really doesn't throw because the promise has already resolved. See my approach. It does the same - the debugger doesn't catch that.

> I did consider writing a test for this one but don’t really see a way to check if TimeoutError was instantiated.

You can't write a failing test for something that works correctly, can you? Note that it isn't our problem - it's Chrome's. Take a look at this [MDN example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race#Using_Promise.race_%E2%80%93_examples_with_setTimeout):

```js
var p3 = new Promise(function(resolve, reject) { 
    setTimeout(() => resolve('three'), 100);
});
var p4 = new Promise(function(resolve, reject) { 
    setTimeout(() => reject(new Error('four')), 500); 
});

Promise.race([p3, p4])
.then(function(value) {
  console.log(value); // "three"
  // p3 is faster, so it resolves
}, function(reason) {
  // Not called
});
```

As you can see, the `onRejection` handler is not called, and thus the debugger shouldn't catch that.

arty-name:
The most important thing to me is that your fix will prevent my problem from happening and I will be thankful for that!

If we go into details though... If it’s a problem in Chrome, then Firefox suffers from it as well. Also keep in mind that there’s a second checkbox called "Pause on caught exceptions" which causes a stop on every throw.

sindresorhus:
> The most important thing to me is that your fix will prevent my problem from happening and I will be thankful for that!

Yup. I agree it's worth having an ugly workaround to fix real use-cases.

> If it’s a problem in Chrome, then Firefox suffers from it as well.

However, this clearly is a DevTools issue in browsers. The existing code is valid JS and should ideally be supported by JS debuggers.

sindresorhus:
Please open issues on Firefox and Chrome ;)

PR #125, opened 2019-04-17
Fix regression for environments without `AbortController`

gmaclennan:
#122 removed the check on whether `abortController` exists before calling `abort()`.
This fixes this check for platforms that don't support abortController (e.g. react-native)

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,7 +118,7 @@ const timeout = (promise, ms, abortController) => new Promise((resolve, reject) 
 	/* eslint-disable promise/prefer-await-to-then */
 	promise.then(resolve).catch(reject);
 	delay(ms).then(() => {
-		if (abortController) {
+		if (supportsAbortController) {
 			abortController.abort();
 		}
 		reject(new TimeoutError());
--- a/index.js
+++ b/index.js
@@ -119,7 +119,7 @@ const timeout = (promise, ms, abortController) => new Promise((resolve, reject) 
 	promise.then(resolve).catch(reject);
 	delay(ms).then(() => {
 		reject(new TimeoutError());
-		abortController.abort();
+		if (abortController) abortController.abort();
 	});
 	/* eslint-enable promise/prefer-await-to-then */
 });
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -118,8 +118,10 @@ const timeout = (promise, ms, abortController) => new Promise((resolve, reject)
 	/* eslint-disable promise/prefer-await-to-then */
 	promise.then(resolve).catch(reject);
 	delay(ms).then(() => {
+		if (abortController) {
```
```suggestion
		if (supportsAbortController) {
```

szmarczak:
Can you test this?

index.js
gmaclennan: timeout() is called with `this.abortController`, which is undefined if `supportsAbortController` is falsey. To avoid bugs down the line I think it's better to test for `abortController` here, not `supportsAbortController` because the method is being called on the instance. But it doesn't really make any difference to the functionality of the code as-is, just a style preference I guess.

gmaclennan: Also did it like this because this was how the test was done before #122

sholladay: Fixed. 👍 

Unrelated: For some reason GitHub's "Commit suggestion" feature said I don't have write access, even though I do (including to the fork). Strange.

gmaclennan:
I tested this on the react-native app I'm working on.

- v0.9.0 works without error
- v0.9.1 throws uncaught error on timeout (which happens whether it times out or not)
- this PR: works without error

sholladay:
To write a test, you would disable the AbortController [polyfill](https://github.com/sindresorhus/ky/blob/7114ba310d31c8839668baa4be1225d3c3512c08/test/_require.js#L7) (e.g. by overwriting `global.AbortController`) and then assert that the promise is rejected with a `TimeoutError` instance and/or assert that `process.on('unhandledRejection')` is _not_ called (to verify there is no `TypeError` from calling `.abort()` on a non-existent function).

gmaclennan:
Cool I won’t have time for this for a while unfortunately. This is just putting back a check that was in 0.9.0, and I can stick with that for now. Would be great if you can merge without a test though, sorry I can’t do more.

index.js
sindresorhus: > I would be fine with removing the supportsAbortController variable altogether in favor of if (abortController) and if (AbortController). To me, the latter is slightly simpler and doesn't really have any serious downsides.

👍

sindresorhus:
A new release is out :)

PR #133, opened 2019-05-09
Simplify the TypeScript types

sindresorhus:


index.d.ts
```diff
@@ -39,7 +33,7 @@ export interface Options extends RequestInit {
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 	*/
-	json?: JSONStringifyable;
```
sindresorhus: This just made the option harder to use in practice. `JSON.stringify()` accepts `unknown` too. And it didn't account for when your have custom types with a `toJSON()` method. So all in all, the strict type had very little value.

index.d.ts
```diff
@@ -122,7 +116,7 @@ export interface ResponsePromise extends Promise<Response> {
 	const result = await ky(…).json<Result>();
 	```
 	*/
-	json<T = JSONValue>(): Promise<T>;
+	json<T = unknown>(): Promise<T>;
```
sindresorhus: Having it default to `JSONValue` doesn't have much value as the whole point of having it strict is to ensure that the user-defined response type is only using JSON-supported types, but that's currently not possible. So you end up having to do `ky() as unknown as CustomType`. Better to leave it as `unknown` until https://github.com/microsoft/TypeScript/issues/15300 is fixed.

Issue #138, opened 2019-06-11
delay causes node process to hang


jamime:
This package and [ky-universal](https://github.com/sindresorhus/ky-universal) is used in [danger-js](https://github.com/danger/danger-js) via the [node-gitlab](https://github.com/jdalrymple/node-gitlab) project.

When ky is used, the node process will not exit until the `setTimout` has completed. 

https://github.com/sindresorhus/ky/blob/07838c79e66b74f4d365b4c5ac82c5c9c18039b8/index.js#L122

I suggest changing the `delay(ms)` function to accept a promise `delay(ms, promise)` when the promise is resolved the timeout can be cancelled. In this instance the promise needs to be the promise copied into timeout.

https://github.com/sindresorhus/ky/blob/07838c79e66b74f4d365b4c5ac82c5c9c18039b8/index.js#L127

This bug was introduced in #122

sindresorhus:
// @arty-name

arty-name:
Since `setTimeout` existed in code before #122 I don’t really understand why it didn’t break things then. 

My instinct would be to avoid using `delay` as a wrapper for setTimeout and instead use setTimeout directly. This would allow calling clearTimeout when the promise is settled. It would be a bit less elegant though. 

Alternatively we could do something like this: 

```javascript
function awaitWithTimeout(promise, timeout) {
  return new Promise((resolve, reject) => {
    const timeoutHandle = setTimeout(() => reject(new TimeoutError()), timeout);
    // eslint-disable-next-line promise/prefer-await-to-then
    promise.then(resolve).catch(reject).then(() => clearTimeout(timeoutHandle));
  })
}

// eslint-disable-next-line promise/prefer-await-to-then
const timeout = (promise, ms, abortController) => awaitWithTimeout(promise, ms).catch((error) => {
  if ((error instanceof TimeoutError) && abortController) {
    abortController.abort();
  }
  throw error
})

jamime:
Ah, my mistake. I thought it was introduced in #122 Looks like this bug did exist before then.

obartra:
looks like `delay` is only used in two spots, I'm happy to make the change `arty-name` suggest if no-one is working on it. I'm running into the same issue @jamime describes

jamime:
@obartra I'm not looking at it, if you have time that would be great.

szmarczak:
@arty-name 

> Alternatively we could do something like this:

[...]

PR #145, opened 2019-06-22
Fix timeout regression that caused Node.js to not terminate

obartra:
Remaining timeout prevents node from terminating the process. These changes clear the timeout on promise resolution.

Closes https://github.com/sindresorhus/ky/issues/138

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,27 +115,34 @@ class TimeoutError extends Error {
 	}
 }
 
-const delay = ms => new Promise((resolve, reject) => {
+function safeTimeout(resolve, ms, reject) {
 	if (ms > 2147483647) { // The maximum value of a 32bit int (see #117)
 		reject(new RangeError('The `timeout` option cannot be greater than 2147483647'));
-	} else {
-		setTimeout(resolve, ms);
 	}
-});
+
+	return setTimeout(resolve, ms);
+}
+
+const delay = ms => new Promise((resolve, reject) => safeTimeout(resolve, ms, reject));
 
 // `Promise.race()` workaround (#91)
-const timeout = (promise, ms, abortController) => new Promise((resolve, reject) => {
-	/* eslint-disable promise/prefer-await-to-then */
-	promise.then(resolve).catch(reject);
-	delay(ms).then(() => {
-		if (supportsAbortController) {
-			abortController.abort();
-		}
+const timeout = (promise, ms, abortController) =>
+	new Promise((resolve, reject) => {
+		const timeoutID = safeTimeout(() => {
+			if (supportsAbortController) {
+				abortController.abort();
+			}
 
-		reject(new TimeoutError());
-	}).catch(reject);
-	/* eslint-enable promise/prefer-await-to-then */
-});
+			reject(new TimeoutError());
+		}, ms, reject);
+
+		/* eslint-disable promise/prefer-await-to-then */
+		promise
+			.then(resolve)
+			.catch(reject)
+			.then(() => clearTimeout(timeoutID));
+		/* eslint-enable promise/prefer-await-to-then */
+	});
 
 const normalizeRequestMethod = input => requestMethods.includes(input) ? input.toUpperCase() : input;
 
--- a/index.js
+++ b/index.js
@@ -140,7 +140,9 @@ const timeout = (promise, ms, abortController) =>
 		promise
 			.then(resolve)
 			.catch(reject)
-			.then(() => clearTimeout(timeoutID));
+			.then(() => {
+				clearTimeout(timeoutID);
+			});
 		/* eslint-enable promise/prefer-await-to-then */
 	});
 
--- a/index.js
+++ b/index.js
@@ -115,15 +115,15 @@ class TimeoutError extends Error {
 	}
 }
 
-function safeTimeout(resolve, ms, reject) {
+const safeTimeout = (resolve, reject, ms) => {
 	if (ms > 2147483647) { // The maximum value of a 32bit int (see #117)
 		reject(new RangeError('The `timeout` option cannot be greater than 2147483647'));
 	}
 
 	return setTimeout(resolve, ms);
-}
+};
 
-const delay = ms => new Promise((resolve, reject) => safeTimeout(resolve, ms, reject));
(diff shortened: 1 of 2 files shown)


szmarczak:
You need to mark `invalid timeout option` as failing and fix `prefixUrl option`.

obartra:
I'm getting all tests to pass locally now but the `prefixUrl` seems to keep failing on CI. I don't see any more output, do you have suggestions as to what may be going on?

image

sindresorhus:
I see more output in Travis. This is the failure details:

```
 /home/travis/build/sindresorhus/ky/node_modules/puppeteer/lib/Launcher.js:351
  TimeoutError {
    message: 'Timed out after 30000 ms while trying to connect to Chrome! The only Chrome revision guaranteed to work is r669486',
  }
```

index.js
```diff
@@ -115,27 +115,34 @@ class TimeoutError extends Error {
 	}
 }
 
-const delay = ms => new Promise((resolve, reject) => {
+function safeTimeout(resolve, ms, reject) {
```
```suggestion
const safeTimeout = (resolve, reject, ms) => {
```

index.js
```diff
@@ -115,27 +115,34 @@ class TimeoutError extends Error {
...
 		reject(new RangeError('The `timeout` option cannot be greater than 2147483647'));
-	} else {
-		setTimeout(resolve, ms);
 	}
-});
+
+	return setTimeout(resolve, ms);
+}
+
+const delay = ms => new Promise((resolve, reject) => safeTimeout(resolve, ms, reject));
```
```suggestion
const delay = ms => new Promise((resolve, reject) => safeTimeout(resolve, reject, ms));
```

test/main.js
```diff
@@ -186,17 +186,31 @@ test('timeout:false option', async t => {
...
-
 	const server = await createTestServer();
 	server.get('/', async (request, response) => {
-		requestCount++;
 		await delay(1000);
 		response.end(fixture);
 	});
 
 	await t.throwsAsync(ky(server.url, {timeout: 21474836470}).text(), RangeError, 'The `timeout` option cannot be greater than 2147483647');
-	t.is(requestCount, 1);
```
szmarczak: Why did you remove this? `requestCount` should be `0` here.

index.js
```diff
@@ -50,7 +50,7 @@ const deepMerge = (...sources) => {
 					value = deepMerge(returnValue[key], value);
 				}
 
-				returnValue = {...returnValue, [key]: value};
+				returnValue = { ...returnValue, [key]: value };
```
szmarczak: Please don't make unrelated changes.

obartra: Yeah, for sure, I forgot to disable prettier, will update

index.js
```diff
@@ -115,27 +115,38 @@ class TimeoutError extends Error {
...
 // `Promise.race()` workaround (#91)
-const timeout = (promise, ms, abortController) => new Promise((resolve, reject) => {
-	/* eslint-disable promise/prefer-await-to-then */
-	promise.then(resolve).catch(reject);
-	delay(ms).then(() => {
-		if (supportsAbortController) {
-			abortController.abort();
-		}
+const timeout = (promise, ms, abortController) =>
+	new Promise((resolve, reject) => {
```
szmarczak: It should be `=> new Promise`. No need to make a new line for that.

index.js
```diff
@@ -115,27 +115,38 @@ class TimeoutError extends Error {
...
+
+		/* eslint-disable promise/prefer-await-to-then */
+		promise
+			.then(resolve)
+			.catch(reject)
+			.then(() => {
+				clearTimeout(timeoutID);
+			});
+		/* eslint-enable promise/prefer-await-to-then */
+	});
```
```suggestion
});
```

index.js
```diff
@@ -115,27 +115,38 @@ class TimeoutError extends Error {
...
+		}, reject, ms);
+
+		/* eslint-disable promise/prefer-await-to-then */
+		promise
+			.then(resolve)
+			.catch(reject)
+			.then(() => {
+				clearTimeout(timeoutID);
+			});
+		/* eslint-enable promise/prefer-await-to-then */
```
szmarczak: This looks much better:
```js
promise.then(resolve, reject).then(() => {
	clearTimeout(timeoutID);
});
```

index.js
```diff
@@ -115,27 +115,38 @@ class TimeoutError extends Error {
 	}
 }
 
-const delay = ms => new Promise((resolve, reject) => {
+const safeTimeout = (resolve, reject, ms) => {
```
szmarczak: Detaching promises is not a good idea. I'd use `try`/`catch` here, but then this piece:

```js
try {
    safeTimeout(resolve, ms);
} catch (error) {
    reject(error);
}
```

would repeat twice. So to follow the DRY principle, I'm good with the current solution :+1:

tiger-seo:
please, merge this, this bug is really annoying

obartra:
Looks like I just needed to rebase, the built error went away now that it's up to date with master.

Regarding the style changes @szmarczak, would you be open to adding [prettier](https://prettier.io/) instead (in a separate PR)? This would remove the need for some of the current linting rules that cause build errors for stylistical / code consistency preferences.

sholladay:
Use `xo --fix` if you want to fix the style issues automatically. It's significantly more opinionated than prettier. In a good way, if you ask me.

obartra:
Oh nice I didn't realize that was an option, would you be open  to adding the`--fix` to the package.json script or as a precommit hook?

sholladay:
We may accept an npm script. IIRC, Sindre has another project that has that, for discoverability.

But it's also very easy for us to take care of it before or after merging.

tiger-seo:
hey guys, @obartra @szmarczak , please please merge this

sindresorhus:
> Regarding the style changes @szmarczak, would you be open to adding prettier instead (in a separate PR)?

No thanks, I have developed a strong dislike for Prettier since I have to waste so much time on people accidentally submitting unwanted style changes because of Prettier.

> Oh nice I didn't realize that was an option, would you be open to adding the--fix to the package.json script or as a precommit hook?

No, you can just run `npx xo --fix`.

PR #147, opened 2019-06-24
Fix passing a `Request` instance as input

### LINKED ISSUE #134: Passing a Request object as first argument to ky fails
I don't know if this is intentional, but the README - and the TypeScript definitions - state that in `ky(input, options)` `ky` accepts the same values as `fetch`.

However, passing a Request instance as the value for `input` does not work, since it's stringified [here](https://github.com/sindresorhus/ky/blob/master/index.js#L171) and it turns into `[object Request]`.

whitecrownclown:
Fixes #134 

Followed up on https://github.com/sindresorhus/ky/issues/134#issuecomment-491044682 and also added a test for it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,6 +23,7 @@ const getGlobal = property => {
 
 const document = getGlobal('document');
 const Headers = getGlobal('Headers');
+const Request = getGlobal('Request');
 const Response = getGlobal('Response');
 const ReadableStream = getGlobal('ReadableStream');
 const fetch = getGlobal('fetch');
--- a/index.js
+++ b/index.js
@@ -157,6 +157,45 @@ class Ky {
 			...otherOptions
 		};
 
+		if (input instanceof Request) {
+			this._input = input;
+
+			// `ky` options have precedence over `Request` options
+			this._options = {
+				...this._options,
+				method: otherOptions.method || input.method,
+				headers: otherOptions.headers || input.headers,
+				body: otherOptions.body || input.body,
+				credentials: otherOptions.credentials || input.credentials
+			};
+		} else {
+			this._input = String(input || '');
+			this._options.prefixUrl = String(this._options.prefixUrl || '');
+
+			if (this._options.prefixUrl && this._input.startsWith('/')) {
+				throw new Error('`input` must not begin with a slash when using `prefixUrl`');
+			}
+
+			if (this._options.prefixUrl && !this._options.prefixUrl.endsWith('/')) {
+				this._options.prefixUrl += '/';
+			}
+
+			this._input = this._options.prefixUrl + this._input;
+
+			if (searchParams) {
+				const url = new URL(this._input, document && document.baseURI);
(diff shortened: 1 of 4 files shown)


sindresorhus:
This doesn't correctly solve it. `Request` is not just the URL, otherwise it would have been a point-less class, as we already have URL. `Request` can contain headers, the body, etc.

https://developer.mozilla.org/en-US/docs/Web/API/Request

sholladay:
Indeed, what I meant to say in the issue is that we'll need to inspect the request object's URL when constructing the `prefuxUrl`. But the request object has other data, such as the headers, that we also need to respect.

whitecrownclown:
> This doesn't correctly solve it. `Request` is not just the URL, otherwise it would have been a point-less class, as we already have URL. `Request` can contain headers, the body, etc.
> 
> [developer.mozilla.org/en-US/docs/Web/API/Request](https://developer.mozilla.org/en-US/docs/Web/API/Request)

@sindresorhus 
How do we want it to work? Deconstruct the `Request` instance into `ky` options or just pass it on as `input`?

To be more specific:

```js
const requestOptions = {headers: {Referer: 'http://google.com'}, method: 'POST'};
const myRequest = new Request('http://example.com', requestOptions);

(async () => {
	await ky(myRequest, {method: 'GET'}); // which method is used?
})();
```

sindresorhus:
> await ky(myRequest, {method: 'GET'}); // which method is used?

`myRequest` should take priority. Options should modify it. We should clearly document this fact though.

I assume that's how `fetch` works too. Can you check how `fetch` behaves in such situation? Doesn't it silently ignore the options, or does it throw an error?

whitecrownclown:
`fetch` does take the options into account. 

> const requestOptions = {headers: {Referer: 'http://google.com'}, method: 'POST'};
const myRequest = new Request('http://example.com', requestOptions);

> (async () => {
	await ky(myRequest, {method: 'GET'}); // which method is used?
})();

`GET` is used here.

sindresorhus:
Hmm, interesting. Then I think we should follow what `fetch` does. We've tried to do that with other features too.

index.js
```diff
@@ -195,6 +195,8 @@ class Ky {
 			this._input = url.toString();
 		}
 
+		this._input = input instanceof Request ? input : this._input;
```
szmarczak: What about moving this to above? @sindresorhus 

```js
if (input instanceof Request) {
    this._input = input;
} else {
    this._input = String(input || '');
    ....
}

**Comment by whitecrownclown** [comment]
@sindresorhus 	This also means that any `url` related options from `ky` don't work when passing in a `Request` since we can't alter the `Request` instance url.

**Comment by sindresorhus** [comment]
Yes, just document that limitation.

File: index.js [hunk 2]
```diff
@@ -157,6 +157,45 @@ class Ky {
 			...otherOptions
 		};
 
+		if (input instanceof Request) {
```
sindresorhus: You need to do `getGlobal` for `Request`.

PR #153, opened 2019-07-23
Move globals to getters to allow them to be mocked

### LINKED ISSUE #151: Does not work when mocking `window.fetch`
This is very bad for testing.
when using `ky` tools like pretender can not work, because `ky` stores a reference to `fetch` immediatly after its loaded with `const fetch = getGlobal('fetch');`.

This means when a tool like `pretender` overwrites `window.fetch` for testing `ky` does not use the new `window.fetch` but the stored reference to the *real* fetch and so does a real request.

Is there a good reason to not call `getGlobal('fetch')` whenever `fetch` is required? (currently only [here](https://github.com/sindresorhus/ky/blob/master/index.js#L325) and [here](https://github.com/sindresorhus/ky/blob/master/index.js#L328))

I'm willing to do a PR, but only if it will be merged.

luxferresum:
This will fix #151

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,24 +4,34 @@ const globals = {};
 
 {
 	const getGlobal = property => {
+		let parent;
+
 		/* istanbul ignore next */
 		if (typeof self !== 'undefined' && self && property in self) {
-			return self[property];
+			parent = self;
 		}
 
 		/* istanbul ignore next */
 		if (typeof window !== 'undefined' && window && property in window) {
-			return window[property];
+			parent = window;
 		}
 
 		if (typeof global !== 'undefined' && global && property in global) {
-			return global[property];
+			parent = global;
 		}
 
 		/* istanbul ignore next */
 		if (typeof globalThis !== 'undefined' && globalThis) {
-			return globalThis[property];
+			parent = globalThis;
 		}
+
+		const globalProperty = parent[property];
+
+		if (typeof globalProperty === 'function') {
+			return globalProperty.bind(parent);
+		}
+
+		return globalProperty;
 	};
 
 	const globalProperties = [
@@ -217,7 +227,7 @@ class Ky {
 			...otherOptions
 		};
 
-		if (input instanceof Request) {
+		if (input instanceof globals.Request) {
 			this._input = input;
 
 			// `ky` options have precedence over `Request` options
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -21,18 +21,27 @@ const getGlobal = property => {
...
 };
 
-const document = getGlobal('document');
-const Headers = getGlobal('Headers');
-const Response = getGlobal('Response');
-const ReadableStream = getGlobal('ReadableStream');
-const fetch = getGlobal('fetch');
-const AbortController = getGlobal('AbortController');
-const FormData = getGlobal('FormData');
+const globals = [
```
szmarczak: Use [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy).

luxferresum: Really? Proxy is not available on all platforms and can not be mocked or transpiled.

sholladay: Personally, I would rather call `getGlobal()` inline where we need it, which will have the same effect. It's a little more verbose, but is much simpler to  read and understand.

szmarczak: @luxferresum

image

Really? You're worried about IE?

@sholladay Yup, that would do the same but IMHO `Proxy` is the way to go. Code is much readable. `globals.URL` > `getGlobal('URL')`

luxferresum: Yes!
Many people still need to support IE!

And I don't think it's much harm to explicit list the globals.

index.js
```diff
@@ -21,18 +21,27 @@ const getGlobal = property => {
...
+	'document',
+	'Headers',
+	'Response',
+	'ReadableStream',
+	'fetch',
+	'AbortController',
+	'FormData'
+].reduce((acc, key) => {
+	Object.defineProperty(acc, key, {
```
sindresorhus: I would prefer using `Object.defineProperties` (note the `s`) and just a normal `for-of` loop.

sindresorhus:
Also make sure that tests are passing.

sindresorhus:
@luxferresum Friendly ping :)

index.js
```diff
@@ -1,46 +1,63 @@
...
+const supportsStreams = typeof globals.ReadableStream === 'function';
+const supportsFormData = typeof globals.FormData === 'function';
 
 const deepMerge = (...sources) => {
 	let returnValue = {};
 
 	for (const source of sources) {
 		if (Array.isArray(source)) {
-			if (!(Array.isArray(returnValue))) {
+		  if (!(Array.isArray(returnValue))) {
```
```suggestion
			if (!(Array.isArray(returnValue))) {
```

szmarczak: image

but I can edit the file directly... hmm...

PR #159, opened 2019-08-06
Add `beforeRetry` hook

lambdalisue:
Add `beforeRetry` hook just like [got](https://github.com/sindresorhus/got#hooksbeforeretry). I needed it to make sure that an access token is fresh prior to retry.

#### Example usage

```typescript
export const client = ky.extend({
  hooks: {
    beforeRequest: [
      async function(input, options) {
        if (store.refreshRequired) {
          await store.refresh();
        }
        if (store.isAuthorized) {
          (options.headers as Headers).set(
            'X-Authorization',
            `Bearer ${store.accessToken}`,
          );
        }
      },
    ],
    beforeRetry: [
      async function(input, options, errors, retryCount) {
        if (store.refreshRequired) {
          await store.refresh();
        }
        if (store.isAuthorized) {
          (options.headers as Headers).set(
            'X-Authorization',
            `Bearer ${store.accessToken}`,
          );
        }
      },
    ],
  },
});
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,6 +13,13 @@ export type BeforeRequestHook = (
 	options: NormalizedOptions,
 ) => void | Promise<void>;
 
+export type BeforeRetryHook = (
+	input: Input,
+	options: NormalizedOptions,
+	error: Error,
+	retryCount: number,
+) => void | Promise<void>;
+
 export type AfterResponseHook = (
 	input: Input,
 	options: NormalizedOptions,
@@ -38,6 +45,15 @@ export interface Hooks {
 	@default []
 	*/
 	beforeRequest?: BeforeRequestHook[];
+
+	/**
+	Before the request is retried.
+
+	This hook enables you to modify the request right before retry. The hook function receives the normalized options as the first argument. You could, for example, modify `options.headers` here.
+
+	@default []
+	*/
+	beforeRetry?: BeforeRetryHook[];
 
 	/**
 	After the response is received.
--- a/index.d.ts
+++ b/index.d.ts
@@ -49,7 +49,7 @@ export interface Hooks {
 	/**
 	Before the request is retried.
 
-	This hook enables you to modify the request right before retry. The hook function receives the normalized options as the first argument. You could, for example, modify `options.headers` here.
+	This hook enables you to modify the request right before retry. The hook function receives the `input`, `options`, `error` and `retryCount` arguments. You could, for example, modify `options.headers` here.
 
 	@default []
 	*/
--- a/test/hooks.js
+++ b/test/hooks.js
@@ -291,3 +291,139 @@ test.failing('`afterResponse` hook is called with input, normalized options, and
 
 	await server.close();
 });
+
+test('beforeRetry hook is never called for the initial request', async t => {
+	const fixture = 'fixture';
+	const server = await createTestServer();
+	server.get('/', async (request, response) => {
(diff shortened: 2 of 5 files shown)


test/hooks.js
```diff
@@ -245,3 +246,49 @@ test('`afterResponse` hook gets called even if using body shortcuts', async t =>
...
+					options.headers.set('unicorn', fixture);
+				}
+			]
+		}
+	}).text(), fixture);
+
+	await server.close();
+});
+
+test('beforeRetry hook allows modifications of non initial requests', async t => {
```
sindresorhus:
```suggestion
test('beforeRetry hook allows modifications of non-initial requests', async t => {
```

test/hooks.js
```diff
@@ -245,3 +246,49 @@ test('`afterResponse` hook gets called even if using body shortcuts', async t =>
...
+			response.end(request.headers.unicorn);
+		} else {
+			response.sendStatus(408);
+		}
+	});
+
+	t.is(await ky.get(server.url, {
+		hooks: {
+			beforeRetry: [
+				options => {
```
sindresorhus: Test the other arguments too, and that the retryCount is correct.

test/hooks.js
```diff
@@ -245,3 +246,49 @@ test('`afterResponse` hook gets called even if using body shortcuts', async t =>
...
+					options.headers.set('unicorn', fixture);
+				}
+			]
+		}
+	}).text(), fixture);
+
+	await server.close();
+});
```
sindresorhus: Make sure you test as much as done for `beforeRetry` in Got: https://github.com/sindresorhus/got/blob/master/test/hooks.ts

lambdalisue:
I'll wait https://github.com/sindresorhus/ky/pull/163

sindresorhus:
@lambdalisue Can you fix the merge conflict and finish this PR?

lambdalisue:
It was a bit difficult to keep previous commits so I rebased from current master and force-pushed.
Please re-review.

lambdalisue:
@sindresorhus So any updates?

szmarczak:
Travis is very slow, so this is why Puppeteer fails. Just ignore it. The PR looks good.

index.d.ts
```diff
@@ -39,6 +46,15 @@ export interface Hooks {
 	*/
 	beforeRequest?: BeforeRequestHook[];
 
+	/**
+	Before the request is retried.
+
+	This hook enables you to modify the request right before retry. The hook function receives the `input`, `options`, `error` and `retryCount` arguments. You could, for example, modify `options.headers` here.
```
sindresorhus: This should have the exact same text and example as the readme. This makes for a better TS experience, and having them in sync makes it easier to maintain both.

// @szmarczak

sindresorhus: Looks like there's more stuff that's out of sync: https://github.com/sindresorhus/ky/commit/8ab3886a4b57ccf8e6c2ac55a8046c82f3ec211d

szmarczak: Sorry, forgot about syncing these. There should be a tool that converts automatically TS docs into markdown, so we don't have to update the same info twice.

sindresorhus: Ideally, an GitHub Action, but for now, we just have to try to make sure all PRs update index.d.ts with both methods and docs and examples.

sholladay: I think a PR template with a checklist would be a good idea until it can be automated.

PR #160, opened 2019-08-07
Expose Options interface

jas0ncn:
- Expose `OptionsWithoutBody` and `OptionsWithBody`
- `patch` has body!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -92,12 +92,12 @@ export interface Options extends RequestInit {
 	onDownloadProgress?: (progress: DownloadProgress, chunk: Uint8Array) => void;
 }
 
-interface OptionsWithoutBody extends _Omit<Options, 'body'> {
+export interface OptionsWithoutBody extends _Omit<Options, 'body'> {
 	method?: 'get' | 'head'
 }
 
-interface OptionsWithBody extends Options {
-	method?: 'post' | 'put' | 'delete'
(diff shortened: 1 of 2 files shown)


joaovieira:
@jas0ncn just noticed we've done different things for the same problem. See https://github.com/sindresorhus/ky/pull/162

I'd rather only export a set of `Options` as it's already the case. But make sure that `ky()` supports them. In my solution you can use `Options` directly as so:

```ts
import ky, { Options } from 'ky';

const postOptions: Options = {
  method: 'post',
  json: { ok: true }  
}; // OK

const getOptions: Options = {
  method: 'get',
  json: { ok: true }
}; // OK

ky('https://example.com', postOptions); // OK
ky('https://example.com', getOptions); // Error
```

jas0ncn:
Yes, I know. But if I want to do the same thing on each method request, I can't  define options type on the function I create. For example:

```typescript
import ky, { Options } from 'ky'

export async function request (url: string, options: Options) {
  try {
    const res = ky(url, options) // options type error
    return [res.json(), e]
  } catch (e) {
    return [null, e]
  }
}
```

jas0ncn:
Duplicate of #162

joaovieira:
Actually, didn't work as I expected 😬 but I've updated and should be good now. I've export `Input` and the merged `Options = OptionsWithoutBody | OptionsWithBody`.

Issue #139, opened 2019-06-14
Document how to override headers in beforeRequest hook


kraniebrud:
Hi, 
thanks for a fantastic lib !
I am trying to modify headers with the `beforeRequest` in `hooks`, but I can't make it work.

I am having the following code

```js
import ky from 'ky-universal'

const prefixUrl = 'https://jsonplaceholder.typicode.com'

const beforeRequestFns = [
  (options) => {
    console.log(options.headers)
    options.headers.unicorn = 'rainbow'
  }
]

const options = {
  prefixUrl,
  hooks: {
    beforeRequest: beforeRequestFns,
  },
}

const client = ky.create({ ...options })

client.extend({ headers: { foo: 'bar' } }).get('todos/1')

export default client

```
From the `Network` tab in Chrome I don't see my 'unicorn' in the request headers, I see the 'foo' is there alright :)

What I am trying to achieve is that every request towards the defined `prefixUrl` is supposed to be able add authentication to the headers do to some conditions.

From what I can tell from the readme this should be possible, so I am guessing that I am just doing it wrong ? 

Can you help? :) -- thanks

sholladay:
Yes, this is possible to do, it's just not entirely obvious at the moment.

Ky [normalizes the headers](https://github.com/sindresorhus/ky/blob/2ded27572ec491e6c9e594ebcd2f94cc8c3957db/index.js#L205) internally to be a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance, so that we can deal with them in a consistent way.

This is briefly mentioned in [`beforeRequest` documentation](https://github.com/sindresorhus/ky#hooksbeforerequest):
> The hook function receives the normalized options as the first argument.

At the moment, we don't have much documentation related to the normalization process. But in this case, the solution is simply to set the header using the appropriate methods of the `Headers` class, as assigning a property to the instance won't do anything.

```js
hooks: {
    beforeRequest : [
        (options) => {
            options.headers.set('unicorn', 'rainbow');
        }
    ]
}
```

Since we already mention overriding headers in the documentation, it would be good to add this as a working example. PRs welcome!

kraniebrud:
Wauv so close 🤦‍♂ , thanks for the reply ! I will try it out when I find the time to refactor my current code at some point.

joaovieira:
[...]

PR #161, opened 2019-08-09
Fix the `BeforeRequestHook` TypeScript type

joaovieira:
`hooks.beforeRequest` hooks do not receive `Options` but a new set of normalized options with a combination of fetch and ky options, plus a more strict `headers` type.

Re: https://github.com/sindresorhus/ky/issues/139


sindresorhus:
Can you fix the merge conflict?

joaovieira:
@sindresorhus done 👍

PR #162, opened 2019-08-09
Fix the options TypeScript type

joaovieira:
The union of `OptionsWithBody` and `OptionsWithoutBody`, both not exported, makes it impossible to reuse an options like so:

```ts
const options: Options = {
  method: 'get'
};
ky(url, options);
```

or if you want to wrap `ky` in your own function:

```ts
function httpClient(input: Input, options: Options) {
  return ky(input, options);
}
```

Given that you can't assign the default `string` method (defined in `Options`) to `'get' | 'head' | 'post' | ...'` expected by the `ky` function.

Screen Shot 2019-08-09 at 16 23 40

Thus, I've exported and refactored both the `Input` and `Options` that make the API of `ky` so one can reuse without conflicts.

----

However, the decision to group these two types of options has the drawback of not supporting other custom HTTP methods (e.g. `trace`, `my-custom-method`). That was already the case, but don't really see the benefit as it's not restricted by the implementation. I.e. in plain JavaScript I can do:

```js
ky(url, {
  method: 'get',
  body: 'test'
});
```

without crashing. Fetch silently handles it.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -96,10 +96,6 @@ interface OptionsWithoutBody extends _Omit<Options, 'body'> {
 	method?: 'get' | 'head'
 }
 
-interface OptionsWithBody extends Options {
-	method?: 'post' | 'put' | 'delete'
-}
-
 /**
 Returns a `Response` object with `Body` methods added for convenience.
 */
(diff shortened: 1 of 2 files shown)


jas0ncn:
Agree!!! Your solution is better then mine.

sindresorhus:
Thanks for fixing this :)

PR #136, opened 2019-06-09
Access options to request hooks (continues PR #115)

tasoskakour:
Continuing the PR to finish work in https://github.com/sindresorhus/ky/pull/115, by updating the tests.

PR #163, opened 2019-08-13
Pass input/options to beforeRequest/afterResponse hooks

lambdalisue:
I needed to access `input` and `options` on `afterResponse` hook to perform retry with a fresh access  token like:

```js
import ky from 'ky';

(async () => {
  await ky.get('https://example.com', {
    hooks: {
      afterResponse: [
	async (input, options, response) => {
	  if (response.status === 403) {
	    // Get a fresh token
	    const token = await ky.get('https://example.com/token').text();
	    // Retry with the token
	    options.headers.set('Authorization', `token ${token}`);
	    return ky(input,  {
	      ...options,
	    });
	  }
	}
      ]
    }
  });
})();
```

I noticed https://github.com/sindresorhus/ky/pull/136 after I wrote the code but the PR seems not updated and ~backword incompatible (arguments order)~ so I decided to make this PR.

**Updated**

- [x] The order become `input, options, response`
- [x] Applied to `beforeRequest` as well

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,11 @@ type Except<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
 export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<void>;
 
-export type AfterResponseHook = (response: Response) => Response | void | Promise<Response | void>;
+export type AfterResponseHook = (
+  response: Response,
+  input: Input,
+  options: NormalizedOptions,
+) => Response | void | Promise<Response | void>;
 
 export interface DownloadProgress {
 	percent: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,9 +5,9 @@ type Except<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<void>;
 
 export type AfterResponseHook = (
-	response: Response,
 	input: Input,
 	options: NormalizedOptions,
+	response: Response,
 ) => Response | void | Promise<Response | void>;
 
 export interface DownloadProgress {
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,10 @@
 
 type Except<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
-export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<void>;
+export type BeforeRequestHook = (
+	input: Input,
+	options: NormalizedOptions,
+) => void | Promise<void>;
 
 export type AfterResponseHook = (
 	input: Input,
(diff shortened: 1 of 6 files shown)


sholladay:
I am open to having the arguments in this order, I can definitely understand the desire to have `response` be first. But I also think the consistency of `input` and `options` always being first for all hooks is kind of nice. You can make a more reusable hook function if we do it that way. I lean slightly in favor of that consistency. I also noticed that a lot of users want to write hooks (including `afterResponse`) thst check the `input` and `options` in order to change their behavior. So I think having `response` be last won't be as inconvenient as you are imagining.

Either way, thanks for the alternative PR. I'll review it fully when I'm a bit more awake.

lambdalisue:
> But I also think the consistency of input and options always being first for all hooks is kind of nice.

I see. Actually, I agree that applying `input/options` first is more natural for `beforeRequest`  hook so I  understand the benefit of the argument order.

I don't have strong favor for the argument order so I'm very welcome to change the argument order in that way (and/or appling it to `beforeRequest` as well like https://github.com/sindresorhus/ky/pull/136).

sindresorhus:
> I agree that applying input/options first is more natural for beforeRequest hook so I understand the benefit of the argument order.

👍 Let's do it that way.

> appling it to beforeRequest as well like #136)

👍

index.d.ts
```diff
@@ -4,7 +4,11 @@ type Except<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
 export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<void>;
 
-export type AfterResponseHook = (response: Response) => Response | void | Promise<Response | void>;
+export type AfterResponseHook = (
+  response: Response,
+  input: Input,
+  options: NormalizedOptions,
```
sindresorhus: Tab-indentation

readme.md
```diff
@@ -226,6 +226,19 @@ import ky from 'ky';
...
+				// Or retry with a fresh token on a 403 error
+				async (response, input, options) => {
+					if (response.status === 403) {
+						// Get a fresh token
+						const token = await ky.get('https://example.com/token').text();
+						
+						// Retry with the token
+						options.headers.set('Authorization', `token ${token}`);
+						
+						return ky(input, {...options});
```
sindresorhus:
```suggestion
						return ky(input, options);
```

?

lambdalisue:
Done. Please re-review 👍

alexcroox:
Will this allow me to change the prefixUrl and url params in this way?

```
class LegacyApi extends ApiCore {
  constructor(model) {
    super(model)

    // Before the API call is made, change the options to point
    // to the legacy API and append the required URL params
    this.hooks.beforeRequest.push(
      (input, options) => {
        console.log({ options })
        options.prefixUrl = process.env.VUE_APP_LEGACY_API_BASE_URL
        options.searchParams = {
          ...options.searchParams,
          "HID": "sellmylivestock"
        }
      }
    )
  }
}
```

 I have an API class that I'd like to use for different API calls that are on different URLs. Hooks in ky seemed a clean way of supporting this without having to set options in my api class for the prefixUrl.

PR #165, opened 2019-08-19
Simplify the `Options` TypeScript type

### LINKED ISSUE #164: TypeScript - Union `Options` complicates `ky` API
I've seen that was added in https://github.com/sindresorhus/ky/pull/87 to "fix" https://github.com/sindresorhus/ky/issues/85. While it works well on its own (i.e. when invoking `ky` or `ky.get`/`ky.post` shorthands), it complicates the usage of the `Options` type if you ever need to manipulate them.

Take my example, we're wrapping `ky` as we'd like to convert the `body`/`json` keys to snake case before sending to the server and vice-versa on response. Something like:

```ts
import { Input, Options } from 'ky';

export default async function (input: Input, options?: Options) {
  const kyOptions = { ...options };
  kyOptions.json = snakeCase(kyOptions.json); // error: property `json` does not exist on type `OptionsWithoutBody`
  
  try{
    const resp = await ky(input, kyOptions);
    return camelCase(resp.json);
  } catch (e) {
    ...
  }
}
```

However, given the current type of `Options = OptionsWithoutBody | OptionsWithBody` we cannot use `kyOptions.json` as it does not exist in `OptionsWithoutBody`. The compiler only allows access to that property once it knows it's of type `OptionsWithBody`, like so:

```ts
if (kyOptions.method && ['post', 'put', 'patch', 'delete'].includes(kyOptions.method.toLowerCase())) {
  kyOptions.json = snakeCase(kyOptions.json);
}
```

This is not ideal as the application would need to implement the same logic as `ky` to differentiate options with and without body based on the method.

---

On top of that, using an union type makes extending the options awkward. When using interfaces you do:

```ts
interface Options extends KyOptions { 
  snakeCaseTransform: boolean;
}
```

With union types it becomes:

```ts
type Options = KyOptions & {
  snakeCaseTransform: boolean;
}
```

---

As I also mentioned in #162, having this strict type prevents custom or non-whitelisted methods (e.g. `trace`), that are allowed by the spec and `fetch` itself:

```ts
ky('https://example.com', {
  method: 'custom',
});
```

Lastly, it is adding something not enforced by the implementation. The native `fetch` types don't do that discrimination.

---

Thus I want to propose to essentially revert https://github.com/sindresorhus/ky/pull/87 and use `Options` everywhere. This makes the API and types much simpler, easier to read and extend.

Thoughts? Happy to do it if agreed.

joaovieira:
Fixes #164 

See discussion in issue above.

This PR:
* makes reusing, extending and manipulating `Options` type more flexible;
* adds support for custom HTTP verbs/methods;
* aligns `ky`'s API with the native `fetch`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,10 @@
 /// <reference lib="dom"/>
 
+type Primitive = null | undefined | string | number | boolean | symbol | bigint;
+type LiteralUnion<LiteralType extends BaseType, BaseType extends Primitive> =
+	| LiteralType
+	| (BaseType & { _?: never });
+
 export type Input = Request | URL | string;
 
 export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<void>;
@@ -40,6 +45,14 @@ export interface Hooks {
 Options are the same as `window.fetch`, with some exceptions.
 */
 export interface Options extends RequestInit {
+	/**
+	HTTP request method.
+	*/
+	method?: LiteralUnion<
+		'get' | 'head' | 'post' | 'put' | 'delete' | 'connect' | 'options' | 'trace' | 'patch',
+		string
+	>;
+
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 	*/
@@ -192,6 +205,14 @@ declare const ky: {
 	get(url: Input, options?: Options): ResponsePromise;
 
 	/**
+	Fetch the given `url` using the option `{method: 'head'}`.
+
+	@param url - `Request` object, `URL` object, or URL string.
+	@returns A promise with `Body` methods added.
+	*/
+	head(url: Input, options?: Options): ResponsePromise;
+
+	/**
 	Fetch the given `url` using the option `{method: 'post'}`.
 
 	@param url - `Request` object, `URL` object, or URL string.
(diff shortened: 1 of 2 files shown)


sindresorhus:
I think it's still useful to have auto-complete for the `method` option. You can copy-paste [this](https://github.com/sindresorhus/type-fest/blob/master/source/literal-union.d.ts) type so we can have auto-complete and also the flexibility. Add the other methods mention in https://fetch.spec.whatwg.org/#methods

joaovieira:
@sindresorhus is there any reason to copy-paste rather than import from `type-fest`?

index.d.ts
```diff
@@ -201,47 +202,71 @@ declare const ky: {
...
-	get(url: Input, options?: OptionsWithoutBody): ResponsePromise;
+	get(url: Input, options?: Options): ResponsePromise;
+
+	/**
+	Fetch the given `url` using the option `{method: 'head'}`.
+
+	@param url - `Request` object, `URL` object, or URL string.
+	@returns A promise with `Body` methods added.
+	*/
+	head(url: Input, options?: Options): ResponsePromise;
```
joaovieira: Just some reordering to align with the `method` literal.

sindresorhus:
>  is there any reason to copy-paste rather than import from type-fest?

I don't want to introduce any dependencies for Ky.

index.d.ts
```diff
@@ -201,47 +202,71 @@ declare const ky: {
...
+	delete(url: Input, options?: Options): ResponsePromise;
 
 	/**
-	Fetch the given `url` using the option `{method: 'head'}`.
+	Fetch the given `url` using the option `{method: 'connect'}`.
+
+	@param url - `Request` object, `URL` object, or URL string.
+	@returns A promise with `Body` methods added.
+	*/
+	connect(url: Input, options?: Options): ResponsePromise;
```
sindresorhus: I don't want methods for `connect`, `options`, `trace`. They are not common.

index.d.ts
```diff
@@ -39,7 +44,15 @@ export interface Hooks {
 /**
 Options are the same as `window.fetch`, with some exceptions.
 */
-interface KyOptions extends RequestInit {
+export interface Options extends RequestInit {
+	/**
+	HTTP request method.
+	*/
+	method?: LiteralUnion<
+		'get' | 'head' | 'post' | 'put' | 'delete' | 'connect' | 'options' | 'trace' | 'patch',
```
sindresorhus: I prefer the ordering to be sorted by common usage: get, post, put, delete, patch, head, connect, options, trace.

Also put them on separate lines.

index.d.ts
joaovieira: Ha! Was just removing them. The fetch spec says `forbidden methods` as well, though they are in the HTTP spec.

joaovieira:
@sindresorhus do you think you can publish a new release with this (and whatever is new in master)? Thanks.

sindresorhus:
Done: https://github.com/sindresorhus/ky/releases/tag/v0.13.0

PR #169, opened 2019-08-27
Fix URL processing

piouc:
Number 0 is not invalid value.
ky(0) expected request to 'http://example.org/0', but currently request 'http://example.org/'.

### DIFF
--- a/index.js
+++ b/index.js
@@ -179,8 +179,8 @@ class Ky {
 				credentials: otherOptions.credentials || input.credentials
 			};
 		} else {
-			this._input = String(input || '');
-			this._options.prefixUrl = String(this._options.prefixUrl || '');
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think we should throw a useful error instead if it’s not a valid input type.

piouc:
How about use toString() for all types includes falsy value as with fetch?

sindresorhus:
No, I prefer strict input types, which from experience, catches a lot of invalid input and obscure errors instead of silently accepting anything. You can do `String(unknownInput)` yourself if you're not sure what the values might be.

sholladay:
In addition to what Sindre said, we couldn't blindly stringify everything, because we also need to support [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) instances.

Would you like to add the type checking?

Issue #168, opened 2019-08-23
Fix Travis


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/ky/issues/168)

https://travis-ci.com/sindresorhus/ky/jobs/227726270

Travis is slow. I think we might just need to increase the Puppeteer timeout somehow.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
### Submitted pull Requests
- [#177 Fix Travis CI - Serialize Browser Test](https://issuehunt.io/r/sindresorhus/ky/pull/177)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/ky/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

sholladay:
The three most recent Travis builds all failed in exactly the same way, always a TimeoutError and always on Node 10 but not Node 12. Seems kinda fishy, no? Also, it says the timeout is 30 seconds which seems pretty reasonable to me.

sindresorhus:
Could be, but from experience, it's almost always Travis being slow, especially when the problem is only on older Node.js versions.

szmarczak:
> Could also be a race issue in the code or test.

It's not definitely a race issue. I've experienced it also when developing `http2-wrapper`.

Ybrin:
I will give this a try

PR #172, opened 2019-08-29
Allow returning `Response` from a `beforeRequest` hook

### LINKED ISSUE #157: Skipping request by returning a Response in beforeRequest
This could be useful for internal caching, mocking, etc..
Would this be an acceptable feature ?

hugomrdias:
This PR adds support to return a Response from the beforeRequest hook and completely skip the HTTP request.
 
closes #157 

/cc @lidel

### DIFF
--- a/index.js
+++ b/index.js
@@ -383,7 +383,10 @@ class Ky {
 	async _fetch() {
 		for (const hook of this._hooks.beforeRequest) {
 			// eslint-disable-next-line no-await-in-loop
-			await hook(this._input, this._options);
+			const hookOutput = await hook(this._input, this._options);
+			if (hookOutput instanceof Response) {
+				return hookOutput;
+			}
 		}
 
 		if (this._timeout === false) {
--- a/index.js
+++ b/index.js
@@ -383,10 +383,10 @@ class Ky {
 	async _fetch() {
 		for (const hook of this._hooks.beforeRequest) {
 			// eslint-disable-next-line no-await-in-loop
-			const hookOutput = await hook(this._input, this._options);
+			const result = await hook(this._input, this._options);
 
-			if (hookOutput instanceof Response) {
-				return hookOutput;
+			if (result instanceof Response) {
+				return result;
 			}
 		}
 
--- a/index.js
+++ b/index.js
@@ -384,6 +384,7 @@ class Ky {
 		for (const hook of this._hooks.beforeRequest) {
 			// eslint-disable-next-line no-await-in-loop
 			const hookOutput = await hook(this._input, this._options);
+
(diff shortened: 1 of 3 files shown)


hugomrdias:
@sindresorhus can you restart travis please ? the failed build seems to be some random puppeteer timeout.

sindresorhus:
@hugomrdias `master` is currently failing on Node.js 10 too, so don't worry about that. See #168.

hugomrdias:
cool then this is ready for review

index.js
```diff
@@ -383,7 +383,10 @@ class Ky {
 	async _fetch() {
 		for (const hook of this._hooks.beforeRequest) {
 			// eslint-disable-next-line no-await-in-loop
-			await hook(this._input, this._options);
+			const hookOutput = await hook(this._input, this._options);
+			if (hookOutput instanceof Response) {
```
```suggestion

			if (hookOutput instanceof Response) {
```

index.js
```diff
@@ -383,7 +383,10 @@ class Ky {
 	async _fetch() {
 		for (const hook of this._hooks.beforeRequest) {
 			// eslint-disable-next-line no-await-in-loop
-			await hook(this._input, this._options);
+			const hookOutput = await hook(this._input, this._options);
```
szmarczak: `hookOutput` -> `result`

readme.md
```diff
@@ -226,6 +226,8 @@ Default: `[]`
 
 This hook enables you to modify the request right before it is sent. Ky will make no further changes to the request after this. The hook function receives normalized input and options as arguments. You could, for example, modify `options.headers` here.
 
+A [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) can be returned from a beforeRequest hook and completely avoid making an HTTP request. This can be used to mock a request, check an internal cache, etc. An **IMPORTANT** consideration when returning a Response from this hook is that, all the following hooks will be skipped, so be **sure to only return a Response from the last one**.
```
szmarczak:
```suggestion
A [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) instance can be returned from a `beforeRequest` hook to completely avoid making an HTTP request. This is very helpful when mocking a request or using an internal cache.

**Note:** all the following hooks will be skipped.
```

sindresorhus: This change makes it sound like the following hooks will be skipped, just by using this hook, which is not correct.

sindresorhus:
The docs part should have been added to index.d.ts. Otherwise, looks good.

donifer:
I think the types should also be updated 🤔 can't get TS to accept returning a `Response`

PR #176, opened 2019-09-16
Validate the `input` argument

### LINKED ISSUE #174: Validate input
Moved from #169

As of now, the input is not validated. It should throw if it's invalid (e.g. a number).

himynameisdave:
Display a meaningful error to the user if they attempt to provide an invalid input (for example, a `Number`).

Resolves #174

### DIFF
--- a/index.js
+++ b/index.js
@@ -242,6 +242,8 @@ class Ky {
 				body: otherOptions.body || input.body,
 				credentials: otherOptions.credentials || input.credentials
 			};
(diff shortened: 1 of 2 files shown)


himynameisdave:
@szmarczak I have some failing tests on CI, but they appear to be breaking on `master` as well...

szmarczak:
Ignore that. #168 - Travis is just too slow. You did it pretty good 👌

sindresorhus:
Nice work :)

Issue #167, opened 2019-08-23
`afterResponse` hook doesn't work with the new expanded `retry` option


sindresorhus:
Introduced by https://github.com/sindresorhus/ky/commit/d4ddb7ab4ec0d5baff73b3c038be48717592384d.

This test now fails: https://github.com/sindresorhus/ky/blob/a50948adeff0c2255626ed28465e126e095a7c04/test/hooks.js#L249

```
 hooks › `afterResponse` hook is called with input, normalized options, and response which can be used to retry

  /Users/sindresorhus/dev/oss/ky/index.js:3

   2:
   3: const getGlobal = property => {
   4:   /* istanbul ignore next */

  Rejected promise returned by test. Reason:

  Error {
    message: 'retry.methods must be an array',
  }

  normalizeRetryOptions (index.js:3:1536)
  new Ky (index.js:4:7)
  ky (index.js:8:1713)
  t.deepEqual.ky.post.hooks.afterResponse (test/hooks.js:275:15)
  fn (index.js:6:55)
```

// @whitecrownclown @lambdalisue

whitecrownclown:
The issue here is that the `afterResponse` hook uses the already normalized options for `retry` and `methods` (`methods` is a `Set`). An error is thrown since they go through the same normalize process again inside this test when calling `ky`.

@sindresorhus
What do you think about not throwing [an error](https://github.com/sindresorhus/ky/pull/110#discussion_r317104937), but a promise rejection and also allow using both `Array` and `Set` as options?

sholladay:
So, this may need a bit more brainstorming...

PR #180 fixed the failing test. It did so by filtering the options passed to hooks such that hooks only receive fetch/unknown options but no Ky options. In other words,`retry` and other Ky options are omitted from the `options` passed to hooks, so when the hook in that test gives its `options` back to Ky, no error is thrown, because `retry` is not there to be validated.

This works well enough and perhaps we should keep it that way. I've gone back and forth multiple times on whether I think hooks should receive Ky options. This issue, and the test being discussed here, made me think, "Yeah, hooks should receive Ky options." But when I went to go implement that with all Ky options included, I ran into a new problem: `json` and `body` are then both present and Ky throws an error because it considers those options to be mutually exclusive. This problem was never experienced before because prior to PR #180, some Ky options weren't given to hooks (including `json`), while others were. I'm not sure any intentional thought was put into it, the codebase just grew that way organically.

Some ways forward from here:
[...]

PR #180, opened 2019-10-01
Use request to normalize options

### LINKED ISSUE #139: Document how to override headers in beforeRequest hook

### LINKED ISSUE #166: Modify the request URL in the `beforeRequest` hook
I want to edit request url in the beforeRequest hook.
Thx.

sholladay:
Closes #166
Closes #139

This PR:
 - Aims to simplify Ky and solve some edge case issues and confusing behavior related to how our options are normalized and merged.
 - Makes it so that internally, and within hooks, a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) instance is now used as the source of truth instead of `input`. This replaces some of the bespoke logic that we've had for processing options. It also makes life easier for people using the `hooks` option, because they will receive a normalized `request`, whose behavior is more obvious and is well-specified by web standards. We can discuss whether the raw input should be available as well.
 - Cleans up the separation betwen Ky options and `fetch` options internally, so there's a clear distinction and it's easy to pass around one or the other, or both.
 - Adds the ability for `beforeRequest` hooks to return a `Request` instance to replace the existing request.

### DIFF
--- a/index.js
+++ b/index.js
@@ -214,110 +214,102 @@ const normalizeRetryOptions = retry => {
 
 class Ky {
 	constructor(input, {
-		timeout = 10000,
 		hooks,
-		throwHttpErrors = true,
-		searchParams,
 		json,
+		onDownloadProgress,
+		prefixUrl,
 		retry = {},
-		...otherOptions
+		searchParams,
+		throwHttpErrors = true,
+		timeout = 10000,
+		...fetchOptions
 	}) {
 		this._retryCount = 0;
-
+		this._input = input;
 		this._options = {
-			method: 'get',
-			credentials: 'same-origin', // TODO: This can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
+			hooks: deepMerge({
+				beforeRequest: [],
+				beforeRetry: [],
+				afterResponse: []
+			}, hooks),
+			json,
+			onDownloadProgress,
+			prefixUrl: String(prefixUrl || ''),
 			retry: normalizeRetryOptions(retry),
-			...otherOptions
+			searchParams,
+			throwHttpErrors,
+			timeout
+		};
+		this._fetchOptions = {
+			// TODO: credentials can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
+			credentials: this._input.credentials || 'same-origin',
+			...fetchOptions,
+			method: normalizeRequestMethod(fetchOptions.method || this._input.method)
 		};
 
-		if (input instanceof globals.Request) {
-			this._input = input;
-
-			// `ky` options have precedence over `Request` options
-			this._options = {
-				...this._options,
-				method: otherOptions.method || input.method,
-				headers: otherOptions.headers || input.headers,
-				body: otherOptions.body || input.body,
-				credentials: otherOptions.credentials || input.credentials
-			};
-		} else if (!(input instanceof URL) && typeof input !== 'string') {
+		if (typeof input !== 'string' && !(input instanceof URL || input instanceof globals.Request)) {
 			throw new TypeError('`input` must be a string, URL, or Request');
-		} else {
-			this._input = String(input || '');
-			this._options.prefixUrl = String(this._options.prefixUrl || '');
+		}
 
-			if (this._options.prefixUrl && this._input.startsWith('/')) {
+		if (this._options.prefixUrl && typeof this._input === 'string') {
+			if (this._input.startsWith('/')) {
 				throw new Error('`input` must not begin with a slash when using `prefixUrl`');
 			}
 
-			if (this._options.prefixUrl && !this._options.prefixUrl.endsWith('/')) {
+			if (!this._options.prefixUrl.endsWith('/')) {
 				this._options.prefixUrl += '/';
 			}
 
 			this._input = this._options.prefixUrl + this._input;
-
-			if (searchParams) {
-				const url = new URL(this._input, globals.document && globals.document.baseURI);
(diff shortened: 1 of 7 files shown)


sholladay:
Maybe I just need some sleep and am missing something obvious, but it's not clear to me why that timeout test is failing. I did not intentionally change any timeout or retry behavior.

Any thoughts @szmarczak?

index.js
```diff
@@ -214,110 +214,102 @@ const normalizeRetryOptions = retry => {
...
+			onDownloadProgress,
+			prefixUrl: String(prefixUrl || ''),
 			retry: normalizeRetryOptions(retry),
-			...otherOptions
+			searchParams,
+			throwHttpErrors,
+			timeout
+		};
+		this._fetchOptions = {
+			// TODO: credentials can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
```
sholladay: I tested Chrome, Firefox, Safari, and Edge 18, and all of them default to `same-origin` now. Not sure about mobile browsers, though. Should we remove this?

sholladay:
I kinda want to change the signature of hooks.

Prior to this PR, we had:
 - **beforeRequest(input, options)**
 - **beforeRetry(input, options, error, retryCount)**
 - **afterResponse(input, options, response)**

This PR changes `input` to `request` in these signatures, anyway. So it is worth considering any other breaking changes we want to do. For example...
 - We could attach the `response` to the `request` passed to `afterResponse`, rather than having it as a separate argument. I find this more elegant since the response _belongs_ to a particular request. Plus it improves the symmetry of the call signature between `beforeRequest` and `afterResponse`.
 - We could attach `retryCount` to the `request` or `error` passed to `beforeRetry`, rather than passing it as a separate argument. I don't have strong feelings about this one, but I do feel that any function with four arguments is a bit much.

sholladay:
Ready for review. Could use help on that one failing test.

// @sindresorhus @szmarczak

sholladay:
For now, I've worked around the timeout test issue by changing it to validate synchronously.

I would argue that:
1. This is how input validation should work, anyway.
2. Even if you disagree with `1`, it still makes our validation more consistent.

sindresorhus:
Promise-returning methods should *always* reject: https://twitter.com/jaffathecake/status/638659818996269056?s=21

sholladay:
Alright, I made it a rejection. It's still validated in the constructor to work around the broken test that failed for reasons I don't understand.

sholladay:
It was a real pain to make it reject asynchronously before the request, because the rejected promise didn't have the `Body` methods, so `.json()` didn't exist and would immediately throw an error, for example. I had to place the rejection in a very specific, non-obvious place.

I wholeheartedly disagree with this design pattern of rejecting _user mistakes_ asynchronously. But I aim to please, so I made it work. The tests are passing. Please review.

At this point, the main thing I'm wondering is what options the hooks should receive. I'm thinking either:
A. Only fetch options
B. All options mixed together

The benefit of `A` is that we can more easily support users directly mutating the options, without forcing them to return a `Request`. It also avoids the illusion that you can override Ky's settings within hooks and prevents the temptation to try to do so.

As for `B`, I'm not sure if there is a concrete use case for reading Ky's options within hooks, but I can sort of imagine it being useful to do something special when `throwHttpErrors` is `false` or when `timeout` has a large value, etc. The only downside is that to support hooks mutating the object, we'd have to go back to a single combined options object (i.e. not separate fetch options vs Ky options), which I'd be fine with, albeit it feels a little less clean.

sholladay:
Anything else needed here?

In addition to the issues that this PR already addresses, I also have fixes for #167, #182, and #187, which I want to implement on top of this refactoring. I could easily include those in this PR (most of them actually remove lines of code), however I don't want to do too many unrelated things and make it harder to review. If we can merge this soon, I will get to those quickly.

It would be preferable to hold off on a release (even though everything here is working as-is). I realized that hooks really should be receiving all options for both ky and fetch mixed together, so that the hooks themselves can call ky and "inherit" the ky options. I'd like to address that at the same time as #167, which as mentioned previously, should probably be done in a separate PR. To avoid doing any unnecessary breaking changes, it would be good to group these together for the next release.

index.js
sindresorhus: Nah, not worth the risk. It's only a few bytes anyway.

sindresorhus:
This is great. Nice work 👌 And sorry for the delay. I reviewed this a week ago and was going to merge and then got distracted and forgot about it...

sholladay:
@sindresorhus is codecov broken? I just noticed that the status check for this PR says "No report found to compare against" and clicking on the details also doesn't bring up anything useful. Maybe it needs to be re-authorized?

sindresorhus:
Ah yeah. It was not enabled on the GitHub side. Done now.

PR #181, opened 2019-10-02
Update typings for BeforeRequestHook

deini:
`BeforeRequestHook` types got out of date.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -11,7 +11,7 @@ export type Input = Request | URL | string;
 export type BeforeRequestHook = (
 	input: Input,
 	options: NormalizedOptions,
-) => Response | void | Promise<void>;
+) => Response | void | Promise<Response | void>;
 
 export type BeforeRetryHook = (
 	input: Input,
--- a/index.d.ts
+++ b/index.d.ts
@@ -11,7 +11,7 @@ export type Input = Request | URL | string;
 export type BeforeRequestHook = (
 	input: Input,
 	options: NormalizedOptions,
-) => void | Promise<void>;
+) => Response | void | Promise<void>;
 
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -11,7 +11,7 @@ export type Input = Request | URL | string;
 export type BeforeRequestHook = (
 	input: Input,
 	options: NormalizedOptions,
-) => void | Promise<void>;
+) => Response | void | Promise<void>;
```
sholladay: We also need `Promise<Response>`, if I'm not mistaken. Would be good to add a test case for it in the type tests, too.

deini: Good catch, updating.

index.test-d.ts
```diff
@@ -37,8 +37,8 @@ ky(url, {
 			(_input, _options) => {
 				return new Response('Test');
 			},
-			(_input, _options) => {
-				return Promise.resolve(new Response('Test'));
+			async (_input, _options) => {
```
deini: 💯

sholladay:
Huh yet another strange error in CI that only occurs on Node 12 and isn't code related...

```js
Error {
  message: 'Protocol error (Target.setDiscoverTargets): Target closed.',
}

node_modules/puppeteer/lib/Connection.js:74:56
Connection.send (node_modules/puppeteer/lib/Connection.js:73:12)
Function.create (node_modules/puppeteer/lib/Browser.js:34:22)
Launcher.launch (node_modules/puppeteer/lib/Launcher.js:184:37)
withPage (test/helpers/with-page.js:4:18)
```

sholladay:
Got CI to pass by re-running it twice.

Thanks for the PR, @deini!

PR #192, opened 2019-11-05
Add package.json funding property

ruyadorno:
**npm** just added a new [`npm fund` command](https://blog.npmjs.org/post/188841555980/updates-to-community-docs-more) in it's [v6.13.0](https://github.com/npm/cli/releases/tag/v6.13.0) release as part of the efforts to help out the OSS community.

This PR adds `funding` info to **ky** so that users depending on it can actually retrieve the funding url when they run `npm fund` in their projects.

Note: The funding information will only be available once a new version of the package is published to the npm registry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
 		"email": "sindresorhus@gmail.com",
 		"url": "sindresorhus.com"
 	},
+	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
 	"engines": {
 		"node": ">=8"
 	},
@@ -23,10 +24,6 @@
 		"umd.js",
 		"umd.d.ts"
 	],
-	"funding": {
-		"type": "GitHub",
-		"url": "https://github.com/sponsors/sindresorhus"
-	},
 	"keywords": [
 		"fetch",
 		"request",
--- a/package.json
+++ b/package.json
@@ -4,12 +4,12 @@
 	"description": "Tiny and elegant HTTP client based on the browser Fetch API",
 	"license": "MIT",
 	"repository": "sindresorhus/ky",
+	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
 	"author": {
 		"name": "Sindre Sorhus",
 		"email": "sindresorhus@gmail.com",
 		"url": "sindresorhus.com"
 	},
-	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
 	"engines": {
 		"node": ">=8"
 	},
--- a/package.json
+++ b/package.json
@@ -23,6 +23,10 @@
 		"umd.js",
 		"umd.d.ts"
 	],
+	"funding": {
+		"type": "GitHub",
+		"url": "https://github.com/sponsors/sindresorhus"
+	},
 	"keywords": [
 		"fetch",
 		"request",


sindresorhus:
This is great, but why can’t npm just use `funding.yml` in the repo? It’s annoying having to duplicate the information. I have 1000+ packages to update now...

ruyadorno:
hi @sindresorhus first of all I'm really sorry this change ends up putting even more pressure on maintainers 😞 it was totally not my intention with it. My goal with this PR was just to raise awareness on the new feature, rest assure I understand your point of view if in case you don't want to go this route for now.

That said, to give you the context I have, I don't think supporting `.github/funding.yml` itself was ever on our radar (you can see the [approved RFC](https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md) and the [discussions prior to it](https://github.com/npm/rfcs/pull/54)) as we have been working closely with the [Node.js foundation package-maintenance working group](https://github.com/nodejs/package-maintenance) on their [package support proposal](https://github.com/nodejs/package-maintenance/issues/241) and the entire `funding` initiative evolved from it as an actionable item we could provide to the community sooner rather than later.

Thanks for your feedback, I really value your input on this and would love to discuss more 😊

ruyadorno:
🤔 now that you mentioned it, technically speaking it seems to me that even if we were to support `.github/funding.yml` itself would also require extra work (in the form of un-ignoring that file + making extra package releases) from open source maintainers since that file is ignored from most packages today.

sholladay:
My expectation would be that npm uses funding.yml as a fallback if this `funding` property in package.json is not used.

This is sort of similar to how npm respects `.gitignore` to determine its own behavior, even though `.gitignore` should not be deployed in the package on the registry. But you can use the `files` property in package.json to specify a whitelist of files instead - both are supported. npm doesn't try to pretend like git doesn't exist, rather it integrates with git smoothly, yet provides an alternative mechanism for those who choose to use it or who are not using git.

sindresorhus:
> now that you mentioned it, technically speaking it seems to me that even if we were to support .github/funding.yml itself would also require extra work (in the form of un-ignoring that file + making extra package releases) from open source maintainers since that file is ignored from most packages today.

I meant that npm already knows my package is on GitHub as I've specified it in the `repo` field, so the *npm server* could look for `funding.yml` in the specified repo. It doesn't really make sense to include the file in the package itself.

sindresorhus:
@ruyadorno I tried it out on a package of mine: https://github.com/sindresorhus/is-installed-globally/blob/f36107f2b4dd90705ec46759115ae8ae3fd40495/package.json#L12 The [RFC says](https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md#implementation) the `funding` property can just be a string, but that seems to not be working. When I change it to use an object, it works. I'm trying to have the least amount of moot data in my package.json and I feel the `type` can be inferred from the URL, hence why I want to specify `funding` as a string.

ruyadorno:
@sindresorhus thank you so much for bringing that up! it does sound like an oversight indeed! (I'll bring that up to the team) for now what works is just ignoring `type`, like this: https://github.com/tapjs/node-tap/blob/master/package.json#L105

sindresorhus:
I had another idea. Would have been nice if we at least could have (no URL):

```json
{
	"funding": {
		"type": "github"
	}
}
```

Which would open https://github.com/sindresorhus/ky?sponsor=1 (try clicking it) for the repo of the package. This is usually desirable for projects with multiple maintainers as it will display the modal as if you had clicked the "Sponsor" repo button, and will show everything from `funding.yml`, including multiple GitHub Sponsors profiles. That would elegantly solve the problem of duplicate metadata.

package.json
```diff
@@ -9,6 +9,7 @@
 		"email": "sindresorhus@gmail.com",
 		"url": "sindresorhus.com"
 	},
+	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
```
ruyadorno: reminder, this won't work at the moment ☺️

sindresorhus: Yes, I'll merge this PR when it's fixed. I don't want to keep changing the format, so I'd rather wait.

sindresorhus:
Merging as support for shorthand is being added in https://github.com/npm/cli/pull/472

sindresorhus:
@ruyadorno I would recommend npm documents the trick of using `https://github.com/<user>/<repo>?sponsor=1"` as URL for projects with multiple GitHub Sponsors profiles.

ruyadorno:
support to string just landed in [v6.13.1](https://github.com/npm/cli/releases/tag/v6.13.1)

> @ruyadorno I would recommend npm documents the trick of using https://github.com/<user>/<repo>?sponsor=1" as URL for projects with multiple GitHub Sponsors profiles.

thanks for the heads up, I wasn't aware of it 👍 I'll see with the rest of the team if we have a place where that could fit in the docs but otherwise I'll make sure to share in blogs, etc 😊

PR #196, opened 2019-11-08
Avoid body method parse errors on 204 responses

### LINKED ISSUE #193: Don't throw on `204 No Content` when parsing response
Hello, I ran into a little problem. Not sure if this is a bug in the library or the expected behavior.
I have a little wrapper over the library:
```js
const api = ky.create({
	prefixUrl: '...' // server api url
});

function request({path, ...params}) {
	return api(path, params)
		.json()
		.catch((err) => {
			if (err.response) {
				return err.response.json().then((err) => {
					throw err;
				});
			} else {
				throw err;
			}
		});
}
```
It works great for get, post and patch requests, since they return json. However, it broke for delete request, which returned `204 No Content`, with error `SyntaxError: Unexpected end of JSON input`.
Should the ky handle such situations and return `null`, for example? Or should I take care of this problem myself?
My current working solution:
```js
function request({path, ...params}) {
	return api(path, params)
		.text()
		.then((text) => text && JSON.parse(text))
		.catch((err) => {
			if (err.response) {
				return err.response.json().then((err) => {
					throw err;
				});
			} else {
				throw err;
			}
		});
}
```

sholladay:
Closes #193

This PR short-circuits the `.json()` body method and makes it return an empty string instead of parsing the body when the response status is 204 No Content. That way we avoid throwing a parse error due to an empty body.

Note that, as of this writing, we make no attempt to read the response body to check if it is empty, rather we trust the response status code from the server.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -291,7 +291,7 @@ interface NormalizedOptions extends RequestInit {
 }
 
 /**
-Returns a `Response` object with `Body` methods added for convenience.
+Returns a `Response` object with `Body` methods added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, an appropriate `Accept` header will be set depending on the body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range of `200...299` and they will return `undefined` if the response status is `204` instead of throwing a parse error due to an empty body.
 */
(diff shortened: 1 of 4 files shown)


sindresorhus:
Can you update index.d.ts too?

sholladay:
@sindresorhus done

sholladay:
Should be ready to merge.

Since we decided to go with empty string, the TypeScript types don't need to be updated.

PR #197, opened 2019-11-08
Add ability to remove headers with `ky.extend()`

### LINKED ISSUE #189: Remove headers
How can i remove header by extends another ky instance?

Something like this:
```javascript
const http = ky.create({
  headers: {
    authorization: 'Bearer TOKEN',
    otherHeader: 'value',
  }
});

const newHttp = http.extend({
  headers: {
    authorization: undefined // <- remove  Authorization header
  }
});
```

Ofc. this not work, this is only a showcase.

luq10:
Closes #189

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 yarn.lock
 .nyc_output
 umd.js
+.idea
--- a/index.d.ts
+++ b/index.d.ts
@@ -153,7 +153,7 @@ export interface RetryOptions {
 /**
 Options are the same as `window.fetch`, with some exceptions.
 */
-export interface Options extends RequestInit {
+export interface Options extends Omit<RequestInit, 'headers'> {
 	/**
 	HTTP method used to make the request.
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -162,6 +162,11 @@ export interface Options extends RequestInit {
 	method?: LiteralUnion<'get' | 'post' | 'put' | 'delete' | 'patch' | 'head', string>;
 
 	/**
+	Override default Headers
+	 */
+	headers?: HeadersInit | { [key: string]: undefined }
+
+	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 
 	Accepts a plain object which will be `JSON.stringify()`'d and the correct header will be set for you.
--- a/index.d.ts
+++ b/index.d.ts
@@ -165,17 +165,27 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 	Override default Headers
 
 	Extra types:
-	-  `{ [key: string]: undefined }` for add ability to remove headers with extend()
+	- `{[key: string]: undefined}` for add ability to remove headers with extend()
 
-	 	@example
-	 	```
-	 	import ky from 'ky';
+	@example
+	```
+	import ky from 'ky';
 
-	 	const original = ky.create({headers: { rainbow: 'rainbow', unicorn: 'unicorn' }});
-	 	const extended = original.extend({headers: { rainbow: undefined }});
-	 	```
-	 */
-	headers?: HeadersInit | { [key: string]: undefined }
+	const original = ky.create({
+		headers: {
+			rainbow: 'rainbow',
+			unicorn: 'unicorn'
+		}
+	});
+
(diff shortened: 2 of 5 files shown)


sindresorhus:
Thanks 🙏 This needs to be documented in the readme and index.d.ts

luq10:
I add types for TS

sindresorhus:
You need to document with English prose that specifying `undefined` as a header value clears it (both in readme and index.d.ts).

test/main.js
```diff
@@ -164,6 +164,33 @@ test('custom headers', async t => {
...
+	server.get('/', (request, response) => {
+		response.send(request.headers);
+		response.end();
+	});
+
+	const original = ky.create({
+		headers: {
+			rainbow: 'rainbow',
+			unicorn: 'unicorn'
+		}
```
sholladay: Can you also test what happens when the `original` instance was created with `new Headers()` instead of a plain object? I'm not certain it will work correctly.

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 yarn.lock
 .nyc_output
 umd.js
+.idea
```
sholladay: You should add this to a global gitignore instead, as most of us don't use WebStorm.

 - http://egorsmirnov.me/2015/05/04/global-gitignore-file.html
 - https://thoughtbot.com/blog/global-gitignore

luq10:
> You need to document with English prose that specifying `undefined` as a header value clears it (both in readme and index.d.ts).

Done.

test/main.js
luq10: I added this test, and as you said: it faild. 
What we should get it this case? I`m no sure that we want to mix plain object and `Headers` instance.

```
const original = ky.create({
	headers: new Headers({
		rainbow: 'rainbow',
		unicorn: 'unicorn'
	})
});
const extended = original.extend({
	headers: {
		rainbow: undefined
	}
});
```

sholladay: Well people use `new Headers()`. We have explicit support for it in our API. So I'd at least prefer to it to not be undefined behavior. What should Ky do in that case? Whatever it is, it should have a test.

luq10: Ok, so I updated expected values in this test

index.d.ts
```diff
@@ -153,14 +153,30 @@ export interface RetryOptions {
...
+export interface Options extends Omit<RequestInit, 'headers'> {
 	/**
 	HTTP method used to make the request.
 
 	Internally, the standard methods (`GET`, `POST`, `PUT`, `PATCH`, `HEAD` and `DELETE`) are uppercased in order to avoid server errors due to case sensitivity.
 	*/
 	method?: LiteralUnion<'get' | 'post' | 'put' | 'delete' | 'patch' | 'head', string>;
 
+	/**
+	Override default Headers
```
sindresorhus: This needs to be better written. Same with the below.

index.d.ts
```diff
@@ -153,14 +153,40 @@ export interface RetryOptions {
...
+			rainbow: 'rainbow',
+			unicorn: 'unicorn'
+		}
+	});
+
+	const extended = original.extend({
+		headers: {
+			rainbow: undefined
+		}
+	});
```
sindresorhus: You need to show the resulting headers as a comment to make it clear what's happening.

test/main.js
```diff
@@ -164,6 +164,60 @@ test('custom headers', async t => {
...
+	});
+	const extended = original.extend({
+		headers: {
+			rainbow: undefined
+		}
+	});
+
+	const data = await extended(server.url).json();
+
+	t.false('unicorn' in data);
```
sholladay: The test should be asserting expected behavior. As-is, it is a bit confusing.

I would prefer to use `t.true()` here and then mark the test as failing with `test.failling(...)`

index.d.ts
sindresorhus: I think I was clear what needs to change here. And please don't copy-paste a generic comment.

index.d.ts
sindresorhus: At least give it a try first.

sindresorhus:
@luq10 Bump :)

sindresorhus:
Closing for lack of response.

Issue #185, opened 2019-10-14
Support retries on certain network failures


spruce-bruce:
In poor network conditions network requests may fail in weird ways. Ky supports retrying on predictable errors (ie, based on response codes), could you support retrying on predictable network failures?

I've recently done a lot of work cataloguing the types of network failures I've seen in my production native apps using fetch, and there are two main failures that could be incorporated into your retry logic:

1. fetch() throws "Network request failed" (occurs when a request is dropped, but also when a domain is unreachable like google.comm)
2. response is returned without a body (stream interrupted during response i think)

It will be difficult to support these without muddying your api, though, because it is not obvious that these should always be retried. You'd have to allow users to provide some kind of a callback that allows them to make a decision based on the request.

sholladay:
The second one sounds problematic to implement. But the first one sounds relatively simple. Can you explain more why you think a callback would be necessary? I'm not sure I follow what a Ky user would do to filter which network errors result in a retry.

sholladay:
The server is in a much better position to know whether an incoming request should be accepted than the client is to know whether an outgoing retry request should be performed. Any solution within or on top of Ky is inherently going to be imperfect when an explicit status code is missing.

I think the best we can do here is give `beforeRetry` hooks the ability to prevent a retry from happening (perhaps by returning a symbol). Then you could, for example, call your API to see if your `POST` request did actually make it through.

spruce-bruce:
Yes I agree - I think we're talking about the same thing - the `beforeRetry` hook would be the synchronous callback that I'm referring to. Hook is a much clearer word for it, of course.
 
Ky can't possibly be responsible for knowing if it should be retrying in these scenarios so you _must_ give the user control of that decision.

sholladay:
@spruce-bruce do you feel that we can close this issue, given that PR https://github.com/sindresorhus/ky/pull/198 was released?

We haven't made any changes regarding which errors are automatically retried by Ky, but you can now `return ky.stop` in the `beforeRetry` hook to prevent a retry. So I think your use case is fully supported because you can filter out certain responses. Is that enough for you? Do you think there's any further improvements we could make?

spruce-bruce:
@sholladay awesome! I just took a look, here's my understanding of how that PR will effect the two "retryable" failure cases that I'm thinking about:

- Fetch throws an error when the request is dropped
[...]

PR #198, opened 2019-11-08
Add ability to stop retries in `beforeRetry` hooks

whitecrownclown:
Hello,

This is related to the discussion from #185.

Summary of the PR:

- exported a new Symbol `NO_RETRY_SYMBOL`
- the Symbol can be imported & returned from a `beforeRetry` hook in order to abandon retries
- I'm unsure if an error should be thrown even if you have `throwHttpErrors` enabled
- added test & tsd

@sholladay @sindresorhus 
Thoughts?

Thank you!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -342,7 +342,7 @@ export class TimeoutError extends Error {
 /**
 Symbol for aborting retries in the `beforeRetry` hook
 */
-export const NO_RETRY_SYMBOL: Symbol;
+export const ABANDON_RETRY: Symbol;
 
 declare const ky: {
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -339,6 +339,11 @@ export class TimeoutError extends Error {
 	constructor();
 }
 
+/**
+Symbol for aborting retries in the `beforeRetry` hook
+*/
+export const NO_RETRY_SYMBOL: Symbol;
+
 declare const ky: {
 	/**
 	Fetch the given `url`.
--- a/index.d.ts
+++ b/index.d.ts
@@ -340,9 +340,9 @@ export class TimeoutError extends Error {
 }
 
 /**
-Symbol for aborting retries in the `beforeRetry` hook
+Symbol for aborting retries in the `beforeRetry` hook.
 */
-export const ABANDON_RETRY: Symbol;
+declare const stop: unique symbol;
 
 declare const ky: {
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -339,32 +339,6 @@ export class TimeoutError extends Error {
 	constructor();
 }
 
-/**
-A `Symbol` that can be returned by a `beforeRetry` hook to stop the retry.
-This will also short circuit the remaining `beforeRetry` hooks.
-
-@example
-```
-import ky, {stop as kyStop} from 'ky';
-
-(async () => {
-	await ky('https://example.com', {
-		hooks: {
-			beforeRetry: [
-				async (request, options, errors, retryCount) => {
-					const shouldStopRetry = await ky('https://example.com/api');
-					if (shouldStopRetry) {
-						return kyStop;
-					}
-				}
-			]
-		}
-	});
-})();
-```
-*/
-export const stop: unique symbol;
-
 declare const ky: {
 	/**
 	Fetch the given `url`.
@@ -449,6 +423,32 @@ declare const ky: {
 	@returns A new Ky instance.
 	*/
 	extend(defaultOptions: Options): typeof ky;
+
+	/**
+	A `Symbol` that can be returned by a `beforeRetry` hook to stop the retry.
+	This will also short circuit the remaining `beforeRetry` hooks.
+
+	@example
+	```
+	import ky from 'ky';
+
+	(async () => {
+		await ky('https://example.com', {
+			hooks: {
+				beforeRetry: [
+					async (request, options, errors, retryCount) => {
+						const shouldStopRetry = await ky('https://example.com/api');
+						if (shouldStopRetry) {
+							return ky.stop;
+						}
+					}
+				]
+			}
+		});
+	})();
+	```
+	*/
+	readonly stop: unique symbol;
 };
 
 export default ky;
--- a/index.d.ts
+++ b/index.d.ts
@@ -342,7 +342,7 @@ export class TimeoutError extends Error {
 /**
 Symbol for aborting retries in the `beforeRetry` hook.
 */
-declare const stop: unique symbol;
+export const stop: unique symbol;
 
 declare const ky: {
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -340,7 +340,28 @@ export class TimeoutError extends Error {
 }
 
 /**
-Symbol for aborting retries in the `beforeRetry` hook.
+A `Symbol` that can be returned by a `beforeRetry` hook to stop the retry.
+This will also short circuit the remaining `beforeRetry` hooks.
+
+@example
+```
+import ky, {stop as kyStop} from 'ky';
+
+(async () => {
+	await ky('https://example.com', {
+		hooks: {
+			beforeRetry: [
+				async (request, options, errors, retryCount) => {
+					const shouldStopRetry = await ky('https://example.com/api');
+					if (shouldStopRetry) {
+						return kyStop;
+					}
+				}
+			]
+		}
+	});
+})();
+```
 */
 export const stop: unique symbol;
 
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -368,16 +371,24 @@ class Ky {
...
+				const beforeRetryResults = new Set();
+
 				for (const hook of this._options.hooks.beforeRetry) {
-					// eslint-disable-next-line no-await-in-loop
-					await hook(
-						this.request,
-						this._options,
-						error,
-						this._retryCount,
+					beforeRetryResults.add(
```
sholladay: It would be more consistent with the behavior of our other hooks if returning the symbol short-circuited the loop. Meaning, return immediately inside of the loop, preventing any further `beforeRetry` hooks from running.

index.js
```diff
@@ -474,5 +485,6 @@ export default createInstance();
 
 export {
 	HTTPError,
-	TimeoutError
+	TimeoutError,
+	NO_RETRY_SYMBOL
```
sholladay: I think it should be named `NO_RETRY` or `noRetry` or `cancel` or `abandon`, etc. Probably shouldn't include "symbol" in its name.

index.js
whitecrownclown: I wasn't sure about that, my reasoning was that you'd want to have all your hooks running before making the "cancel retry" decision.

index.js
whitecrownclown: I agree, `ABANDON_RETRY` sounds accurate. What do you think?

index.js
sholladay: I get what you are saying. I do see how this case is a bit different, as the symbol is not a true "value" being returned by the hook, unlike the other cases where a hook may return a `Request` or `Response` instance.

Still, I think it would be surprising for `beforeRetry` to behave differently in this respect than the other hook types. IMO, we need to have a holistic approach on whether returning a value from a hook short-circuits the chain. And we should apply that logic consistently.

index.js
sholladay: I'm wondering if a more generic name would be preferable. Maybe someone will want to use this symbol in a `beforeRequest` hook to ignore a request. We can always rename it later, though...

sindresorhus: I don't think it should be uppercase. I like `stopRetry`, or just `stop` if we can to make it generic.

index.d.ts
```diff
@@ -339,6 +339,11 @@ export class TimeoutError extends Error {
 	constructor();
 }
 
+/**
+Symbol for aborting retries in the `beforeRetry` hook
+*/
+export const ABANDON_RETRY: Symbol;
```
sindresorhus: Should be `unique symbol`, see https://github.com/sindresorhus/find-up/blob/6c32f0caed1684ef778053b6f79b13a772e22ba4/index.d.ts#L3 and https://github.com/sindresorhus/find-up/blob/6c32f0caed1684ef778053b6f79b13a772e22ba4/index.d.ts#L134

index.js
```diff
@@ -131,6 +132,8 @@ const retryAfterStatusCodes = [
 	503
 ];
 
+const ABANDON_RETRY = globals.Symbol('abandon-retry');
```
sindresorhus: Don't use `globals` here. Symbol is an expected JS feature.

test/hooks.js
```diff
@@ -378,6 +378,38 @@ test('beforeRetry hook is called with error and retryCount', async t => {
...
+					t.truthy(error);
+					t.true(retryCount === 1);
+
+					return ABANDON_RETRY;
+				}
+			]
+		}
+	});
+
+	t.true(requestCount === 1);
```
sindresorhus: Use `t.is`

index.d.ts
```diff
@@ -339,6 +339,11 @@ export class TimeoutError extends Error {
 	constructor();
 }
 
+/**
+Symbol for aborting retries in the `beforeRetry` hook
```
sindresorhus:
```suggestion
Symbol for aborting retries in the `beforeRetry` hook.
```

Would be useful to describe how it's supposed to be used and an example.

sindresorhus:
It also needs to be documented in the readme.

index.d.ts
whitecrownclown: Yes, i was going to document its usage as well. Thanks.

index.js
sholladay: Agreed about camel case. I like `abandon` or `abort` the best, but I'm fine with `stop`, too.

readme.md
```diff
@@ -385,6 +385,26 @@ Exposed for `instanceof` checks. The error has a `response` property with the [`
...
+			beforeRetry: [
+				async (request, options, errors, retryCount) => {
+					const stopRetry = await ky('https://example.com/api');
+					if (stopRetry) return stop;
+				}
+			]
+		}
+	});
+})();
+```
```
sindresorhus: This should be in index.d.ts too. The readme and index.d.ts should be in sync.

sindresorhus:
@sholladay Looks good to you?

index.d.ts
```diff
@@ -339,6 +339,32 @@ export class TimeoutError extends Error {
 	constructor();
 }
 
+/**
+A `Symbol` that can be returned by a `beforeRetry` hook to stop the retry.
+This will also short circuit the remaining `beforeRetry` hooks.
+
+@example
+```
+import ky, {stop as kyStop} from 'ky';
```
szmarczak: I'd name it `RETRY_STOP`.

index.d.ts
```diff
@@ -339,6 +339,32 @@ export class TimeoutError extends Error {
...
+
+(async () => {
+	await ky('https://example.com', {
+		hooks: {
+			beforeRetry: [
+				async (request, options, errors, retryCount) => {
+					const shouldStopRetry = await ky('https://example.com/api');
+					if (shouldStopRetry) {
+						return kyStop;
```
szmarczak: This mechanism could be also applied to Got :)

index.d.ts
sindresorhus: We went with something generic so we can use the symbol for other places we want to support stopping.

index.d.ts
sindresorhus: Yup. Let’s try it out here first and see how it works out.

index.d.ts
szmarczak: Okay, then it should be `SYMBOL_STOP` or `symbol_stop` IMO :P

szmarczak: Stop is too general I think

sindresorhus: I don't really like hungarian notation. Prefixing it with "symbol" doesn't make the intent any clearer.

sindresorhus:
Any thoughts on making it a property on the default export instead of a named export? It doesn't really make sense on its own and `ky.stop` would make the intend clearer, instead of having to rename the named export to `kyStop`.

index.test-d.ts
```diff
@@ -25,6 +25,7 @@ expectType<typeof ky>(ky.create({}));
 expectType<typeof ky>(ky.extend({}));
 expectType<HTTPError>(new HTTPError(new Response));
 expectType<TimeoutError>(new TimeoutError);
+expectType<symbol>(stop);
```
sholladay: I'm not sure if it matters, but shouldn't this be `Symbol`?

sindresorhus: No, it’s a primitive, so should be lowercased (both work, at least until I get XO working with ambient TS files).

index.d.ts
sholladay: @szmarczak thankfully that will generate a `TypeError`. But I get your point. I have the same icky feelings about objects with plural names. I tend to reserve plural names for arrays and it can be confusing if there is no distinction in some APIs. Symbols add to the problem because they are often used to signal intent for an action, similar to a function, so the naming styles conflict. TypeScript and IntelliSense make this a bit more obvious, though.

whitecrownclown:
> Any thoughts on making it a property on the default export instead of a named export? It doesn't really make sense on its own and `ky.stop` would make the intend clearer, instead of having to rename the named export to `kyStop`.

I went ahead and made this change. I think it makes sense.

sholladay:
> Any thoughts on making it a property on the default export instead of a named export?

@sindresorhus I also prefer `ky.stop` instead of named export. 👍 

What do you think about doing this for the other exports as well? I believe it would solve that Rollup warning during the build and make it easier for CommonJS users to import the package, as they should then be able to use `ky` directly instead of `ky.default`.

sindresorhus:
> What do you think about doing this for the other exports as well? I believe it would solve that Rollup warning during the build and make it easier for CommonJS users to import the package, as they should then be able to use ky directly instead of ky.default.

👍

PR #200, opened 2019-11-08
Allow overriding the default accept headers

### LINKED ISSUE #199: Can't override Request 'Accept' header
```js
ky.post(`/foo/bar`, {
  headers: {
    accept: 'application/vnd.VENDER_NAME+json; version=2',
  },
  json,
});
```

Ky ignores the supplied `accept` header and instead decides to use `application/json` (probably due to setting the `json` property).

sholladay:
Fixes #199 

This PR makes it possible to override the HTTP `accept` header that Ky sets by default when a [body method](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) shortcut is used, such as `.json()`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -315,7 +315,7 @@ class Ky {
 
 		for (const [type, mimeType] of Object.entries(responseTypes)) {
 			result[type] = async () => {
-				this.request.headers.set('accept', mimeType);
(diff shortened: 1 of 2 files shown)


jacob-fueled:
So quick! Thank you 🙌

PR #201, opened 2019-11-10
Add `response` to `beforeRetry` hook and change its parameter to a single object

### LINKED ISSUE #194: beforeRetry hook should expose failed response
Hi

I was using `beforeRetry` to handle token refreshing - which I eventually moved to `afterResponse`.

However I think that hooks, being a kind of "setup" api (lets say I set them once to be used in the background, so I dont have to handle them when I do requests) should expose raw data instead of parsed one.

So, this interface

```ts
export type BeforeRetryHook = (
	input: Input,
	options: NormalizedOptions,
	error: Error,
	retryCount: number,
) => void | Promise<void>;
```

I would change to 

```ts
export type BeforeRetryHook = (
	input: Input,
	options: NormalizedOptions,
	response: Response,
	retryCount: number,
) => void | Promise<void>;
```

or to avoid breaking changes

```ts
export type BeforeRetryHook = (
	input: Input,
	options: NormalizedOptions,
	error: Error,
	retryCount: number,
    response: Response
) => void | Promise<void>;
```

This allows me to control if and on what conditions request should be called again, for example my (real life) scenario.

> I want  to refresh tokens in retry hook. I want them to refresh `only` if I have status 403 and backend error code (in body) is token_expired. Otherwise I dont want tokens to be refreshed.

So having access to response to failed request will be very helpful. I guess the philosophy is the same (I have errors) but they should not be parsed.

What do you think guys?

lkostrowski:
Fixes #194

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,12 +13,13 @@ export type BeforeRequestHook = (
 	options: NormalizedOptions,
 ) => Request | Response | void | Promise<Request | Response | void>;
 
-export type BeforeRetryHook = (
+export type BeforeRetryHook = (options: {
 	request: Request,
 	options: NormalizedOptions,
 	error: Error,
+	response: Response;
 	retryCount: number,
-) => void | Promise<void>;
+}) => void | Promise<void>;
 
 export type AfterResponseHook = (
 	request: Request,
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,9 +15,9 @@ export type BeforeRequestHook = (
 
 export type BeforeRetryHook = (options: {
 	request: Request,
+	response: Response,
 	options: NormalizedOptions,
 	error: Error,
-	response: Response;
 	retryCount: number,
 }) => void | Promise<void>;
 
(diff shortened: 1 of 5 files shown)


sholladay:
Should we make all hooks take an object for consistency? Or only once they grow to this many arguments?

Also, what do you think of my idea to attach the response to the request as `request.response`?

lkostrowski:
1. Yes, maybe all hooks should be changed if we aim for breaking changes anyway
2. I think Response isn't exactly a part of Request. Also you mess with native interface. IMO It should be left separated

sindresorhus:
> Should we make all hooks take an object for consistency? Or only once they grow to this many arguments?

Yeah, let's change the other ones too.

sindresorhus:
You need to update the readme (https://github.com/sindresorhus/ky#hooksbeforeretry), examples, and relevant tests too.

sindresorhus:
@lkostrowski Bump :)

lkostrowski:
Sorry I was on vacation. I will try to finish it during xmas

sindresorhus:
Bump :)

lkostrowski:
Should be fine, sorry for delay

sindresorhus:
> You need to update the readme (sindresorhus/ky#hooksbeforeretry)

You need to update the docs text too, not just the examples.

sindresorhus:
Thanks :)

PR #203, opened 2019-11-12
Fix global property lookup

### LINKED ISSUE #202: Regression in global property lookup
PR #153 was intended to make it easier to mock globals that Ky relies on.

But it actually _broke_ my app's mocking...

In fact, in environments that have `globalThis` (including Node.js), the logic for determining which global object to use is completely broken. It will always choose `globalThis` no matter what. Anyone using an artificial `window` or `self` (e.g. with [jsdom](https://github.com/jsdom/jsdom)), which is necessary in React projects, is no longer able to put things there and have Ky pick up on them. They just get ignored.

This is happening because the `return` statements were removed from [here](https://github.com/sindresorhus/ky/blob/a50948adeff0c2255626ed28465e126e095a7c04/index.js#L4-L21). Instead, the `if` statements now "fall through" [here](https://github.com/sindresorhus/ky/blob/65f38f9f000413525d19d16599512b24e0a496aa/index.js#L9-L26) until it hits `globalThis`, which doesn't have the same properties as `window` in my case, yet Ky chooses to use `globalThis` anyway, even though the property doesn't exist there.

Another consequence of falling through like this, even if your environment doesn't have `globalThis`, is that the order of precedence is reversed from what it was previously. Before it would stop at the earliest match found, now it stops at the last match found.

I would prefer to revert this commit and rethink how it works, but a simple temporary fix for the first problem would be to check that the `property` exists on `globalThis` before setting the `parent` to `globalThis`.

// cc @luxferresum

sholladay:
Fixes #202 

This PR restores the global property lookup behavior that we had prior to PR #153, while still supporting the use case outlined in #151.

This fix is critical for anyone who mocks `window` in Node.js with any of our globals.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,62 +2,43 @@
 
 const globals = {};
 
-{
-	const getGlobal = property => {
-		let parent;
-
-		/* istanbul ignore next */
-		if (typeof self !== 'undefined' && self && property in self) {
-			parent = self;
-		}
-
-		/* istanbul ignore next */
-		if (typeof window !== 'undefined' && window && property in window) {
-			parent = window;
-		}
-
-		if (typeof global !== 'undefined' && global && property in global) {
-			parent = global;
-		}
-
-		/* istanbul ignore next */
-		if (typeof globalThis !== 'undefined' && globalThis) {
-			parent = globalThis;
-		}
-
-		if (typeof parent === 'undefined') {
-			return;
-		}
-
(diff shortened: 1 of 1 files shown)


sindresorhus:
Tests are failing.

sholladay:
Tests are fixed. Needed a null check for older environments like Node 10.

PR #206, opened 2019-11-16
Move named exports to properties of default export

### LINKED ISSUE #205: Move named exports to properties of default export
We want to restructure the exports of Ky to only use `export default` and no named exports. This involves attaching `HTTPError` and `TimeoutError` to the Ky instance and removing the existing export statements for them.

The reasoning behind this:
 - `ky.TimeoutError` is clearer than just `TimeoutError`
 - The current situation of mixing default and named exports hurts the ergonomics for CommonJS users since they have to use `ky.default`
 - The alternative is to _only_ use named exports, but then what would we name the current default export? `import { ky } from 'ky'` would be a bit strange and `import * as ky` would then involve using `ky.ky`, which almost seems worse than `ky.default`. We should just stick to using either named exports or the default export. In Ky's case, default export seems to make the most sense.

See discussion in https://github.com/sindresorhus/ky/pull/198#issuecomment-554024403

sholladay:
Closes #205 

This PR moves `HTTPError` and `TimeoutError` to properties of the default export rather than named exports.

CommonJS users can now just use `ky` instead of `ky.default`. Also eliminates a warning logged during the build.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -453,8 +453,8 @@ declare const ky: {
 	```
 	*/
 	readonly stop: unique symbol;
-	readonly TimeoutError: TimeoutError;
-	readonly HTTPError: HTTPError;
+	readonly TimeoutError: typeof TimeoutError;
+	readonly HTTPError: typeof HTTPError;
 };
 
 export default ky;
--- a/index.d.ts
+++ b/index.d.ts
@@ -331,7 +331,7 @@ export interface ResponsePromise extends Promise<Response> {
 /**
 The error has a response property with the `Response` object.
 */
-export class HTTPError extends Error {
+declare class HTTPError extends Error {
 	constructor(response: Response);
 	response: Response;
 }
@@ -339,7 +339,7 @@ export class HTTPError extends Error {
 /**
 The error thrown when the request times out.
 */
-export class TimeoutError extends Error {
+declare class TimeoutError extends Error {
 	constructor();
 }
 
@@ -453,6 +453,8 @@ declare const ky: {
 	```
 	*/
 	readonly stop: unique symbol;
+	readonly TimeoutError: TimeoutError;
+	readonly HTTPError: HTTPError;
 };
 
 export default ky;
--- a/index.js
+++ b/index.js
@@ -455,6 +455,8 @@ const createInstance = defaults => {
 		ky[method] = (input, options) => new Ky(input, validateAndMerge(defaults, options, {method}));
 	}
 
+	ky.HTTPError = HTTPError;
+	ky.TimeoutError = TimeoutError;
 	ky.create = newDefaults => createInstance(validateAndMerge(newDefaults));
 	ky.extend = newDefaults => createInstance(validateAndMerge(defaults, newDefaults));
 	ky.stop = stop;
@@ -463,8 +465,3 @@ const createInstance = defaults => {
 };
 
 export default createInstance();
-
-export {
-	HTTPError,
-	TimeoutError
-};
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -1,5 +1,5 @@
 import {expectType} from 'tsd';
-import ky, {HTTPError, TimeoutError, ResponsePromise, DownloadProgress, Options, NormalizedOptions, Input} from '.';
+import ky, {ResponsePromise, DownloadProgress, Options, NormalizedOptions, Input} from '.';
 
 const url = 'https://sindresorhus';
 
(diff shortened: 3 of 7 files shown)


sholladay:
@sindresorhus I couldn't figure out how to update the TypeScript types correctly. I tried making the `TimeoutError` constructor a property of `ky`, similar to `stop`, but then  it complained that `ky` is not a namespace. Could use some help there.

sindresorhus:
I think you can do:

```ts
/**
The error has a response property with the `Response` object.
*/
declare class HTTPError extends Error {
	constructor(response: Response);
	response: Response;
}

/**
The error thrown when the request times out.
*/
declare class TimeoutError extends Error {
	constructor();
}

// ...

declare const ky: {
	...
	readonly HTTPError: HTTPError;
	readonly TimeoutError: TimeoutError;
};

export default ky;

**Comment by sholladay** [comment]
I tried that, but it didn't work, unfortunately (see the [tsd errors in the CI logs](https://travis-ci.com/sindresorhus/ky/jobs/258949473#L336-L342)). I got even more errors when I tried to make `ky` a `namespace`.

**Comment by sindresorhus** [comment]
Weird. https://github.com/sindresorhus/ky/pull/206/commits/02818df636b3d89ff665e974923fa7537d41b8ec is passing for me locally.

**Comment by sholladay** [comment]
Maybe a caching issue? I don't know if `tsd` uses a cache, but I've occasionally had problems like that with XO where the cache was somehow invalid or out of date and I had to delete the cache to make it notice the new code. Not sure what causes it or how to reproduce it.

**Comment by sindresorhus** [comment]
Yeah, it was some kind of caching issue. I'm honestly not sure how to make this work.

// @joaovieira @luv-sic

**Comment by joaovieira** [comment]
My take:

As the error says when you use a type as `x.y` it looks for the `y` export in the `x` namespace. But `ky` is not a namespace. It's just an exported variable. For CJS modules, exports are usually defined as, e.g.:

```ts
declare namespace ky {
  export class HTTPError extends Error {}
  export class TimeoutError extends Error {}
};

declare function ky(): Promise;

export = ky;
```

However doing that makes this a CJS module similar to:

```ts
module.exports = {
  default: () => {},
  HTTPError: class {},
  TimeoutError: class {},
};
```

and would then require you to use `esModuleInterop` flag to import`ky` as a default export in ESM. TS would then decide whether use the namespace or the variable depending on usage.

That said, at a time everyone's switching to ESM, [including Node.js (starting from Node 13.2)](https://github.com/nodejs/node/pull/29866), I'm not sure this is the right approach. It's turning a completely fine ESM into a CJS-like module, limiting it to always have a single export. Moreover, you have the memory footprint of adding classes to every "instance" of `ky`. Arguably irrelevant, but still, the previous exports had more isolation and semantic meaning (the errors are not related to the "instances" in any way). Same could be said for the stop symbol.

Would also be interested to see the type definitions TSC would output, but there are too many things for fix to make it compile :)

sindresorhus:
@joaovieira Adding properties to an default exported function is completely fine in JavaScript (ESM), so I'm hoping there's a way to do it in TypeScript too.

joaovieira:
Yeah, it's fine, but before wasn't wrong either :)

Here's my investigation:

```ts
// index.ts
  constructor() {
    super();
    this.name = 'HTTPError';
  }
}

class Ky {
  constructor() {
  }
}

const createInstance = () => {
  const ky = () => new Ky();
  ky.create = () => createInstance();
  ky.HTTPError = HTTPError;
  return ky;
};

export default createInstance();
```

Results in:

```ts
// index.d.ts
declare class HTTPError extends Error {
  constructor();
}
declare class Ky {
  constructor();
}
declare const _default: {
  (): Ky;
  create(): any;
  HTTPError: typeof HTTPError;
};
export default _default;
```

However, the test still complains it can't find `ky` namespace:

```ts
import {expectType} from 'tsd';
import ky from '.';

expectType<ky.HTTPError>(new ky.HTTPError());
```

Both from `tsd`:
Screen Shot 2019-11-22 at 19 18 02


And with `tsc`:
Screen Shot 2019-11-22 at 19 18 12

joaovieira:
`InstanceType<typeof ky.HTTPError>` in the tests and `typeof HTTPError` in the type declaration should do it.

Though I'm still of the opinion named exports are better. TypeScript users will have to use that `InstanceType<typeof ky.HTTPError>` wherever they were previsouly just using `HTTPError` in type expressions.

sholladay:
Using `typeof` and `InstanceType` did indeed work. Thanks, @joaovieira.

@sindresorhus the tests are passing now and this is ready to merge.

sindresorhus:
Awesome. This is much nicer 👍

PR #208, opened 2019-11-22
Remove `content-type` header check for FormData body

### LINKED ISSUE #204: Error: The `content-type` header cannot be used with a FormData body. It will be set automatically.
In the browser:

```js
const req = new Request('http://google.com', { method: 'POST', body: new FormData() })
req.headers.has('content-type')
// true
```

It means we cannot pass this check even when no `content-type` header has explicitly been set.

https://github.com/sindresorhus/ky/blob/1a86d5bf16f9282e00cd35b149d2ec00826dbef7/index.js#L246-L248

So my call to ky is erroring in v0.16:

```js
    const res = await ky.post('dag/put', {
      timeout: options.timeout,
      signal: options.signal,
      headers: options.headers,
      searchParams,
      body: toFormData(serialized)
    }).json()
```

https://github.com/ipfs/js-ipfs-http-client/pull/1159

rangermeier:
Closes #204 

This PR removes the check for absence of a `content-type` header when the request body is FormData or URLSearchParams.

When there are `searchParams` the updated request will not be based on the existing Request object but from `this,_options`. Otherwise the resulting new Request would omit any attached files.

### DIFF
--- a/test/browser.js
+++ b/test/browser.js
@@ -182,17 +182,21 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
 test('FormData with searchParams', withPage, async (t, page) => {
 	const server = await createTestServer();
 	server.get('/', (request, response) => {
-		response.end('nothing');
+		response.end();
 	});
 	server.post('/', async (request, response) => {
 		const pBody = util.promisify(body);
-		response.end(await pBody(request));
+		const requestBody = await pBody(request);
+
+		t.regex(requestBody, /bubblegum pie/);
+		t.deepEqual(request.query, {foo: '1'});
+
+		response.end();
 	});
 
 	await page.goto(server.url);
 	await page.addScriptTag({path: './umd.js'});
-
-	const requestBody = await page.evaluate(url => {
+	await page.evaluate(url => {
 		window.ky = window.ky.default;
 		const formData = new window.FormData();
 		formData.append('file', new window.File(['bubblegum pie'], 'my-file'));
@@ -200,10 +204,8 @@ test('FormData with searchParams', withPage, async (t, page) => {
 			method: 'post',
 			searchParams: 'foo=1',
 			body: formData
-		}).text();
+		});
 	}, server.url);
-
-	t.regex(requestBody, /bubblegum pie/);
 
 	await server.close();
 });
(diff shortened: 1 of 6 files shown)


sindresorhus:
Can you add a test for the searchParams issue?

rangermeier:
My first try didn't capture the issue, i.e. it didn't fail when reverting 2be9a92 .

With the changes from 0d246c9 it's actually capturing the issue plus I fixed the build errors.

sholladay:
Thanks for working on this. Generally, I think your solution is pretty reasonable.

Unfortunately, I think this will introduce new bugs of its own. For example, if the `input` argument is a `Request`, its headers and other settings will be ignored. Though, that's probably preferable to the current situation.

What we _really_ want to do here is update the URL of the existing `this.request`. I didn't find a way to do it when I wrote this code, other than creating a new `Request` instance.

rangermeier:
Thanks for reviewing, @sholladay. I added a test case for the example you gave. I also changed the code for updating the URL with the `searchParams`, so there is no loss of configuration. 

```
```

test/browser.js
```diff
@@ -176,3 +178,56 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
...
+	await page.goto(server.url);
+	await page.addScriptTag({path: './umd.js'});
+
+	const requestBody = await page.evaluate(url => {
+		window.ky = window.ky.default;
+		const formData = new window.FormData();
+		formData.append('file', new window.File(['bubblegum pie'], 'my-file'));
+		return window.ky(url, {
+			method: 'post',
+			searchParams: 'foo=1',
```
sholladay: The test should assert that the search params are respected

test/browser.js
```diff
@@ -176,3 +178,56 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
...
+	await page.goto(server.url);
+	await page.addScriptTag({path: './umd.js'});
+
+	const requestBody = await page.evaluate(url => {
+		window.ky = window.ky.default;
+		const request = new window.Request(url, {
+			headers: {'content-type': 'text/css'}
+		});
+		return window.ky(request, {
+			searchParams: 'foo=1'
```
sholladay: Same here - make sure test fails if search params are missing from the request to the server.

test/browser.js
rangermeier: I added an assertion for the search params and moved the assertions into the server callback.

test/browser.js
rangermeier: done

PR #211, opened 2019-11-28
Delete `Content-Type` header for multipart requests

### LINKED ISSUE #209: Multipart boundary mismatch when using search params
Multipart boundary in `content-type` does not match boundary in body when using ky with `searchParams`. This bug was introduced in https://github.com/sindresorhus/ky/pull/208

The following example is effectively what happens here https://github.com/sindresorhus/ky/blob/d6ff489b62468b0fb3af3a23bfb4533888b20b57/index.js#L242

```js
const fd = new FormData
fd.append('test', new Blob(['data'], { type: 'application/octet-stream' }))

const opts =  { method: 'post', body: fd }
const r0 = new Request('/test', opts)
const r1 = new Request('/test?foo=bar', r0)
const r2 = new Request(r1, opts)

console.log(r2.headers.get('content-type'))
// multipart/form-data; boundary=---------------------------5618055171927212007156353948

await r2.text()
/*
"-----------------------------14683037551046622608547620079
Content-Disposition: form-data; name=\"test\"; filename=\"blob\"
Content-Type: application/octet-stream

data
-----------------------------14683037551046622608547620079--
"
*/
```

gydroperit:
We need to delete Content-Type header every time we create new instance of Request to properly set boundary in headers and in body.
Fixes #209

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,6 +46,7 @@ for (const property of globalProperties) {
 const isObject = value => value !== null && typeof value === 'object';
 const supportsAbortController = typeof globals.AbortController === 'function';
 const supportsStreams = typeof globals.ReadableStream === 'function';
+const supportsFormData = typeof globals.FormData === 'function';
 
 const deepMerge = (...sources) => {
 	let returnValue = {};
@@ -240,6 +241,12 @@ class Ky {
 		if (this._options.searchParams) {
 			const url = new URL(this.request.url);
 			url.search = new URLSearchParams(this._options.searchParams);
+
+			// To provide correct form boundary, Content-Type header should be deleted each time when new Request instantiated from another one
+			if (((supportsFormData && this._options.body instanceof globals.FormData) || this._options.body instanceof URLSearchParams) && !(this._options.headers && this._options.headers['content-type'])) {
+				this.request.headers.delete('content-type');
+			}
+
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
--- a/test/browser.js
+++ b/test/browser.js
@@ -173,7 +173,7 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
 });
 
 test('FormData with searchParams', withPage, async (t, page) => {
-	t.plan(2);
+	t.plan(3);
 
 	const server = await createTestServer();
 	server.get('/', (request, response) => {
@@ -181,10 +181,12 @@ test('FormData with searchParams', withPage, async (t, page) => {
 	});
 	server.post('/', async (request, response) => {
 		const requestBody = await pBody(request);
+		const contentType = request.headers['content-type'];
+		const boundary = contentType.split('boundary=')[1];
 
+		t.truthy(requestBody.includes(boundary));
 		t.regex(requestBody, /bubblegum pie/);
 		t.deepEqual(request.query, {foo: '1'});
-
 		response.end();
 	});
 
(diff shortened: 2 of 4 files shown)


sindresorhus:
// @rangermeier

rangermeier:
Not sure if it's a good idea to *always* remove `Content-Type` header for multipart requests. As discussed in https://github.com/sindresorhus/ky/issues/204#issuecomment-554613000 there might be need for setting the content-type when sending multipart.

 I think it would be safer to move the new code into its preceding `if()` statement:
```
		if (this._options.searchParams) {
			const url = new URL(this.request.url);
			url.search = new URLSearchParams(this._options.searchParams);

			if (((supportsFormData && this._options.body instanceof globals.FormData) || this._options.body instanceof URLSearchParams) && this.request.headers.has('content-type')) {
				this.request.headers.delete('content-type');
			}

		}
```

sholladay:
> I think it would be safer to move the new code into its preceding `if()` statement:

Agreed. At the end of the day, this is a hack, and thus the scope of its impact should be reduced as much as possible so it only takes effect when really necessary. I also think it's worth leaving a code comment here describing why it's being done, perhaps with a link to the original problem.

sindresorhus:
@gydroperit Ping :)

sindresorhus:
Tests are failing

gydroperit:
Added test cases for request with FormData and searchParams and custom content-type header

sindresorhus:
> test case fails because of some strange puppeteer error

Try to do some debugging. You could start by Googling the error.

gydroperit:
Did it

test/browser.js
```diff
@@ -226,6 +227,10 @@ test.failing('FormData with searchParams ("multipart/form-data" parser)', withPa
 					resolve([null, error_]);
 				}
 			});
+			busboy.on('finish', () => {
+				response.writeHead(303, {Connection: 'close', Location: '/'});
+				response.end();
+			});
```
sholladay: Why is listening for `finish` necessary?

It's not clear to me when the `response.end()` on line 232 will be called vs the one on line 238. Is calling `response.end()` twice safe?

test/browser.js
```diff
@@ -173,18 +173,20 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
...
 	const server = await createTestServer();
 	server.get('/', (request, response) => {
 		response.end();
 	});
 	server.post('/', async (request, response) => {
 		const requestBody = await pBody(request);
+		const contentType = request.headers['content-type'];
+		const boundary = contentType.split('boundary=')[1];
 
+		t.regex(requestBody, new RegExp(boundary));
```
sholladay: I don't think it's wise to dynamically construct a regular expression for use in a test assertion. If the `boundary` happens to include some regex special characters, it could cause the test to produce unexpected results.

I feel that `requestBody.includes(boundary)` would be better.

index.js
```diff
@@ -240,6 +241,11 @@ class Ky {
 		if (this._options.searchParams) {
 			const url = new URL(this.request.url);
 			url.search = new URLSearchParams(this._options.searchParams);
+
+			if (((supportsFormData && this._options.body instanceof globals.FormData) || this._options.body instanceof URLSearchParams) && !(this._options.headers && this._options.headers['content-type'])) {
```
sholladay: Leave a code comment that links to the issue explaining this hack, as it's not obvious to new contributors why we do this.

test/browser.js
gydroperit: I deleted `response.end()` on line 238. But we need `response.writeHead(303, {Connection: 'close', Location: '/'});` to safely end response, without that -  puppeteer is crashing on `page.close()` line in withPage helper.

test/browser.js
```diff
@@ -226,6 +227,10 @@ test.failing('FormData with searchParams ("multipart/form-data" parser)', withPa
 					resolve([null, error_]);
```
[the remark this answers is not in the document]
gydroperit: Ok

gydroperit:
Ping

Issue #179, opened 2019-10-01
Port tests from Got


sindresorhus:
[Got](https://github.com/sindresorhus/got) has a very good coverage test-wise. Would be nice to port relevant tests that we're missing over to Ky.

Note: Not for https://github.com/sindresorhus/ky/blob/master/test/browser.js, but for everything else.

sindresorhus:
I looked through all the tests and could not really find any useful and relevant ones to port.

PR #214, opened 2019-12-08
Port header tests from Got

braedongough:
Porting headers re: #179 

Ported all tests that seemed relevant. These are the tests I didn't: 
- [does not set `accept-encoding` header when `options.decompress` is false'](https://github.com/sindresorhus/got/blob/62cdeec4838293bf61a80820b42ecc84c408c973/test/headers.ts#L60)
   - This did set accept-encoding header. I think `options.decompress` doesn't exist on fetch, so I've omitted. 
- [stream as `options.body` sets `content-length` header](https://github.com/sindresorhus/got/blob/62cdeec4838293bf61a80820b42ecc84c408c973/test/headers.ts#L163)
   - I'm not sure this is relevant for KY. If it is, I can add the test. 
- [throws on null value headers](https://github.com/sindresorhus/got/blob/62cdeec4838293bf61a80820b42ecc84c408c973/test/headers.ts#L186)
   - This threw a 404 `HTTPError`, not a `TypeError`. Maybe I'm doing something wrong? 
- [non-existent headers set to undefined are omitted](https://github.com/sindresorhus/got/blob/62cdeec4838293bf61a80820b42ecc84c408c973/test/headers.ts#L210)
   - These weren't omitted. It looked to me this functionality isn't support yet it KY. 

Additionally, there were two other tests where ended up with different assertions but opted to keep them in:
- [accept` header with `json` option'](https://github.com/sindresorhus/got/blob/62cdeec4838293bf61a80820b42ecc84c408c973/test/headers.ts#L70)
   - When setting `accept` to `''`, `headers.accept` was still 'application/json' 
- [setting `content-length` to 0'](https://github.com/sindresorhus/got/blob/62cdeec4838293bf61a80820b42ecc84c408c973/test/headers.ts#L103)
   - Setting content length to 0 did not set the content length to 0.

### DIFF
--- a/test/headers.js
+++ b/test/headers.js
@@ -198,6 +198,19 @@ test('removes undefined value headers', async t => {
 	t.is(headers['user-agent'], 'undefined');
 });
 
+test.failing('non-existent headers set to undefined are omitted', async t => {
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky.get(server.url, {
+		headers: {
+			blah: undefined
+		}
+	}).json();
+
+	t.false(Reflect.has(headers, 'blah'));
+});
+
 test('preserve port in host header if non-standard port', async t => {
 	const server = await createTestServer();
 	server.get('/', echoHeaders);
--- a/test/headers.js
+++ b/test/headers.js
@@ -1,0 +1,227 @@
+import createTestServer from 'create-test-server';
+import FormData from 'form-data';
+import test from 'ava';
+import ky from '..';
+
+const echoHeaders = (request, response) => {
+	request.resume();
+	response.end(JSON.stringify(request.headers));
+};
+
+test('`user-agent`', async t => {
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky.get(server.url).json();
+	t.is(
+		headers['user-agent'],
+		'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)'
+	);
+});
+
+test('`accept-encoding`', async t => {
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky.get(server.url).json();
+
+	t.is(headers['accept-encoding'], 'gzip,deflate');
+});
+
+test('does not override provided `accept-encoding`', async t => {
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky
+		.get(server.url, {
+			headers: {
+				'accept-encoding': 'gzip'
+			}
+		})
+		.json();
+	t.is(headers['accept-encoding'], 'gzip');
+});
+
+test('does not remove user headers from `url` object argument', async t => {
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky
+		.get(server.url, {
+			hostname: server.hostname,
+			port: server.port,
+			responseType: 'json',
+			protocol: 'http:',
+			headers: {
+				'X-Request-Id': 'value'
+			}
+		})
+		.json();
+
+	t.is(headers.accept, 'application/json');
+	t.is(
+		headers['user-agent'],
+		'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)'
+	);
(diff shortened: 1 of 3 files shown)


sholladay:
> stream as options.body sets content-length header
>     I'm not sure this is relevant for KY. If it is, I can add the test.

Seems like it should probably apply to Ky as well. But how the heck does Got know the content length of a stream? Maybe I'm misunderstanding but I don't see how this could be.

> non-existent headers set to undefined are omitted
>     These weren't omitted. It looked to me this functionality isn't support yet it KY.

This is being worked on in PR #197. It would be great to have those tests when landing that feature. Maybe we should add the tests here but mark them with `.failing`.

braedongough:
@sholladay I've added a `.failing` for 
> non-existent headers set to undefined are omitted

Regarding
>stream as options.body sets content-length header
I'm not super familiar with stream to be honest. In my my attempts, I couldn't get it to pass when importing `fs` and `path`. Should it remain omitted for now? 
If you think it should be included, I'll keep messing around for a bit and see if I can make it work.

sholladay:
> Should it remain omitted for now?

I'm fine with that, personally. It sounds like strange behavior anyway. If the request is sent with `Transfer-Encoding: chunked` (which I assume it would be for a stream), then a `Content-Type` is not even allowed, per the HTTP specification. (The whole point of a stream is to start sending data in chunks before the whole file is read, hence its length is unknown.)

test/headers.js
```diff
@@ -0,0 +1,240 @@
...
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky.get(server.url, {
+		headers: {
+			blah: undefined
+		}
+	}).json();
+
+	t.false(Reflect.has(headers, 'blah'));
```
sholladay: nit: personally, I prefer using `'blah' in headers` over `Reflect.has(headers, 'blah')`. Not sure how @sindresorhus feels about it, though.

test/headers.js
```diff
@@ -0,0 +1,240 @@
...
+test('does not remove user headers from `url` object argument', async t => {
+	const server = await createTestServer();
+	server.get('/', echoHeaders);
+
+	const headers = await ky
+		.get(server.url, {
+			hostname: server.hostname,
+			port: server.port,
+			responseType: 'json',
+			protocol: 'http:',
```
sholladay: Most of these properties are not supported by Ky. I guess we could keep it anyway, if having identical tests is worth some cruft, which I'm not entirely against. But it probably makes the test less valuable, which isn't entirely obvious just from glancing at it.

I suppose what I'm getting at is that we should decide how much we want to tailor these tests for Ky vs leave them alone.

test/headers.js
szmarczak: <details>
<summary>I did some benchmarks</summary>

image

</details>

sindresorhus: Micro-benchmarks are not that useful. I don't feel strongly either way other than being consistent. 

Is `Reflect.has` currently enforced? If not, maybe worth opening an issue on https://github.com/sindresorhus/eslint-plugin-unicorn about adding a rule for enforcing either of the options?

I think I started using `Reflect.has` initially as I naively thought it was stricter, but it's the same, so maybe `in` is just better.

sindresorhus:
> I'm fine with that, personally. It sounds like strange behavior anyway. If the request is sent with Transfer-Encoding: chunked (which I assume it would be for a stream), then a Content-Type is not even allowed, per the HTTP specification. (The whole point of a stream is to start sending data in chunks before the whole file is read, hence its length is unknown.)

@szmarczak Do we need to fix this in Got?

test/headers.js
sindresorhus: I think we should tailor the tests for Ky. Having unsupported properties in there just makes it confusing to read the tests.

@braedongough Can you remove unsupported properties?

szmarczak:
We do.

braedongough:
Apologies for the delay, I lost track of this over the holidays. 

I've addressed the PR comments and re-requested review! Apologies @sholladay, I accidentally re-requested yours as well.  

Is there anything else that needs to be addressed on this? 😄

sindresorhus:
This is excellent. Thank you 👌

PR #219, opened 2020-01-13
Fix search params for react native

alexevdakov:
This PR will fix search params using React Native.

### DIFF
--- a/index.js
+++ b/index.js
@@ -238,7 +238,18 @@ class Ky {
 
 		if (this._options.searchParams) {
 			const url = new URL(this.request.url);
-			url.search = new URLSearchParams(this._options.searchParams);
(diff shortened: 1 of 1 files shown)


sholladay:
Some problems...
 - It is unclear what problem is being solved, because there was no bug report and there are no code comments.
 - The code is dirty (e.g. using private and presumably undocumented properties).
 - The solution is brittle and there are no tests, so we are very prone to breaking it by accident at any point.
 - This appears to be a workaround for an issue with the host environment itself, not a problem with Ky or our usage of any APIs.

On that basis, I am inclined to close this PR until a proper bug report and research can be applied. Then we should ping the React Native team, if appropriate, and ask for their feedback.

sindresorhus:
I'm sorry, but we're not interested in working around React Native quirks.

PR #223, opened 2020-01-16
Fix handling of falsy value with the `json` option

### LINKED ISSUE #222: falsey values in `json` prop are not sent to server
When submitting a POST request (not exclusively, this is just where I encountered this bug), if a falsey value (`0`, `false`, etc) is provided to the json prop, nothing is sent in the request body. 

Conversely, values like `1` and `true` *are* sent, as expected.

elijahsmith:
Currently, falsey json values will get mistaken for an empty json value -- this makes the comparison more specific, and will match any json value other than `undefined`

Fixes #222

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -166,7 +166,7 @@ export interface Options extends RequestInit {
 
 	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: object | Array | string | number | boolean | null;
+	json?: unknown;
 
 	/**
 	Search parameters to include in the request URL. Setting this will override all existing search parameters in the input URL.
--- a/index.d.ts
+++ b/index.d.ts
@@ -164,9 +164,9 @@ export interface Options extends RequestInit {
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 
-	Accepts a plain object which will be `JSON.stringify()`'d and the correct header will be set for you.
+	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: unknown;
+	json?: any;
 
 	/**
 	Search parameters to include in the request URL. Setting this will override all existing search parameters in the input URL.
--- a/index.d.ts
+++ b/index.d.ts
@@ -166,7 +166,7 @@ export interface Options extends RequestInit {
 
 	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: object | Array | string | number | boolean | null | undefined;
+	json?: object | Array | string | number | boolean | null;
 
 	/**
 	Search parameters to include in the request URL. Setting this will override all existing search parameters in the input URL.
--- a/index.d.ts
+++ b/index.d.ts
@@ -166,7 +166,7 @@ export interface Options extends RequestInit {
 
 	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: any;
+	json?: object | Array | string | number | boolean | null | undefined;
 
 	/**
 	Search parameters to include in the request URL. Setting this will override all existing search parameters in the input URL.
--- a/index.js
+++ b/index.js
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if ('json' in this._options) {
+		if (this._options.json !== undefined) {
 			this._options.body = JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', 'application/json');
 			this.request = new globals.Request(this.request, {body: this._options.body});
--- a/index.js
+++ b/index.js
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if (this._options.json) {
+		if (this._options.json !== undefined) {
 			this._options.body = JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', 'application/json');
 			this.request = new globals.Request(this.request, {body: this._options.body});
--- a/index.js
+++ b/index.js
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if (this._options.json !== undefined) {
+		if (this._options.hasOwnProperty('json')) {
 			this._options.body = JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', 'application/json');
 			this.request = new globals.Request(this.request, {body: this._options.body});
--- a/index.js
+++ b/index.js
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if (Object.prototype.hasOwnProperty.call(this._options, 'json')) {
+		if ('json' in this._options) {
 			this._options.body = JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', 'application/json');
 			this.request = new globals.Request(this.request, {body: this._options.body});
--- a/index.js
+++ b/index.js
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if (this._options.hasOwnProperty('json')) {
+		if (Object.prototype.hasOwnProperty.call(this._options, 'json')) {
 			this._options.body = JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', 'application/json');
 			this.request = new globals.Request(this.request, {body: this._options.body});
--- a/readme.md
+++ b/readme.md
@@ -139,7 +139,7 @@ Internally, the standard methods (`GET`, `POST`, `PUT`, `PATCH`, `HEAD` and `DEL
 
 ##### json
 
-Type: `object | Array | string | number | boolean | null | undefined`
+Type: `object | Array | string | number | boolean | null`
 
 Shortcut for sending JSON. Use this instead of the `body` option. Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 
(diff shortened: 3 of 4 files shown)


sindresorhus:
Would you be able to add a test too? So we can be sure to not regress it in the future. https://github.com/sindresorhus/ky/blob/master/test/main.js

elijahsmith:
I've changed the implementation just a bit -- now, setting `json` to *anything* is functionally equivalent to setting `body` to the `JSON.serialize`d result of the same anything... including for edge cases like undefined!

I also updated docs to reflect this (set their types to `any` and updated the descriptions).

Thanks for the feedback!

index.js
```diff
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if (this._options.json) {
+		if (Object.prototype.hasOwnProperty.call(this._options, 'json')) {
```
szmarczak: `'json' in this._options`

index.d.ts
```diff
@@ -164,9 +164,9 @@ export interface Options extends RequestInit {
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 
-	Accepts a plain object which will be `JSON.stringify()`'d and the correct header will be set for you.
+	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: unknown;
+	json?: any;
```
szmarczak: This should be correctly typed. The keys are strings and the values can be: string, number, boolean, null, object, array.

index.js
elijahsmith: done

index.d.ts
elijahsmith: I'm not sure what you mean by "the keys are strings".  I've updated the type to explicitly list those types, and included `undefined`, which is technically allowed (though it serializes to `null`...)

sholladay: For TypeScript purposes, I think it should probably stay as `unknown`.

https://stackoverflow.com/questions/51439843/unknown-vs-any/51439876

index.js
```diff
@@ -242,7 +242,7 @@ class Ky {
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
-		if (this._options.json !== undefined) {
+		if (this._options.hasOwnProperty('json')) {
```
sholladay: This will make it hard to _unset_ the `json` option. Also, it will try to stringify and send `undefined` which will cause `JSON.parse()` on the server to throw an error.

I think `!== undefined` is the way to go.

index.d.ts
```diff
@@ -164,9 +164,9 @@ export interface Options extends RequestInit {
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 
-	Accepts a plain object which will be `JSON.stringify()`'d and the correct header will be set for you.
+	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: unknown;
+	json?: object | Array | string | number | boolean | null | undefined;
```
szmarczak: undefined is unnecessary here

readme.md
```diff
@@ -139,9 +139,9 @@ Internally, the standard methods (`GET`, `POST`, `PUT`, `PATCH`, `HEAD` and `DEL
 
 ##### json
 
-Type: `object`
+Type: `any`
```
sholladay: Maybe instead of an actual type (which would be very verbose), we should just say "Any valid JSON value", linking to a reference on MDN or elsewhere that specifies what that is.

index.d.ts
elijahsmith: that's fine... but what's your reasoning?  `undefined` is a valid value to pass the json prop here... that said, this is my first time working with ts so I'm not aware of the subtleties...

index.js
elijahsmith: to unset the prop, you'd `delete` it... the tricky thing here is that setting `json` is (described as) a "shortcut" for setting `.body = JSON.stringify(...)`, so presumably anything you can pass to `JSON.stringify` should be able to be passed to `json`!  and, `undefined` will be stringified to `null` so the server will be required to parse that (which should be fine).

Having the `json` prop act ever so slightly different than how it's described is IMHO a recipe for weird edge cases...

readme.md
elijahsmith: this is where I can't contribute, as I have no ts experience. I'm happy to go with whatever y'all decide...

elijahsmith:
the build is now failing on the test `headers › non-existent headers set to undefined are omitted`, but I can't see why!  it started failing when I changed to using `'json' in this._options`.  Changes to the docs or .ts files would not trigger tests failures, would they?

readme.md
sholladay: This is just the documentation for the API, my comment isn't TypeScript related.

index.js
sholladay: > `undefined` will be stringified to `null`

This is incorrect. `JSON.stringify(undefined)` returns `undefined`, which `JSON.parse()` fails to parse.

index.d.ts
```diff
@@ -164,9 +164,9 @@ export interface Options extends RequestInit {
 	/**
 	Shortcut for sending JSON. Use this instead of the `body` option.
 
-	Accepts a plain object which will be `JSON.stringify()`'d and the correct header will be set for you.
+	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: unknown;
+	json?: object | Array | string | number | boolean | null;
```
sholladay: I am going to undo this change because it is unrelated to supporting non-object JSON values and there is some history behind why the TypeScript type is `unknown`. @sindresorhus made this change and explained it in https://github.com/sindresorhus/ky/pull/133#discussion_r282433699. As far as I know, there still isn't a great solution for that

We can talk about it in the comments here in this PR or you can open a new issue if you want the TypeScript types to be stricter. For now, I think `unknown` is sufficient to support the changes in this PR.

szmarczak: I'm :-1: on undoing this. I think Sindre was talking about response, not request. When it comes to parsing response, then yes, you have to cast it on your own because you don't know if it's a string, a number or whatever (one of the types mentioned later in this comment). On the other hand, we don't care about the JSON body as long as it is valid JSON. See, it's valid only when it is `string | number | boolean | null | {[key: string]: ...} | Array<...>`.

@sindresorhus Can you clarify please?

sholladay: Apologies, I linked to the wrong comment. I edited and fixed the link.

szmarczak: > Apologies, I linked to the wrong comment. I edited and fixed the link.

No problem.

> And it didn't account for when your have custom types with a toJSON() method.

Well, what about this:

```ts
interface JSONable extends Object {
    toJSON(): string;
}
```

(not sure if it works, not at home atm)

sindresorhus: `unknown` is the only pragmatic type here. I've tried making it strict before and it just turned out to be an inconvenience to users without much value.

sindresorhus:
Thanks for fixing this 👌

elijahsmith:
I can open a new ticket for this discussion if you'd rather, but, I'm not completely comfortable with the way this was finally resolved.  

With the current fix, given 
```
var foo;  // undefined
```

there is a functional difference between:
```
ky.put('/endpoint', {
    body: JSON.serialize(foo)
};
```
...which will send the body `null` to the server, and:
```
ky.put('/endpoint', {
    json: foo
};
```
...which will send no body at all.

I know this seems like a little thing but, especially with low-level libraries like this, small deviations are surprising and frustrating.  Is there a reason not to use `'json' in this._options` or `Object.hasOwnProperty(this._options, 'json')`?

sholladay:
> I can open a new ticket for this discussion if you'd rather, but, I'm not completely comfortable with the way this was finally resolved.

I am happy to discuss it further. Let's continue here for now.

> there is a functional difference between ...
> `body: JSON.serialize(foo)`
> ...
> `json: foo`

No, there isn't. You have made some faulty assumptions.

You seem to believe that `undefined` serializes to `null`, but as I mentioned [here](https://github.com/sindresorhus/ky/pull/223#discussion_r368183608), that is incorrect. In reality, `undefined === JSON.stringify(undefined)`. Both are equivalent: they send no body and the `Content-Length` is `0`, regardless of whether you use the `json` option or the `body` option.

> Is there a reason not to use `'json' in this._options` ... ?

Yes. As I mentioned [here](https://github.com/sindresorhus/ky/pull/223#discussion_r368102826), this would make it hard to _unset_ the `json` option, especially on extended instances. You are assuming that the user can `delete options.json` to unset it, but that is only true if the user uses `ky()` directly. Here is an example where that would fail, because the [options merger](https://github.com/sindresorhus/ky/blob/bb46ca86e36d8998c0425ac1fa3b3faf9972c82b/index.js#L50-L72) used by `ky.extend()` wouldn't be able to detect that you `delete`'d a property, so it would simply think you _aren't overriding_ `json`.

```js
const myOptions = {
    json : { time : Date.now() }
};
const myKy1 = ky.extend(myOptions);
myKy1.post('/ping');  // sends json

delete myOptions.json;  // does not affect myKy1, as options are already set
const myKy2 = ky.extend(myOptions);
myKy2.post('/ping');  // bug: still sends json, even though we intended not to
```

The above code is not "fixed" by using `this._options.json !== undefined`, but it _does_ mean you can unset the option by simply setting it to `undefined`, as shown below.

```js
const myOptions = {
    json : { time : Date.now() }
};
const myKy1 = ky.extend(myOptions);
myKy1.post('/ping');  // sends json

myOptions.json = undefined;  // does not affect myKy1, as options are already set
const myKy2 = ky.extend(myOptions);
myKy2.post('/ping');  // does not send json (yay!)
```

> Is there a reason not to use ... `Object.hasOwnProperty(this._options, 'json')`?

This has the same problems as `'json' in this._options`, in that overrides won't work correctly. Beyond that, I don't see a need for a `hasOwnProperty` check. We [spread the `options` argument](https://github.com/sindresorhus/ky/blob/bb46ca86e36d8998c0425ac1fa3b3faf9972c82b/index.js#L198) into `this._options`. All properties of `this._options` (other than the language built-ins) are own properties, including `json`. Yet it's also not meaningful whether `json` is an own property or a prototype property.

elijahsmith:
thanks for the followup...

> You seem to believe that undefined serializes to null, but as I mentioned here, that is incorrect. In reality, undefined === JSON.stringify(undefined)

this is correct, and I was wrong, I extrapolated incorrectly from `JSON.stringify([undefined]) == "[null]"`, and should have investigated further!   I appreciate the thorough response.

PR #227, opened 2020-01-28
Add failing test for using FormData with searchParams

umidbekkarimov:
More detailed test that reproduces the bug described in #209.

### DIFF
--- a/test/browser.js
+++ b/test/browser.js
@@ -203,7 +203,7 @@ test('FormData with searchParams', withPage, async (t, page) => {
 	await server.close();
 });
 
-// FIXME: More detailed test that reproduces the bug described in https://github.com/sindresorhus/ky/issues/209.
+// FIXME: More detailed test that reproduces the bug described in https://github.com/sindresorhus/ky/issues/209
 test.failing('FormData with searchParams ("multipart/form-data" parser)', withPage, async (t, page) => {
 	t.plan(3);
 	const server = await createTestServer();
--- a/test/browser.js
+++ b/test/browser.js
@@ -211,10 +211,10 @@ test.failing('FormData with searchParams ("multipart/form-data" parser)', withPa
 		response.end();
 	});
 	server.post('/', async (request, response) => {
-		t.deepEqual(request.query, {foo: '1'});
 		const body = await new Promise((resolve, reject) => {
 			const busboy = new Busboy({headers: request.headers});
-			busboy.on('file', async (fieldname, file, filename, encoding, mimetype) => {
+      busboy.on('error', reject);
+      busboy.on('file', async (fieldname, file, filename, encoding, mimetype) => {
 				let fileContent = '';
 				try {
 					for await (const chunk of file) {
@@ -226,10 +226,10 @@ test.failing('FormData with searchParams ("multipart/form-data" parser)', withPa
 					reject(error);
 				}
 			});
-			busboy.on('error', reject);
 			request.pipe(busboy);
 		});
-		t.deepEqual(body, {
+    t.deepEqual(request.query, {foo: '1'});
+    t.deepEqual(body, {
 			fieldname: 'file',
 			filename: 'my-file',
 			encoding: '7bit',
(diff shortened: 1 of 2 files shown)


sholladay:
Thank you for submitting a failing test case. This is very useful. I haven't checked yet whether the test case is fixed by PR #211, but I will soon.

Do you think there's an easy way to avoid depending on Busboy? We already have a utility for collecting the request body, seems like we just need a parser.

umidbekkarimov:
> Do you think there's an easy way to avoid depending on Busboy? We already have a utility for collecting the request body, seems like we just need a parser.

I tried to use [body/form](https://github.com/Raynos/body/blob/master/form.js), but it doesn't work `multipart/form-data`. I also tried to use `multer`, not sure why it didn't work with `create-test-server`, so finally, I ended up with `busboy`.

sindresorhus:
I'm fine with using Busboy. Can you fix the merge conflict?

umidbekkarimov:
@sindresorhus done, but I'm not sure why tests are hanging 🤔

test/browser.js
```diff
@@ -171,27 +172,48 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
 	await server.close();
 });
 
-test('FormData with searchParams', withPage, async (t, page) => {
-	t.plan(2);
-
+test.failing('FormData with searchParams', withPage, async (t, page) => {
```
sholladay: I think it would be good to leave the existing test in place and add this as a _new_ test. That should help illustrate the specifics of what is broken.

test/browser.js
```diff
@@ -171,27 +172,48 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
...
-
-		t.regex(requestBody, /bubblegum pie/);
 		t.deepEqual(request.query, {foo: '1'});
+		const body = await new Promise((resolve, reject) => {
+			const busboy = new Busboy({headers: request.headers});
+
+			busboy.on('file', (fieldname, file, filename, encoding, mimetype) => {
+				let fileContent = '';
+
+				file.on('data', data => {
```
sholladay: You should be able to use async iteration on the `file` stream.

```js
for await (const chunk of file) {
    fileContent += chunk;
}
```

This may simplify things, as errors will be thrown automatically and you should be able to `resolve()` after the loop without having to put it in an `end` callback.

Read more here:
https://2ality.com/2019/11/nodejs-streams-async-iteration.html#helper-function%3A-readabletostring()

test/browser.js
```diff
@@ -171,27 +172,48 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
...
+					fileContent += data.toString('utf-8');
```
sholladay: Not that it really matters, but I think `.toString()` is unnecessary.

test/browser.js
```diff
@@ -171,27 +172,48 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
...
+
+			busboy.on('error', reject);
+
+			request.pipe(busboy);
+		});
+		t.is(body.fieldname, 'file');
+		t.is(body.filename, 'my-file');
+		t.is(body.encoding, '7bit');
+		t.is(body.mimetype, 'text/plain');
+		t.is(body.fileContent, 'bubblegum pie');
```
sholladay: I think we should use `t.deepEqual()` to ensure we are asserting _all_ properties of `body`.

umidbekkarimov:
@sholladay done. And finally figured out why CI wasn't working.

test/browser.js
```diff
@@ -202,6 +203,60 @@ test('FormData with searchParams', withPage, async (t, page) => {
 	await server.close();
 });
 
+// FIXME: More detailed test that reproduces the bug described in https://github.com/sindresorhus/ky/issues/209.
```
szmarczak:
```suggestion
// FIXME: More detailed test that reproduces the bug described in https://github.com/sindresorhus/ky/issues/209
```

PR #231, opened 2020-01-30
Fix retries for requests with a body

### LINKED ISSUE #217: Issue with Retries after v0.16.0
This could very well be something I'm doing wrong or misunderstanding and if so I apologize.

Starting at v0.16.0 and continuing to v.0.16.1, there seems to be an issue with retries reusing a Request object (at least that's what the error tells me). Ky's fetch promise rejects with this:

```
TypeError: Failed to execute 'fetch' on 'Window': Cannot construct a Request with a Request object that has already been used.
    at Ky._fetch (index.js:395)
    at fn (index.js:257)
    at async Ky._retry (index.js:346)
    at async Promise.Ky.result.<computed> [as json] (index.js:299)
```

In v0.15.0 & below I would just get the error response from my API.

sholladay:
Fixes #217

I now clone the request before using it so that the body is not consumed and thus `fetch()` does not freak out when we try to reuse the request during retries, since it can still read the body.

I tried to write a test for this, however it seems to be hard to test in Node, as `node-fetch` doesn't seem to throw an error under the same circumstances that actual `fetch()` does in the browser. It should be possible to test this with Puppeteer, though I haven't tried.

### DIFF
--- a/test/browser.js
+++ b/test/browser.js
@@ -254,7 +254,7 @@ test('retry with body', withPage, async (t, page) => {
 	server.put('/test', async (request, response) => {
 		requestCount++;
 		await pBody(request);
-		response.sendStatus(408);
+		response.sendStatus(502);
 	});
 
 	await page.goto(server.url);
@@ -269,7 +269,7 @@ test('retry with body', withPage, async (t, page) => {
 		}).text();
 		return request.catch(error_ => error_.toString());
 	}, server.url);
-	t.is(error, 'HTTPError: Request Timeout');
+	t.is(error, 'HTTPError: Bad Gateway');
 	t.is(requestCount, 2);
 
 	await server.close();
--- a/test/browser.js
+++ b/test/browser.js
@@ -4,6 +4,8 @@ import {serial as test} from 'ava';
 import createTestServer from 'create-test-server';
 import withPage from './helpers/with-page';
 
+const pBody = util.promisify(body);
+
 test('prefixUrl option', withPage, async (t, page) => {
 	const server = await createTestServer();
 	server.get('/', (request, response) => {
@@ -73,22 +75,24 @@ test('throws TimeoutError even though it does not support AbortController', with
 		response.end();
 	});
 
-	server.get('/endless', () => {});
+	server.get('/slow', (request, response) => {
+		setTimeout(() => {
+			response.end('ok');
+		}, 1000);
+	});
 
 	await page.goto(server.url);
 	await page.addScriptTag({path: './test/helpers/disable-abort-controller.js'});
 	await page.addScriptTag({path: './umd.js'});
 
-	// TODO: make set a timeout for this evaluation so we don't have to wait 30s
 	const error = await page.evaluate(url => {
 		window.ky = window.ky.default;
 
-		const request = window.ky(`${url}/endless`, {timeout: 500}).text();
+		const request = window.ky(`${url}/slow`, {timeout: 500}).text();
 		return request.catch(error_ => error_.toString());
 	}, server.url);
 	t.is(error, 'TimeoutError: Request timed out');
 
-	// A note from @szmarczak: await server.close() hangs on my machine
 	await server.close();
 });
 
@@ -180,12 +184,13 @@ test('throws if does not support ReadableStream', withPage, async (t, page) => {
 });
 
 test('FormData with searchParams', withPage, async (t, page) => {
+	t.plan(2);
+
 	const server = await createTestServer();
 	server.get('/', (request, response) => {
 		response.end();
 	});
 	server.post('/', async (request, response) => {
-		const pBody = util.promisify(body);
 		const requestBody = await pBody(request);
 
 		t.regex(requestBody, /bubblegum pie/);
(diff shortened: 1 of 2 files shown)


DanielRuf:
Hm, MDN says:

https://developer.mozilla.org/en-US/docs/Web/API/Request/clone

> The clone() method of the Request interface creates a copy of the current Request object.
> 
> clone() throws a TypeError if the response Body has already been used. In fact, the main reason clone() exists is to allow multiple uses of Body objects (when they are one-use only.)
> 
> If intend to modify the request, you may prefer the Request constructor.

DanielRuf:
I think it makes sense to add a unit test which fails without the tests and should succeed with the changes.

sholladay:
> I think it makes sense to add a unit test which fails without the tests and should succeed with the changes.

Well, yes, except that `node-fetch` _doesn't fail_ when reusing the request. Hard to test for a failure that never happens. Like I said in the PR description, we'll probably need Puppeteer to properly test this, so that a real `Request` and `fetch` implementation is used. Otherwise the test will always succeed, even with versions of Ky that have the bug.

sindresorhus:
You could write the test in https://github.com/sindresorhus/ky/blob/master/test/browser.js ?

test/browser.js
```diff
@@ -73,22 +75,24 @@ test('throws TimeoutError even though it does not support AbortController', with
...
 	const error = await page.evaluate(url => {
 		window.ky = window.ky.default;
 
-		const request = window.ky(`${url}/endless`, {timeout: 500}).text();
+		const request = window.ky(`${url}/slow`, {timeout: 500}).text();
 		return request.catch(error_ => error_.toString());
 	}, server.url);
 	t.is(error, 'TimeoutError: Request timed out');
 
-	// A note from @szmarczak: await server.close() hangs on my machine
```
sholladay: It was hanging for me, too. I think because the server's `response` stream was never being closed. That behavior doesn't seem necessary for this test, so I simply made the Ky timeout less than the server timeout and it works as expected.

test/browser.js
```diff
@@ -236,3 +243,34 @@ test('headers are preserved when input is a Request and there are searchParams i
...
+		window.ky = window.ky.default;
+		const request = window.ky(url + '/test', {
+			body: 'foo',
+			method: 'PUT',
+			retry: 2
+		}).text();
+		return request.catch(error_ => error_.toString());
+	}, server.url);
+	t.is(error, 'HTTPError: Bad Gateway');
+	t.is(requestCount, 2);
```
sholladay: For reasons that I do not understand, the `requestCount` assertion was failing and I was only able to solve it by changing the status code from `408` to something else. When the status is `408`, the server receives roughly double the number of requests I am expecting (i.e. roughly 2x whatever `retry` is). When I set the `retry` option to `1`, then `requestCount` was `2` and the assertion passed. When I set `retry` to `8`, then `requestCount` was `14` and the assertion failed (note: not quite double).

So, of course it seems like a bug in Ky. And sure enough, when I debugged it with `await puppeteer.launch({ devtools: true });` in [test/helpers/with-page.js](https://github.com/sindresorhus/ky/blob/78814e74efaaab65728103a13541820441b05d93/test/helpers/with-page.js#L4), I was able to see these requests in the Network tab and the timings looked like they were respecting our [backoff algorithm](https://github.com/sindresorhus/ky/blob/78814e74efaaab65728103a13541820441b05d93/index.js#L337-L338), so these requests are likely being triggered within Ky.

The thing is, we have no special handling of `408`, nor `502` in Ky, so I don't see how this could be fixed by changing from `408` to `502`. I also wasn't able to easily reproduce this behavior manually in Chrome using endpoints from https://httpstat.us/. To me, it's a mystery.

I chose to just change the status code in commit https://github.com/sindresorhus/ky/pull/231/commits/2808b0fb953688cf7fdc04972822428581188582.

sholladay:
@sindresorhus I added a test for this. Took a lot of time due to https://github.com/sindresorhus/ky/pull/231#discussion_r373779133.

test/browser.js
szmarczak: Thank you! :relaxed:

tristanreid:
Hi there - this is in reference to a Puppeteer issue rather than a KY issue, but I hope it will help anyone struggling with this particular bug:

If you clone a POST request object and use it in a fetch, it will succeed, but if you intercept that request using puppeteer, [the `postData` will be empty](https://github.com/puppeteer/puppeteer/issues/6111).  This is causing me an issue because the site that I'm trying to screenshot with puppeteer uses KY, and ever since the fix to this retries issue that introduces `request.clone()`, the site's POST requests all fail.  This is only happening because of the combination of the puppeteer bug along with the fact that I need to proxy requests and therefore use their `postData`.  

I believe the proper fix should be puppeteer changing, not KY, and I don't expect this edge case to impact many people, but it was difficult for me to track down, so I want to share here as well as the bug that I filed over there.

sholladay:
That Puppeteer behavior is just... awful. The retries are implicit and seemingly undocumented. The retries are only for certain status codes without a logical pattern (possibly only `408`?). They aren't reported or exposed programmatically. They happen in Puppeteer but not Chrome. This has to be someone's idea of making tests more "reliable", giving everyone a false sense of security.

PR #235, opened 2020-02-03
Add ability to remove headers using `ky.extend()`

### LINKED ISSUE #189: Remove headers

whitecrownclown:
Closes #189

Hello! 
Summary of this PR: 

- Continues the work from previous PR #197
- Adds the possibility to remove headers by using ky.extend()
- `options.headers` supports both Headers instance & plain object
- Added tests
- Updated tsd & readme
- Moved globals and utility functions into globals.js, respectively utils.js (I can revert this change or update it if that's not the direction you want this going). My reasoning was that It's an easier way to read and test the code.

Thanks!

cc @sindresorhus @sholladay


utils.js
```diff
@@ -0,0 +1,66 @@
...
+		if (Array.isArray(source)) {
+			if (!(Array.isArray(returnValue))) {
+				returnValue = [];
+			}
+
+			returnValue = [...returnValue, ...source];
+		} else if (isObject(source)) {
+			for (let [key, value] of Object.entries(source)) {
+				if (isObject(value) && Reflect.has(returnValue, key)) {
+					if (key === 'headers') {
```
szmarczak: It shouldn't be done like this. You check **every** object for `headers`. Something like this would be much better:

```js
let headers = {};

for (const source of sources) {
	if (Array.isArray(source)) {
		...
	} else if (isObject(source)) {
		...
	}

	if (source.headers) {
		headers = mergeHeaders(headers, source.headers);
	}
}

returnValue.headers = headers;
```

sholladay:
I think the moving of globals and utils should be reverted. Otherwise, it is going to be inconvenient for people who [use Ky without a bundler](https://github.com/sindresorhus/ky/tree/d5cb15a362519232b6f6203359070a10dfabfdb0#how-do-i-use-this-without-a-bundler-like-webpack), as they'll have to upload multiple files. Those instructions would need to be updated, too. Also, testing `mergeHeaders` outside of Ky isn't super useful, since it's not meant to be a public API.

szmarczak:
> Otherwise, it is going to be inconvenient for people who use Ky without a bundler

Well, we already use rollup. We can just output one file for modern JavaScript (like we do for UMD). People still could use the CDN version or upload it manually etc.

utils.js
whitecrownclown: You're right, I'll update the PR.

readme.md
```diff
@@ -352,6 +352,29 @@ Create a new `ky` instance with some defaults overridden with your own.
...
+});
+
+const extended = original.extend({
+	headers: {
+		rainbow: undefined
+	}
+});
+
+console.log('rainbow' in extended); // false
+console.log('unicorn' in extended); // true
```
sholladay: This doesn't seem right. `extended.unicorn` is `true`?

sindresorhus:
> Well, we already use rollup. We can just output one file for modern JavaScript (like we do for UMD). People still could use the CDN version or upload it manually etc.

Having everything in one file has had one upside. File size. When it becomes easy to add more files and dependencies, the size usually tends to grow exponentially. Having everything in one file forces us to be careful.

index.d.ts
```diff
@@ -153,14 +153,42 @@ export interface RetryOptions {
...
+export interface Options extends Omit<RequestInit, 'headers'> {
 	/**
 	HTTP method used to make the request.
 
 	Internally, the standard methods (`GET`, `POST`, `PUT`, `PATCH`, `HEAD` and `DELETE`) are uppercased in order to avoid server errors due to case sensitivity.
 	*/
 	method?: LiteralUnion<'get' | 'post' | 'put' | 'delete' | 'patch' | 'head', string>;
 
+	/**
+	Override default Headers
```
sindresorhus: Make sure you see all the feedback given in the original PR.

https://github.com/sindresorhus/ky/pull/197#discussion_r345189241

readme.md
whitecrownclown: I've updated the PR. I meant `response` instead of extended.

sindresorhus:
Tests are failing.

index.js
```diff
@@ -47,8 +47,32 @@ const isObject = value => value !== null && typeof value === 'object';
...
 
+const mergeHeaders = (...sources) => {
+	const result = {};
+
+	for (const source of sources) {
+		if (!isObject(source)) {
+			throw new TypeError('The `headers` argument must be an object');
+		}
+
+		const headers = new globals.Headers(source);
```
szmarczak: This line is unnecessary

index.js
```diff
@@ -47,8 +47,32 @@ const isObject = value => value !== null && typeof value === 'object';
...
+
+		for (const [key, value] of headers) {
```
szmarczak: `of Object.entries(headers)`

index.js
```diff
@@ -47,8 +47,32 @@ const isObject = value => value !== null && typeof value === 'object';
...
+			// Headers constructor changes the value to a string
+			if (value === 'undefined' || typeof value === 'undefined') {
```
szmarczak: Replace these conditions with `value === undefined`

index.js
```diff
@@ -47,8 +47,32 @@ const isObject = value => value !== null && typeof value === 'object';
...
+				Reflect.deleteProperty(result, key);
```
szmarczak: `delete result[key];`

index.js
```diff
@@ -47,8 +47,32 @@ const isObject = value => value !== null && typeof value === 'object';
...
+			} else {
+				Reflect.set(result, key, value);
```
szmarczak: `result[key] = value;`

index.js
szmarczak: Oh, actually it is necessary since the input can be a Request instance, and `request.headers` is an instance of `Headers` :+1:

index.js
szmarczak: [Wait, well done.](https://github.com/sindresorhus/ky/pull/235/files#r382930813)

index.js
szmarczak: It should be mentioned in the readme that `{something: 'undefined'}` and `{something: undefined}` will be omitted.

sholladay: IMO, `'undefined'` should only be omitted if it comes from a `Headers` instance.

szmarczak: > IMO, 'undefined' (as a string) should only be omitted if it comes from a Headers instance.

Sure, as long as it is mentioned in the readme :)

whitecrownclown:
As a heads up, I'm working on the feedback and will update the PR as soon as possible. Thanks!

sindresorhus:
@whitecrownclown What's the status on this?

whitecrownclown:
@sindresorhus Sorry for the delay.

I couldn't figure out why the test `remove undefined headers` actually failed. The problem is that [node-fetch](https://github.com/node-fetch/node-fetch/) has default headers that we cannot override. It seems they've added a way to overwrite this particular `user-agent` default in a [beta release](https://github.com/node-fetch/node-fetch/blob/master/docs/CHANGELOG.md#v300). I will add a comment indicating this and mark the test as failing. I initially thought the issue was from my PR not doing things right and that's why this took so long.

sindresorhus:
I think you missed https://github.com/sindresorhus/ky/pull/235/files#r382929869 and https://github.com/sindresorhus/ky/pull/235/files#r382929133

test/headers.js
```diff
@@ -212,7 +212,8 @@ test('buffer as `options.body` sets `content-length` header', async t => {
 	t.is(Number(headers['content-length']), buffer.length);
 });
 
-test('removes undefined value headers', async t => {
+// Todo: Enable this when node-fetch allows for removal of default headers
```
sindresorhus, line 215:
```suggestion
// TODO: Enable this when node-fetch allows for removal of default headers
```

and link to the relevant issue or commit.

whitecrownclown:
I've updated the PR.

index.js
```diff
@@ -48,10 +48,10 @@ const supportsAbortController = typeof globals.AbortController === 'function';
 const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
-const mergeHeaders = (...sources) => {
+const mergeHeaders = (source1, source2) => {
 	const result = new globals.Headers();
 
-	for (const source of sources) {
+	[source1, source2].forEach(source => {
```
sindresorhus, line 54: I would still prefer a `for-of` loop.

readme.md
```diff
@@ -352,6 +352,38 @@ Create a new `ky` instance with some defaults overridden with your own.
...
+	}
+});
+
+const extended = original.extend({
+	headers: {
+		rainbow: undefined
+	}
+});
+
+const response = await extended(`${url}/test`).json();
```
szmarczak, line 378: Missing context. What is `${url}/test`? Use `https://httpbin.org/anything` instead. Same for the docs in TS definitions file.

index.js
```diff
@@ -48,8 +48,27 @@ const supportsAbortController = typeof globals.AbortController === 'function';
 const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
+const mergeHeaders = (source1, source2) => {
```
szmarczak, line 51:
```suggestion
const mergeHeaders = (...args) => {
```

index.js
```diff
@@ -48,8 +48,27 @@ const supportsAbortController = typeof globals.AbortController === 'function';
...
+	const result = new globals.Headers();
+
+	for (const source of [source1, source2]) {
```
szmarczak, line 54:
```suggestion
	for (const source of args) {
```

index.js
```diff
@@ -48,8 +48,27 @@ const supportsAbortController = typeof globals.AbortController === 'function';
...
+		const isHeadersInstance = source instanceof globals.Headers;
+
+		for (const [key, value] of new globals.Headers(source)) {
```
szmarczak, line 57: Define `new globals.Headers(source)` before the `for` loop. Two keywords near each other? It should be illegal :P

Alternatively you can do this:

```suggestion
		const iterator = isHeadersInstance ? headers.entries() : Object.entries(headers);

		for (const [key, value] of iterator) {
```

index.js
whitecrownclown: That's exactly the opposite of what @sholladay asked me to do.

readme.md
whitecrownclown: Will do.

index.js
sholladay: Sorry for the conflicting requests. My stance on this is that it's basically a reducer function, should take two arguments only (existing headers and override headers), and return the computed headers. To do that, I'm pretty sure we only need to coerce the first argument to a `Headers` instance and then loop over the second argument to override the header values. I don't think we need to loop over the list of arguments. But I could be wrong.

If it's true that we actually do need to loop over the arguments list, then I am fine with using spread syntax here, as it's pretty equivalent to making the array ourselves..

whitecrownclown: You are correct. I initially had it this way, I don't recall why It was changed to this form. Thanks! @sholladay

index.js
whitecrownclown: I agree with the keywords :)
The iterator does not differ, either way, the source is now always a `Headers` instance.

index.js
szmarczak: I'm fine with either :)

index.js
```diff
@@ -205,6 +228,7 @@ class Ky {
 			// TODO: credentials can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
 			credentials: this._input.credentials || 'same-origin',
 			...options,
+			headers: mergeHeaders(this._input instanceof globals.Request ? this._input.headers : {}, options.headers),
```
szmarczak, line 231: This can be simplified to:

```suggestion
			headers: mergeHeaders(this._input.headers, options.headers),
```

PR #238, opened 2020-02-22
Fix request cancellation on timeout

### LINKED ISSUE #221: Request not cancelled when timeout occurs
When a Ky request times out AND no explicit signal was passed in the options, the request is not being cancelled.

[Reproduction stackblitz here](https://stackblitz.com/edit/react-ts-udrt4s?file=index.tsx) - when the code runs it makes two requests to GET root.  In Chrome's dev tools, the first request is shown to be cancelled, but the second is not.

Looking at the code, it appears the issue is with [this line](https://github.com/sindresorhus/ky/blob/v0.16.1/index.js#L233-L233). The `this._options.signal` is only set if there's already a signal set on the options.

This initial PR that added the cancel functionality [had this coded differently](https://github.com/sindresorhus/ky/pull/27/files#diff-168726dbe96b3ce427e7fedce31bb0bcR149-R157) but it was changed in [PR 180](https://github.com/sindresorhus/ky/pull/180/files#diff-168726dbe96b3ce427e7fedce31bb0bcL280-L283).  It does not appear that the change in behaviour was intended?

I've worked around this issue for now by setting a (never used) signal in the default options for ky, but I thought it worth reporting.

Edit: Tested Ky version: 0.16.1

sholladay:
Fixes #221 

This PR fixes a regression that was caused by a simple refactoring mistake in PR #180, which caused the request to not be aborted after a TimeoutError.

Unfortunately, while I did confirm that it works, I haven't figured out how to write a test for it. I was expecting that the abort operation would cause the client to hang up the connection, and I figured that would cause an `error` event on the test server's `request` object that I could watch for. But the abort doesn't seem to cause such an error to occur. I'm not familiar enough with the details of what AbortController does at the network connection / HTTP layer to know if there's anything else to look for.

### DIFF
--- a/index.js
+++ b/index.js
@@ -230,8 +230,9 @@ class Ky {
 				this._options.signal.addEventListener('abort', () => {
 					this.abortController.abort();
 				});
-				this._options.signal = this.abortController.signal;
 			}
+
+			this._options.signal = this.abortController.signal;
 		}
 
 		this.request = new globals.Request(this._input, this._options);


sindresorhus:
> I haven't figured out how to write a test for it. I was expecting that the abort operation would cause the client to hang up the connection

Can you open an issue about writing a test for this? So we'll get to it eventually.

sholladay:
> Can you open an issue ...

Done: #239

sindresorhus:
Great. Thanks for fixing this.

Would you be able to take a quick look at https://github.com/sindresorhus/ky/pull/235? I'd like to get out a new version today.

PR #241, opened 2020-02-29
Add TimeoutError and HTTPError to ky TypeScript namespace

### LINKED ISSUE #237: HTTPError type is not exported
### Problem

In 0.13, we could `import { HTTPError } from 'ky'` in TypeScript.
In 0.17 the value is accessible as `ky.HTTPError`, but the type is `InstanceType<typeof ky.HTTPError>`, what's a bit inconvenient.

### Possible solution

Instead of `declare const ky` there could be a `namespace ky` in `index.d.ts`. Namespaces can contain types.

hasparus:
Closes #237

### DIFF
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -142,15 +142,15 @@ ky(url, {
 });
 
 try {
-	ky.get(url);
-} catch (err) {
-	if (err instanceof ky.HTTPError) {
-		expectType<ky.HTTPError>(err);
-		const status = err.response.status;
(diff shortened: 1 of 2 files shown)


index.test-d.ts
```diff
@@ -140,3 +140,17 @@ ky(url, {
 		afterStatusCodes: []
 	}
 });
+
+try {
+	await ky.get(url);
```
hasparus: :heart:

PR #242, opened 2020-03-05
Make ky usable in snowpack (and other places where ky is imported as module)

ubergesundheit:
I wanted to use ky as a module with snowpack, but it seems for this to work packages need a `module` key in their package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
 		"umd.js",
 		"umd.d.ts"
 	],
+	"main": "index.js",
+	"module": "index.js",
 	"keywords": [
 		"fetch",
 		"request",


sholladay:
Seems like a problem with the bundler. Rollup and Webpack handle it just fine.

We did discuss doing this before:
https://github.com/sindresorhus/ky/pull/81

ubergesundheit:
Ok I see. Would you accept this if I would change the `main` field to `umd.js` suggested in https://github.com/sindresorhus/ky/pull/81#issuecomment-456511883 ?

sholladay:
I'm neutral on it, mostly because I see it as inverting the problem without any clear advantage - the burden is now on tools to understand a new field in package.json rather than assuming or automatically handling ESM syntax. I don't know to what extent this understanding is deployed. Sindre pointed out [a couple of major bugs](https://github.com/sindresorhus/ky/pull/81#issuecomment-454729544) in Webpack that remain unresolved in regards to usage of the `module` field, so it would appear that this is no silver bullet.

Let's see if the other maintainers have stronger feelings about it.

sindresorhus:
This should be an issue on the Snowpack repo instead.

sindresorhus:
@sholladay Slightly unrelated and I'm not sure if bundlers type it into account, but we should probably addd a `type` field? https://nodejs.org/api/esm.html#esm_package_json_type_field And we could use the `exports` field for the `umd.js` file. Thoughts?

ubergesundheit:
Is there any middle ground we could agree on? Looking at the [snowpack source part of how it decides if a package has module ready code or not](https://github.com/pikapkg/snowpack/blob/master/src/index.ts#L156-L160), `'browser:module`, `main:esnext` and `browser` also seem to work.

If you don't want to add non-standard fields to your package.json, I think your suggestion of adding a `"type": "module"` field would be a very good starting point for adding support for this key to snowpack

sholladay:
I have a suspicion that `"type": "module"` will break users who do `require('ky/umd')`, as it will be treated as an ES module. That seems problematic to me.

If I'm wrong about that, then yeah, I'm fine with using the `type` field.

sindresorhus:
@sholladay That’s why I’m suggesting using the `exports` field too.

sholladay:
Maybe I missed something, but it seems like `exports` is just a way to define one or more entry points. I think it will still be treated as ESM though, which won't work with `require`.

ubergesundheit:
So I would suggest changing my branch to feature a `type` and `exports` field. I’ll see if I can then import it using webpack, and a require based loader (I’m open to suggestions).

sindresorhus:
@sholladay See https://nodejs.org/api/esm.html#esm_conditional_exports

@ubergesundheit If you want to do a new PR to add `type` and `exports` that would be great.

PR #243, opened 2020-03-08
Fix the NormalizedOptions.headers TypeScript type

### LINKED ISSUE #240: NormalizedOptions interface giving wrong Headers type
The issue I'm experience specifically is the options param in the hook functions. I'm trying to set the `Authorization` header with `options.headers['Authorization'] = token` but `NormalizedOptions.headers` claims to be a `Headers` class, when it's actually just a Record of key/value strings like in the `create`/`extend` options.

sholladay:
Fixes #240 

This PR removes a vestigial part of our `NormalizedOptions` TypeScript interface that is no longer necessary and causes issues for TypeScript users who pass the `headers` option as a plain object (as opposed to a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance) and then receive it inside of a hook. The definition of `headers` now comes from `RequestInit`.

For context, in older versions of Ky, we aggressively normalized the options object passed by the user before running any hooks. However, newer versions of Ky try to avoid such normalizations in favor of constructing a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) instance and passing that around, as it does a lot of the work for us, including normalizing the headers.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -286,9 +286,6 @@ export interface NormalizedOptions extends RequestInit {
 	retry: Options['retry'];
 	prefixUrl: Options['prefixUrl'];
 	onDownloadProgress: Options['onDownloadProgress'];
-
-	// New type.
-	headers: Headers;
 }
 
 /**


sindresorhus:
Sorry. I missed this. Looks good 👌

PR #244, opened 2020-03-09
Declare package as module and specify exports files

ubergesundheit:
Follow up to #242 after a discussion there.

This pull request adds the fields `type` and `exports` to `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,11 @@
 		"umd.js",
 		"umd.d.ts"
 	],
+	"type": "module",
+	"exports": {
+		"import": "./index.js",
+		"require": "./umd.js"
+	},
 	"keywords": [
 		"fetch",
 		"request",
--- a/package.json
+++ b/package.json
@@ -10,6 +10,11 @@
 		"email": "sindresorhus@gmail.com",
 		"url": "sindresorhus.com"
 	},
+	"type": "module",
+	"exports": {
+		"import": "./index.js",
+		"require": "./umd.js"
+	},
 	"engines": {
 		"node": ">=10"
 	},
@@ -24,11 +29,6 @@
 		"umd.js",
 		"umd.d.ts"
 	],
-	"type": "module",
-	"exports": {
-		"import": "./index.js",
-		"require": "./umd.js"
-	},
 	"keywords": [
 		"fetch",
 		"request",


ubergesundheit:
Mh could it be possible that declaring `"type": "module"` breaks this for all node versions below 13? Wdyt about just specifying the `exports` field.

sindresorhus:
Why would it break? `type` is a new property for Node.js 13.

ubergesundheit:
Tests with node 12 and 10 fail with error `AVA cannot yet load ESM files.`

sholladay:
Tests are also failing for me locally with Node v13.8.0. Lots of errors like...

```
Uncaught exception in test/browser.js


file:///private/tmp/ky/test/_require.js:1
import fetch, {Headers, Request, Response} from 'node-fetch';
               ^^^^^^^
SyntaxError: The requested module 'node-fetch' does not provide an export named 'Headers'
```

sindresorhus:
@ubergesundheit Bump :)

ubergesundheit:
I dont have the time right now to push this further. Feel free to pick up or close. Sorry :(

sholladay:
I played with it a bit and couldn't figure it out. Seems like it should work, but `"type": "module"` changes something in the tooling that causes imports within the test helper to fail. Probably something in AVA or the `esm` package.

sindresorhus:
Slightly related, but Webpack 5 will finally support the `exports` field: https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.18

rifler:
> Tests are also failing for me locally with Node v13.8.0. Lots of errors like...
> 
> ```
> Uncaught exception in test/browser.js
> 
> 
> file:///private/tmp/ky/test/_require.js:1
> import fetch, {Headers, Request, Response} from 'node-fetch';
>                ^^^^^^^
> SyntaxError: The requested module 'node-fetch' does not provide an export named 'Headers'
> ```

Maybe the thing is that node cannot do named imports from CJS package, you have to write

```
import fetch from 'node-fetch';

const { Headers, Request, Response } = fetch;
```

you can try 14.13.0, where support for named import was added

https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#2020-09-29-version-14130-current-mylesborins

sindresorhus:
Closing for lack of activity. Continued in https://github.com/sindresorhus/ky/issues/283.

ubergesundheit:
Sorry for the lack of activity :( 

I'll try to read the linked issue . If there is anything I could do, LMK

sholladay:
@ubergesundheit could you try the solution suggested in https://github.com/sindresorhus/ky/pull/244#issuecomment-703902208? Seems promising.

Issue #248, opened 2020-04-08
An undefined HTTPError message breaks New Relic


kumar303:
An app I work on uses [New Relic browser monitoring](https://newrelic.com/products/browser-monitoring) for errors and in the case where `ky` throws an `HTTPError` with an undefined message, New Relic breaks like this:

<img width="348" alt="Screen Shot 2020-04-08 at 10 49 59 AM" src="https://user-images.githubusercontent.com/55398/78824816-946d2600-79a4-11ea-9ea4-2f2701db1493.png">

The top line shows 215 `ky` `HTTPError` instances but you can't see what they are and you can't click on them. I used the stack traces to confirm they are `ky.HTTPError()` instances. 

This is most certainly a New Relic bug but I think it would be nice if `ky` always defined an `HTTPError` message.

Why would an `HTTPError` message be undefined? I couldn't reproduce it exactly but I think it's because browsers like Safari, IE, and others do not define [`response.statusText`](https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText#Browser_compatibility).

kumar303:
I proposed a patch which should fix it: https://github.com/sindresorhus/ky/pull/249

sholladay:
Could some of these requests be using HTTP/2?

Look how much of a mess this is...
https://github.com/whatwg/fetch/issues/599

kumar303:
Ah, possibly. In my error metrics I see that they are coming from recent browser versions (Safari, Chrome). I still cannot reproduce them myself when using the same browser versions, though.

PR #249, opened 2020-04-08
Ensure HTTPError always has a defined message

kumar303:
Fixes https://github.com/sindresorhus/ky/issues/248

This ensures HTTPError will always have a defined message in various situations.

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,7 +118,7 @@ const stop = Symbol('stop');
 
 class HTTPError extends Error {
 	constructor(response) {
-		super(response.statusText);
+		super(response.statusText || response.status || 'Unknown response error');
 		this.name = 'HTTPError';
 		this.response = response;
 	}
--- a/index.js
+++ b/index.js
@@ -118,10 +118,14 @@ const stop = Symbol('stop');
 
 class HTTPError extends Error {
 	constructor(response) {
+		// Set the message to the status text, such as Unauthorized,
+		// with some fallbacks. This message should never be undefined.
 		super(
 			response.statusText ||
-			(response.status && String(response.status)) ||
-			'Unknown response error'
+			String(
+				response.status === 0 || response.status ?
+					response.status : 'Unknown response error'
+			)
 		);
 		this.name = 'HTTPError';
 		this.response = response;
--- a/index.js
+++ b/index.js
@@ -123,7 +123,7 @@ class HTTPError extends Error {
 		super(
 			response.statusText ||
 			String(
-				response.status === 0 || response.status ?
+				(response.status === 0 || response.status) ?
 					response.status : 'Unknown response error'
 			)
 		);
--- a/index.js
+++ b/index.js
@@ -118,7 +118,11 @@ const stop = Symbol('stop');
 
 class HTTPError extends Error {
 	constructor(response) {
-		super(response.statusText || response.status || 'Unknown response error');
+		super(
+			response.statusText ||
+			(response.status && String(response.status)) ||
+			'Unknown response error'
+		);
 		this.name = 'HTTPError';
 		this.response = response;
 	}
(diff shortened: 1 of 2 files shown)


yunyu:
This would also help with Sentry error reporting.

index.js
```diff
@@ -118,7 +118,7 @@ const stop = Symbol('stop');
 
 class HTTPError extends Error {
 	constructor(response) {
-		super(response.statusText);
+		super(response.statusText || response.status || 'Unknown response error');
```
sindresorhus, line 121: `.status` is a number, right? If so, it should be coerced to a string.

sholladay: The Error constructor handles coercion. But maybe TypeScript doesn't account for that?

sindresorhus: So why didn’t it handle coercion from undefined?

sholladay: Good question. I'm not entirely sure. Normally, `Error` coerces `undefined` to an empty string.

Maybe New Relic overwrites the global `Error` constructor to make it more strict? Hopefully, it would still accept numbers, though.

kumar303: > `.status` is a number, right? If so, it should be coerced to a string.

I can add string coercion. 

> So why didn’t it handle coercion from undefined?

Throwing `new Error()` with an argument would also result in an undefined message so it's not an invalid call.

> Maybe New Relic overwrites the global `Error` constructor to make it more strict? 

I think it's just a New Relic bug. It should be handling errors with undefined message properties.

kumar303:
Thanks for the quick review!

> I like the implementation but for the tests, I would prefer to use more realistic objects and patterns that align better with the other existing tests. For example, use an actual test server that responds without status text. 

Yeah, I struggled with this a bit. I couldn't see an easy way to make a response with an undefined `statusText` using the test server (it doesn't provide low enough access to the response).

> Or instead of `http.ServerResponse`, return a WHATWG `Response` inside of a hook.

Ah, yes, a WHATWG `Response` would be more realistic. I'll try that.

kumar303:
@sholladay I switched to a proper fetch `Response`. Let me know how it looks.

index.js
```diff
@@ -118,7 +118,11 @@ const stop = Symbol('stop');
 
 class HTTPError extends Error {
 	constructor(response) {
-		super(response.statusText);
+		super(
+			response.statusText ||
+			(response.status && String(response.status)) ||
```
sholladay, line 123: I think we should handle the case where `response.status` is `0`. I realize it is non-standard, but I found a few search results of real-world cases where people receive `0` as a status code.

Example: CloudMailin uses status code 0 for "Unreachable"

https://docs.cloudmailin.com/receiving_email/http_status_codes/

One way to solve this is to use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator).

```js
super(response.statusText || String(response.status ?? 'Unknown response error'));
```

kumar303: Sounds like a good idea, thanks. I'll also add a test for it.

kumar303:
@sholladay this supports status codes of 0 now.

test/http-error.js
```diff
@@ -0,0 +1,46 @@
...
+import ky from '..';
+
+function createFakeResponse({status, statusText}) {
+	// Start with a realistic fetch Response.
+	const response = {...new Response()};
+
+	// Assign these values because the Response() constructor doesn't
+	// support setting them to undefined.
+	response.status = status;
+	response.statusText = statusText;
```
sholladay, line 12: These can be set in the `init` argument for the `Response` constructor. Should avoid the need to spread the response into a new object, too.

kumar303: It doesn't work because these values need to be set to `undefined` and from looking at the `node-fetch` source, they are using old school truthy checks to tell if these values are set (thus, they can never be set to `undefined`).

I realize it makes the code look a bit odd -- that's why I added a comment explaining it.

kumar303:
@sholladay let me know if my response makes sense. Thanks.

Issue #254, opened 2020-05-13
Support custom JSON parser


kirillgroshkov:
In our codebase (that we're migrating to ky) we use a "JSON reviver" (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse), which works like this:

```typescript
const parsed = JSON.parse(myJson, myReviver)
```

I don't see that ky supports it.

I did a workaround (that looks ugly) now via afterResponseHook:

```typescript
const afterResponseHook: AfterResponseHook = async (_req, _opt, res) => {
  if (!res.ok) return   
  // Parse with JSON Reviver (if possible)   
  const data = JSON.parse(await res.text(), backendResponseReviver)

  // Weird, but pack the "revived" json object back to string, so it can be parsed without reviver then :)
  return new Response(JSON.stringify(data, backendResponseReviver), res)
}
```

Nicer API would be (for example; not certain about the naming of `json()` options):
```typescript
const response = await ky.get(url).json({ reviver: myReviver })
```

What do you think?

I can try to submit a PR if you think it's a feasible approach.

sholladay:
I think I'm slightly against having a `reviver` option, as it's kind of a niche use case and it _is_ possible to accomplish your end goal with a hook already, as you demonstrated. Maybe we could generalize this a bit and have a `parseJson` function option. That way you could use a reviver, but you could also use an alternative parsing implementation, such as [bourne](https://github.com/hapijs/bourne), which useful in some cases.

kirillgroshkov:
Agree that `bourne` is a good use-case.

With my hook implementation there's one discrepancy though that broke our code in one place. "json-wise" the result is the same, but we had a reviver that constructs a certain class (called `Price` in our case) and we had a code that tests for `a instanceof Price` that failed now (if reviver wasn't used). We'll change the code (in this specific case), but still a limitation.

Agree that reviver is a niche-case (I'd prefer us not to have them, but it's out of control at the moment).

So, `parseJson` option should solve all cases I believe!

kirillgroshkov:
Off topic here, but same `parseJson` option would be nice to have in `got` for same reasons (bourne/revivers). If you agree - I can open issue in `got` too

sindresorhus:
I'm ok with a `parseJson` option. The docs should mention the two use-cases for it mentioned here.

sindresorhus:
[...]

PR #257, opened 2020-05-17
Add `parseJson` option

kirillgroshkov:
fixes https://github.com/sindresorhus/ky/issues/254

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -209,10 +209,9 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 	/**
 	User-defined JSON-parsing function.
 
-	Useful e.g in these cases:
-
-	1. Parse JSON via [bourne](https://github.com/hapijs/bourne) to protect from prototype pollution.
-	2. Parse JSON with [reviver](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).
+	Use-cases:
+	1. Parse JSON via the [`bourne` package](https://github.com/hapijs/bourne) to protect from prototype pollution.
+	2. Parse JSON with [`reviver` option of `JSON.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).
 
 	@default JSON.parse()
 
@@ -228,7 +227,6 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 	})();
 	```
 	*/
-
 	parseJson?: (text: string) => unknown
 
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -207,6 +207,31 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 	json?: unknown;
 
 	/**
+	User-defined JSON-parsing function.
+
+	Useful e.g in these cases:
+
+	1. Parse JSON via [bourne](https://github.com/hapijs/bourne) to protect from prototype pollution.
+	2. Parse JSON with [reviver](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).
+
+	@default JSON.parse()
+
+	@example
+	```
+	import ky from 'ky';
+	import bourne from '@hapijs/bourne';
+
+	(async () => {
+		const parsed = await ky('https://example.com', {
+			parseJson: text => bourne(text)
+		}).json();
+	})();
+	```
+	*/
+
+	parseJson?: (text: string) => unknown
+
+	/**
 	Search parameters to include in the request URL. Setting this will override all existing search parameters in the input URL.
 
 	Accepts any value supported by [`URLSearchParams()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams).
--- a/readme.md
+++ b/readme.md
@@ -348,12 +348,14 @@ import ky from 'ky';
 
 ##### parseJson
 
-Type: `Function`
+Type: `Function`\
+Default: `JSON.parse()`
 
-User-defined function to handle JSON parsing. Useful e.g in these cases:
+User-defined JSON-parsing function.
 
-1. Parse JSON via [bourne](https://github.com/hapijs/bourne) to protect from prototype pollution.
-2. Parse JSON with [reviver](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).
+Use-cases:
+1. Parse JSON via the [`bourne` package](https://github.com/hapijs/bourne) to protect from prototype pollution.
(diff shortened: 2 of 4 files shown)


readme.md
```diff
@@ -346,6 +346,26 @@ import ky from 'ky';
...
+1. Parse JSON via [bourne](https://github.com/hapijs/bourne) to protect from prototype pollution.
+2. Parse JSON with [reviver](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).
+
+```js
+import ky from 'ky';
+import bourne from '@hapijs/bourne'
+
+(async () => {
+	const parsed = await ky('https://example.com', {
+        parseJson: text => bourne(text)
```
sholladay, line 364: Indentation here should use tabs

index.d.ts
```diff
@@ -27,6 +27,8 @@ export type AfterResponseHook = (
 	response: Response,
 ) => Response | void | Promise<Response | void>;
 
+export type ParseJson = <T = unknown>(text: string) => T
```
sindresorhus, line 30: I don't think it needs an exported type. Can just put the type inline.

index.d.ts
```diff
@@ -206,6 +208,19 @@ export interface Options extends Omit<RequestInit, 'headers'> {
...
 
+	/**
+	User-defined JSON-parsing function.
+
+	Useful e.g in these cases:
+
+	1. Parse JSON via [bourne](https://github.com/hapijs/bourne) to protect from prototype pollution.
+	2. Parse JSON with [reviver](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).
+
+	@default JSON.parse()
```
sindresorhus, line 219: Missing the example from the readme. Readme and index.d.ts should be in sync.

index.d.ts
sholladay: I'm not sure I fully understand the use of the type variable `T` here. What's the benefit over just `unknown`?

kirillgroshkov: > I'm not sure I fully understand the use of the type variable `T` here. What's the benefit over just `unknown`?

Fair, will make it just `unknown`

readme.md
kirillgroshkov: fixed

index.d.ts
kirillgroshkov: fixed

index.d.ts
kirillgroshkov: fixed

sholladay:
Do we want to clarify that this only works if you call `.json()` on the promise, using the `ky().json()` shortcut? It doesn't work if you call `.json()` on the response after awaiting the promise, like this:

```js
const response = await ky('./foo');
const parsed = await response.json();
```

I think it's fine that it only works using the shortcut, just wondering whether it will be obvious enough. Some people seem to get slightly confused about the body method shortcuts vs their long form.

kirillgroshkov:
> I think it's fine that it only works using the shortcut, just wondering whether it will be obvious enough. Some people seem to get slightly confused about the body method shortcuts vs their long form.

I didn't even think about the `body.json()` usage. It's not mentioned anywhere in the documentation, and using shortcut methods like `.json()` or `.text()`, in my opinion, is one of the first (very straightforward) selling points of `ky`.

Also, as a typescript user, I usually explore the options via typescript declarations, and I won't even try to use `parseJson` option on a method that doesn't have it.

Having said that, better be obvious than unclear - I'll add a short comment about it in readme now.

kirillgroshkov:
Wait, now I understand what you meant.

I actually forgot that I implemented it as an option to the `ky` method itself, not `.json(parseJson: ...)`. So yes, your comment is more relevant now.

Adding readme coment.

kirillgroshkov:
@sholladay @sindresorhus please review again

readme.md
```diff
@@ -346,6 +346,36 @@ import ky from 'ky';
...
+
+(async () => {
+	const parsed = await ky('https://example.com', {
+		parseJson: text => bourne(text)
+	}).json();
+})();
+```
+
+Please note that this option only works when using it like `ky(…).json()` (like in the example above).
+It won't work if you first get the response object and then call `.json()` on it:
```
sindresorhus, line 372: Why can't we make this work?

sholladay: @sindresorhus unlike the promise that Ky returns and its `promise.json()` method, we don't construct the response object, as it's the return value from `fetch()`. As far as I know, we can't easily change the behavior of its `response.json()`. I suppose we may be able to monkey patch it. How would you like to approach that?

kirillgroshkov: I think I'm with @sholladay on that. I think that monkey-patching the native `fetch.json` is a bit too far to go.

I don't have the stats of how often people use ky's json vs ky.response.json, but IMO the big selling point of ky is (among other things) these shorthand methods. There should be strong reasons to not use them AND also use parseJson option in the same time.

sindresorhus: We could just create a `Proxy` wrapper.

sholladay: Doesn't seem to work, even with an empty / transparent proxy (tried in Firefox 76):

```js
const response = new Response(JSON.stringify({foo:'bar'}));
const proxy = new Proxy(response, {});
proxy.json().then(console.log);
// TypeError: 'json' called on an object that does not implement interface Response.
```

sindresorhus: I think you actually have to implement the `apply` hook and use `Reflect.apply` correctly.

sindresorhus:
Bump :)

readme.md
sholladay: Looking at this with a fresh pair of eyes, I think it may be sufficient to overwrite the `json` method of the `response` object on this line: https://github.com/sindresorhus/ky/blob/2c5d6dda8ac889048408a337c3f5d068769981b2/index.js#L330

We would do something like this...

```js
if (this._options.parseJson) {
    response.json = this._options.parseJson
}
return response;
```

Earlier, I was thinking that it might not be enough due to usage of `response.clone()`, as the custom `json` function would not be cloned. However, after looking at the places we use response cloning, I don't think it's actually a problem.

sindresorhus: Right. Let's try that. 👍

^ @kirillgroshkov

sholladay:
I implemented the support for `response.json()` (as opposed to `promise.json()`).

Still two issues:
1. Docs need to be updated to reflect that there's no longer a limitation.
2. This PR is suffering from the same glitch that #261 did, wherein pushes soft fail (an error is logged but it actually succeeds) and the Travis run fails on Node 10. One of the weirdest glitches I've seen in a while. I was able to solve it by creating a new branch and PR. @kirillgroshkov ideally you should do that so that it's clear you were the author.

kirillgroshkov:
> 1. Docs need to be updated to reflect that there's no longer a limitation.
> 2. This PR is suffering from the same glitch that #261 did, wherein pushes soft fail (an error is logged but it actually succeeds) and the Travis run fails on Node 10. One of the weirdest glitches I've seen in a while. I was able to solve it by creating a new branch and PR. @kirillgroshkov ideally you should do that so that it's clear you were the author.

ok, will do!

PR #261, opened 2020-06-20
Default to empty object instead of undefined for headers param

### LINKED ISSUE #260: Edge <= 18 Invalid argument exception
Pre-Chromium Microsoft Edge blows up when `undefined` is passed as an argument to the `window.Headers` constructor. As such, ky blows up when trying to make a request due to [these lines](https://github.com/sindresorhus/ky/blob/2c5d6dda8ac889048408a337c3f5d068769981b2/index.js#L52).

JosNun:
This fixes #260 by giving the arguments to the `mergeHeaders` function default params so `undefined` never gets passed to the `Headers` constructor.

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,6 +46,7 @@ for (const property of globalProperties) {
 const isObject = value => value !== null && typeof value === 'object';
 const supportsAbortController = typeof globals.AbortController === 'function';
 const supportsStreams = typeof globals.ReadableStream === 'function';
+const supportsFormData = typeof globals.FormData === 'function';
 
 const deepMerge = (...sources) => {
 	let returnValue = {};
@@ -240,6 +241,12 @@ class Ky {
 		if (this._options.searchParams) {
 			const url = new URL(this.request.url);
 			url.search = new URLSearchParams(this._options.searchParams);
+
+			// To provide correct form boundary, Content-Type header should be deleted each time when new Request instantiated from another one
+			if (((supportsFormData && this._options.body instanceof globals.FormData) || this._options.body instanceof URLSearchParams) && !(this._options.headers && this._options.headers['content-type'])) {
+				this.request.headers.delete('content-type');
+			}
+
 			this.request = new globals.Request(new globals.Request(url, this.request), this._options);
 		}
 
--- a/index.js
+++ b/index.js
@@ -54,7 +54,7 @@ const mergeHeaders = (source1, source2) => {
 	const source = new globals.Headers(source2 || {});
 
 	for (const [key, value] of source) {
-		if (isHeadersInstance && (value === undefined || value === 'undefined')) {
+		if ((isHeadersInstance && value === 'undefined') || value === undefined) {
 			result.delete(key);
 		} else {
 			result.set(key, value);
--- a/index.js
+++ b/index.js
@@ -49,12 +49,12 @@ const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
 const mergeHeaders = (source1, source2) => {
-	const result = new globals.Headers(source1);
+	const result = new globals.Headers(source1 || {});
 	const isHeadersInstance = source2 instanceof globals.Headers;
-	const source = new globals.Headers(source2);
+	const source = new globals.Headers(source2 || {});
 
 	for (const [key, value] of source) {
-		if ((isHeadersInstance && value === 'undefined') || value === undefined) {
+		if (isHeadersInstance && (value === undefined || value === 'undefined')) {
 			result.delete(key);
 		} else {
 			result.set(key, value);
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -48,7 +48,7 @@ const supportsAbortController = typeof globals.AbortController === 'function';
 const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
-const mergeHeaders = (source1, source2) => {
+const mergeHeaders = (source1 = {}, source2 = {}) => {
```
Jensen95, line 51: This would work with undefined perfectly with undefined, but should null be passed as an argument it will not get the default value. And as you wrote in https://github.com/sindresorhus/ky/issues/260#issuecomment-646725006 it will blow up in the Browser. So the best solution would properly be like suggested in https://github.com/sindresorhus/ky/issues/260#issuecomment-646718654 with a ternary check

sholladay:
We'll need a test for this, otherwise we'll undoubtedly break it by accident in the future. Might want to hold off on that until Sindre chimes in, though. I'm personally fine with this if it's such a small change. The test would need to override the `Headers` class to match the behavior of Edge.

sindresorhus:
I'm fine with this if it has a test.

belgattitude:
@sindresorhus @JosNun  this hit us quite badly, any way I can help to release this ?

index.js
```diff
@@ -49,12 +49,12 @@ const supportsStreams = typeof globals.ReadableStream === 'function';
...
 const mergeHeaders = (source1, source2) => {
-	const result = new globals.Headers(source1);
+	const result = new globals.Headers(source1 || {});
 	const isHeadersInstance = source2 instanceof globals.Headers;
-	const source = new globals.Headers(source2);
+	const source = new globals.Headers(source2 || {});
 
 	for (const [key, value] of source) {
-		if ((isHeadersInstance && value === 'undefined') || value === undefined) {
+		if ((isHeadersInstance && (value === undefined || value === 'undefined')) {
```
belgattitude, line 57: The expression lacks a parenthesis

sholladay: Thank you. Fixed.

sholladay:
@belgattitude the best way to help would be to write a test that fails on the `master` branch but passes with the changes in this PR. If no one gets around to it, I plan to take care of it sometime this week. But that would definitely speed things up. We can get this released quickly once there is a test.

belgattitude:
@JosNun I proposed I different approach here: #264 feel free to comment

JosNun:
I can't get around to writing a test until at least Sunday.

sindresorhus:
Tests are failing on Node.js 10: https://travis-ci.com/github/sindresorhus/ky/jobs/359302555

sholladay:
Some weird things:
1. In CI, the tests only fail on Node 10 and they do so consistently, I tried restarting the build but it didn't help.
2. It's the browser tests that are failing, not the headers tests. We didn't touch the browser tests in this PR.
3. The failure is due to a `SyntaxError` that points at Puppeteer, but it looks like Puppeteer is itself tested on Node 10. Granted, I haven't looked into their test coverage and their test setup is a little complex, it's possible something slipped by, but based on their [Travis config](https://github.com/puppeteer/puppeteer/blob/790c277f0ebdf2396b616d0ba823cd3416013787/.travis.yml), they seem to do plenty of testing on Node 10. Additionally, the line that is referenced in the error looks syntactically valid to me. Why isn't it showing the problematic identifier or column number??
3. I can't exactly reproduce the test failure. Locally, with Node 10 **and** 14, `npx ava test/browser.js` mostly passes but the "onDownloadProgress works" test just hangs. Interestingly, in that test, I tried changing the server response headers to `transfer-encoding: chunked` instead of `content-length: 4` and it actually solved the hanging problem, but of course then the test assertions failed because that causes Ky to lack the necessary data for the `percent` and `totalBytes` values. This leads me to believe that the problem is not actually in Puppeteer. Or that there are multiple problems happening at once. Very strange.

sholladay:
LOL I removed everything except for a single whitespace change in the headers tests, and the browser tests still fail on Node 10. It fails consistently. Yet, the CI run that happened only a few hours ago for the v0.21 release succeeded (https://travis-ci.com/github/sindresorhus/ky/jobs/360237701).

I would be at a total loss here, except there's one clue: when I've been pushing to this branch, I've been receiving some weird `remote rejected` errors on the command line. But the commits do actually get pushed and appear in the GitHub UI. It's a stab in the dark, but I am thinking there may actually be some kind of data corruption issue on GitHub's backend on this branch.

I'm going to make a new branch from scratch with the same changes and see if it works. Will be kinda sad if this turns out to be the case. Hopefully it fails and there's a more mundane reason.

sholladay:
Well, PR #270 passed in CI. It's the same code, copy pasted. Fun times. Pretty sure this has to be GitHub/Travis related, but unclear exactly how. My guess is some kind of buggy caching mechanism. "It's always a cache bug".

sholladay:
Looks like this has been happening since at least May. Here's the same issue on a different branch from a different PR and author:

https://github.com/sindresorhus/ky/runs/701925183

belgattitude:
@sholladay thank you so much ! I could not find time to help. But really appreciate.

With 0.21 my test suite is actually failing. I'm looking right now to see why...

PR #268, opened 2020-07-03
reproduce mergeHeaders typeerror

belgattitude:
**DO NOT MERGE, just a test, just downtracking**

Related to #264.

- [x] Reproduce type error failure on master branch -> https://github.com/sindresorhus/ky/pull/268/commits/4cbc981ad21bd1aef96e2eded9b2aefcdb2ddbe2 (breaks as expected)
- [x] Testing a quick fix -> https://github.com/sindresorhus/ky/pull/268/commits/7def7f338061608e488ea76876a141558cdb4989 (pass as expected)

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,6 +49,10 @@ const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
 const mergeHeaders = (source1, source2) => {
+	if (!isObject(source1)) {
+		throw new TypeError('The `source1` argument must be an object');
+	}
+
 	const result = new globals.Headers(source1);
 	const isHeadersInstance = source2 instanceof globals.Headers;
 	const source = new globals.Headers(source2);
--- a/index.js
+++ b/index.js
@@ -228,11 +228,12 @@ class Ky {
 	constructor(input, options = {}) {
 		this._retryCount = 0;
 		this._input = input;
+		const headers = isObject(this._input.headers) ? this._input.headers : {};
 		this._options = {
 			// TODO: credentials can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
 			credentials: this._input.credentials || 'same-origin',
 			...options,
-			headers: mergeHeaders(this._input.headers, options.headers),
+			headers: mergeHeaders(headers, options.headers),
 			hooks: deepMerge({
 				beforeRequest: [],
 				beforeRetry: [],


belgattitude:
@sholladay trying to fix the #264 in the constructor rather than in the mergeHeaders(). 

So here's the reproduction case when doing typecheck in the mergeHeaders (most browsers will assume undefined as {}, but this reproduce the case, looks a legit to me).

Fix is not not a problem. 

Test for me is hard to do. Not used to ava, no constructor exposed. I appreciate your guidance

What I try to achieve:

- constructor
  - Given that `input`is a string (ie: '/api/test')
     - the headers are taken from argument `options.headers`
  - Given than input is an URL or `global.Request` (and have headers defined).
    - the headers must be merged with arguments in `options.headers` 

There's no reason why we would want to throw a type error when input is a string (it's part of the contract)

An *exhaustive* version of the constructor would reflect the idea +/- like this


```js
class Ky {
  constructor(input, options = {}) {
    // early return
   if (typeof input !== 'string' && !(input instanceof URL || input instanceof globals.Request)) {
     throw new TypeError('`input` must be a string, URL, or Request');
   }
   let headers;
   if (isObject(input) && isObject(input.headers)) {
     headers = mergeHeaders(input.headers, options.headers);
   } else {
     headers = isObject(options.headers) ? options.headers : {};
   }
   /// here a lot more
   this._options = {
     headers
   }
```

The alternative approach still exists in my mind: setting default {} in `mergeHeaders` when the type is wrong (actually what's happening in modern js, I think).

index.js
```diff
@@ -49,6 +49,10 @@ const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
 const mergeHeaders = (source1, source2) => {
+	if (!isObject(source1)) {
```
belgattitude, line 52: cause yes mergeHeaders will fail in edge 18 (maybe others too)...

index.js
```diff
@@ -224,11 +228,12 @@ class Ky {
 	constructor(input, options = {}) {
 		this._retryCount = 0;
 		this._input = input;
+		const headers = isObject(this._input.headers) ? this._input.headers : {};
```
belgattitude, line 231: This can be simplified ! just checking what's the practice here.

Possible 
```
  this._options = {
    headers: mergeHeaders(this._input.headers || {}, options.headers);
  }
```

belgattitude:
Closing as just it was included in #264

PR #270, opened 2020-07-12
Fix headers construction in old browsers

### LINKED ISSUE #260: Edge <= 18 Invalid argument exception

sholladay:
Closes #260 

_This is a redo of PR #261, because the tests on that PR are exhibiting some unexpected behavior that I haven't been able to determine the root cause of, but are most likely git related, possibly even a backend issue with GitHub._

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,9 +49,9 @@ const supportsStreams = typeof globals.ReadableStream === 'function';
 const supportsFormData = typeof globals.FormData === 'function';
 
 const mergeHeaders = (source1, source2) => {
-	const result = new globals.Headers(source1);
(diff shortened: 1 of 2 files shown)


belgattitude:
Looks good to me ! Thanks a lot @sholladay. Very appreciated

Issue #149, opened 2019-07-07
Handle errors in afterResponse hook


gutenye:
The API returns `{ status: 'error', message: 'foo' }` for errors, and I want to handle it in `afterResponse` hook, here's the code.

```js
const api = ky.create({
  prefixUrl: 'http://localhost',
  hooks: {
    afterResponse: [
      async res => {
        const body = await res.json()
        if (body.status === 'error') {
          throw new Error(body.message)
        }
      },
    ]
  }
})
const users = await api.get('users').json()
```

The problem is it sends two requests to the server.
```
GET http://localhost/users
GET http://localhost/users
```

Expect it only send one request to the server. (Because of retry)

whitecrownclown:
@gutenye 	That's because `ky` has a default `retry` [value](https://github.com/sindresorhus/ky#retry).

[codesandbox](https://codesandbox.io/s/stupefied-matsumoto-rbelu)

You could do

```javascript
const users = api.get('users', {retry: 0}).json();
```

sholladay:
Yep, this is behaving as expected. By throwing an error, you have told Ky that the request failed, so it tries again. I concur with @whitecrownclown's suggestion to disable retries, if that is what you desire.

gutenye:
```
Retry failed requests made with one of the below methods that result in a network error or one of the below status codes.

Methods: GET PUT HEAD DELETE OPTIONS TRACE
Status codes: 408 413 429 500 502 503 504
```

The doc said it retries in a network error. But I want a central place to handle the API error (not network error) while keeping the retry feature, is there a way to do it?

sholladay:
Well, I would argue that there are different categories of errors going on here, and thus they need to be handled differently. There are fatal errors, like invalid syntax, and then there are operational errors, like a rate limit being reached. The latter should be retried. It would be significantly easier if the API just responded with the appropriate HTTP status codes, so that Ky could handle these errors automatically. However, you can simulate that behavior by doing `return new Response()` in the hook and mapping the API's fatal errors to the appropriate status code. That will tell Ky not to retry, as instructed by the HTTP specification.

Or you could simply throw operational errors in the hook and leave the rest to handle after the `ky.get()` call.

gutenye:
For anyone comes into this problem, I use `throw new ky.HTTPError({ statusText: body.error_message })` in the hooks with retry enabled. Thanks for the helps.

sindresorhus:
[...]

PR #271, opened 2020-07-15
Add support for relative URLs and searchParams even in non-browsers

### LINKED ISSUE #236: TypeError invalid URL when using relative URL and searchParams
hello!
when using ky with `@nuxt/http` but encountered an error.
```js
    const response = await $http.$get('list', {
      prefixUrl: '/api,
      searchParams: {
        page: 1,
      },
    });
```
the code above triggers an error:
```
UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_URL]: Invalid URL: /api/list
    at onParseError (internal/url.js:243:9)
    at new URL (internal/url.js:319:5)
    at new Ky (node_modules/ky/umd.js:246:0)
```

and then i check `node_modules/ky/umd.js` and [MDN URL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)
```js
if (this._options.searchParams) {
  const url = new URL(this.request.url);
  url.search = new URLSearchParams(this._options.searchParams);
}
```

sholladay:
Closes #236 

This PR aims to solve the last problem I am aware of related to URL resolution in Ky. Historically, we've used the WHATWG URL constructor (`new URL()`) in various situations to parse, resolve, and modify URLs. But it's been highly problematic because that API only supports absolute URLs and throws an error if given anything else. It's difficult for Ky to know what base URL to resolve against just to satisfy that constraint, even in browsers (e.g. web workers do not have access to `document.baseURI`), not to mention Deno, Node, React Native, and other environments. This makes `new URL()` quite a footgun.

I have chipped away at these kinds of URL resolution problems over time (e.g. #59 and #180). These days, we mostly rely on the much better behaved `Request` class which _does_ support relative URLs, however it doesn't cover all the same use cases. There was one remaining place we still used `new URL()`: when the user provides `searchParams` to modify the `input` URL. This is the kind of thing `new URL()` is perfectly suited for... unfortunately, as mentioned, it's just too finicky. Hopefully, someday they'll add support for relative URLs. Until then, we'll do our own string replacement.

RIP to the WHATWG URL API. ⚰️

### DIFF
--- a/index.js
+++ b/index.js
@@ -272,8 +272,8 @@ class Ky {
 		this.request = new globals.Request(this._input, this._options);
 
 		if (this._options.searchParams) {
-			const url = new URL(this.request.url);
-			url.search = new URLSearchParams(this._options.searchParams);
+			const searchParams = '?' + new URLSearchParams(this._options.searchParams).toString();
+			const url = this.request.url.replace(/(?:\?.*?)?(?=#|$)/, searchParams);
 
 			// To provide correct form boundary, Content-Type header should be deleted each time when new Request instantiated from another one
 			if (((supportsFormData && this._options.body instanceof globals.FormData) || this._options.body instanceof URLSearchParams) && !(this._options.headers && this._options.headers['content-type'])) {
--- a/package.json
+++ b/package.json
@@ -58,9 +58,9 @@
 		"form-data": "^3.0.0",
 		"node-fetch": "^2.5.0",
 		"nyc": "^15.0.0",
-		"puppeteer": "^3.0.4",
+		"puppeteer": "^5.1.0",
 		"rollup": "^2.10.2",
-		"tsd": "^0.11.0",
+		"tsd": "^0.13.1",
 		"xo": "^0.25.3"
 	},
 	"sideEffects": false,
--- a/test/browser.js
+++ b/test/browser.js
@@ -302,7 +302,7 @@ test('retry with body', withPage, async (t, page) => {
 	});
 	server.put('/test', async (request, response) => {
 		requestCount++;
-		t.is(request.body, 'foo');
+		t.is(await pBody(request), 'foo');
 		response.sendStatus(502);
 	});
 
--- a/test/browser.js
+++ b/test/browser.js
@@ -9,6 +9,7 @@ const pBody = util.promisify(body);
 
 test('prefixUrl option', withPage, async (t, page) => {
 	const server = await createTestServer();
+
 	server.get('/', (request, response) => {
 		response.end('zebra');
 	});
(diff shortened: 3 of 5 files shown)


index.js
```diff
@@ -272,8 +272,8 @@ class Ky {
 		this.request = new globals.Request(this._input, this._options);
 
 		if (this._options.searchParams) {
-			const url = new URL(this.request.url);
-			url.search = new URLSearchParams(this._options.searchParams);
+			const searchParams = '?' + new URLSearchParams(this._options.searchParams).toString();
+			const url = this.request.url.replace(/(?:\?.*?)?(?=#|$)/, searchParams);
```
sindresorhus, line 276:
```suggestion
			const url = this.request.url.replace(/(?:\?.*?)(?=#|$)/, searchParams);
```

?

Because `.*` already means 0 or more matches.

sindresorhus:
Have you considered opening an issue over at https://github.com/whatwg/url ?

index.js
sholladay: That wouldn't work in the case where the URL is missing a query string and we want to add one.

The non-capturing group `(?:\?.*?)` matches an existing query string. But we don't know if the URL has one, so the group is optional.

I added tests to cover this, which should fail if you make the suggested change. The relevant assertion is:

```js
 t.is(await ky('/unicorn', {searchParams: {foo: 'bar'}}).text(), '/unicorn?foo=bar');
```

sholladay:
https://github.com/whatwg/url/issues/531

sindresorhus:
🙌🏻

index.js
```diff
@@ -272,8 +272,8 @@ class Ky {
 		this.request = new globals.Request(this._input, this._options);
 
 		if (this._options.searchParams) {
-			const url = new URL(this.request.url);
```
szmarczak, line 275: Sorry I'm late to the party. Does this change anything? `this.request.url` is an absolute URL anyway... I'm just asking because I don't like RegExps.

sholladay: I don't like regex either. `this.request.url` can be a relative URL in some environments, e.g. Node and Deno.

szmarczak: Then why not this?

```js
const requestUrl = this.request.url;
const searchIndex = requestUrl.indexOf('?');
const urlWithoutParameters = searchIndex === -1 ? requestUrl : requestUrl.slice(0, searchIndex);
const url = `${urlWithoutParameters}${searchParams.toString()}`;
```

sholladay: I believe your code would remove the hash fragment from a URL, unlike the `new URL()` code we had before.

Granted, the hash fragment shouldn't be sent to the server anyway, so that shouldn't really matter. But I figure we should be spec compliant and unobtrusive by only modifying the search params and nothing more. It's possible that a custom `fetch` function might care about the hash fragment, for example.

szmarczak: Hmm... Fair point. We could also `indexOf('#')` but that would add more complexity. The regex solution is much clearer.

index.js
szmarczak: What about defining the regex as a constant, so V8 doesn't have to create it every time?

index.js
sholladay: Yep, my thoughts exactly. I did add tests that assert we don't modify the hash fragment. So if someone comes up with a clean non-regex solution, go ahead and give it a shot. It'll be checked thoroughly by the tests.

asaneh:
#297 #149 #432 #412 #430 #

asaneh:
#238

sholladay:
Hi, @asaneh. Is there a reason you linked those issue numbers here? It's unclear without any context.

szmarczak:
Given his activity on other repos I suppose he's just a spammer.

PR #273, opened 2020-07-17
Add option for custom `fetch` implementation

### LINKED ISSUE #269: Can external "fetch dependency be injected to ky"
I would like to inject an enhanced fetch to ky (fetch is a dependency to ky).

In order to use with https://github.com/AxaGuilDEv/react-oidc/tree/master/packages/context-fetch in a "functionnal way"

Is it possible to do something like : 

`
const parsed = await ky(fetch).post('https://example.com', {json: {foo: true}}).json();
`

?

Regards,

Guillaume

Skayo:
As requested in #269, this PR adds the ability to pass a `fetch` option when creating a ky instance.
This allows to use custom fetch implementations like [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic-unfetch) or when using SSR in some frameworks (Sapper in my case).

## TODO
- [X] Implementation
- [x] Documentation
- [x] Tests

Fixes #269

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -333,6 +333,26 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 	*/
 	onDownloadProgress?: (progress: DownloadProgress, chunk: Uint8Array) => void;
 
+	/**
+	User-defined fetch function.
+
+	Use-cases:
+	1. Use custom `fetch` implementations like [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic-unfetch).
+	2. Use the `fetch` wrapper function provided by some frameworks that use server-side rendering (SSR).
+
+	 @default fetch()
+
+	```js
+	import ky from 'ky';
+	import fetch from 'isomorphic-unfetch';
+
+	(async () => {
+		const parsed = await ky('https://example.com', {
+			fetch: fetch
+		}).json();
+	})();
+	```
+	 */
 	fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -334,15 +334,16 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 	onDownloadProgress?: (progress: DownloadProgress, chunk: Uint8Array) => void;
 
 	/**
-	User-defined fetch function.
+	User-defined `fetch` function.
 
 	Use-cases:
-	1. Use custom `fetch` implementations like [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic-unfetch).
+	1. Use custom `fetch` implementations like [`isomorphic-unfetch`](https://www.npmjs.com/package/isomorphic-unfetch).
 	2. Use the `fetch` wrapper function provided by some frameworks that use server-side rendering (SSR).
 
-	 @default fetch()
+	@default fetch
 
-	```js
+	@example
+	```
 	import ky from 'ky';
 	import fetch from 'isomorphic-unfetch';
 
@@ -352,7 +353,7 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 		}).json();
 	})();
 	```
-	 */
+	*/
 	fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -335,6 +335,7 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 
 	/**
 	User-defined `fetch` function.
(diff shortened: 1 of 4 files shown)


Skayo:
Definitely need some help with the tests as I don't have any experience with ava or whatever you're using here

sindresorhus:
> Definitely need some help with the tests as I don't have any experience with ava or whatever you're using here

Give it a try and we'll be happy to help if you get stuck.

https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md

sholladay:
The tests for this feature should go in here:
https://github.com/sindresorhus/ky/blob/master/test/fetch.js

You can run the tests in that file with a command like this:
```sh
npx ava test/fetch.js
```

Or to run all of the tests:
```sh
npx ava
```

Skayo:
Just took the other test in `test/fetch.js` and modified it to use the option instead of the global for the custom fetch function - hope this is enough!
Let me know if you need anything else

Skayo:
@sholladay @sindresorhus Is this ready to merge?

readme.md
```diff
@@ -368,6 +368,28 @@ import bourne from '@hapijs/bourne';
 })();
 ```
 
+##### fetch
+
+Type: `Function`\
+Default: `fetch`
+
+User-defined `fetch` function.
```
sindresorhus, line 376: I think we should mention that it *has* to be fully compatible with https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

test/fetch.js
```diff
@@ -16,3 +16,16 @@ test.serial('relative URLs are passed to fetch unresolved', async t => {
 	t.is(await ky('unicorn', {prefixUrl: '/api/'}).text(), '/api/unicorn');
 	global.fetch = originalFetch;
 });
+
+test.serial('fetch option takes a custom fetch function', async t => {
+	const customFetch = async input => {
+		return new Response(input.url);
+	};
```
sindresorhus, line 23:
```suggestion
	const customFetch = async input => new Response(input.url);
```

test/fetch.js
```diff
@@ -16,3 +16,16 @@ test.serial('relative URLs are passed to fetch unresolved', async t => {
 	t.is(await ky('unicorn', {prefixUrl: '/api/'}).text(), '/api/unicorn');
 	global.fetch = originalFetch;
 });
+
+test.serial('fetch option takes a custom fetch function', async t => {
```
sholladay, line 20: This test doesn't need to use `.serial` because, unlike the other test, it doesn't have any global side-effects.

test/fetch.js
sholladay: I think we should add some assertions inside of the `customFetch` function and maybe use `t.plan()` to test that the function actually gets called the expected number of times.

So at the top of the test, we would add something like `t.plan(7)` and then in the `customFetch` function:

```js
const customFetch = async (input) => {
	t.is(input instanceof Request);
	return new Response(input.url);
};
```

I should have done this in the other test, too.

readme.md
```diff
@@ -368,6 +368,28 @@ import bourne from '@hapijs/bourne';
...
+1. Use custom `fetch` implementations like [`isomorphic-unfetch`](https://www.npmjs.com/package/isomorphic-unfetch).
+2. Use the `fetch` wrapper function provided by some frameworks that use server-side rendering (SSR).
+
+```js
+import ky from 'ky';
+import fetch from 'isomorphic-unfetch';
+
+(async () => {
+	const parsed = await ky('https://example.com', {
+		fetch: fetch
```
sholladay, line 388: Since the key and value are the same, you can use the shorthand syntax and just write `fetch` instead of `fetch: fetch`.

index.d.ts
```diff
@@ -332,6 +332,29 @@ export interface Options extends Omit<RequestInit, 'headers'> {
...
+	@default fetch
+
+	@example
+	```
+	import ky from 'ky';
+	import fetch from 'isomorphic-unfetch';
+
+	(async () => {
+		const parsed = await ky('https://example.com', {
+			fetch: fetch
```
sholladay, line 352: Same as in the readme, you can use shorthand syntax here: `fetch` instead of `fetch: fetch`

Skayo: Had it like that at first but then I thought it could be kinda confusing for newbies... But sure I'll change it

sholladay: It's definitely something to get used to. Ultimately it simplifies things, though.

test/fetch.js
Skayo: added two more assertions now

Skayo:
@sholladay yup! Ready to merge on my side 🚀

Skayo:
I think the travis build is broken @sholladay

sholladay:
Hmm. The logs on Travis indicate that the build completed successfully, but for some reason GitHub shows the build as still running.

Issue #283, opened 2020-10-05
add "main", "module" and "types" field


rifler:
Because `ky` delivers index.js in esm format and untranspiled, it generates a bit of a problem for "universal" applications

If I use `ky/umd` it works well in node and client, until I want to transpile all of my node_modules, to support old browsers.
After that babel+webpack breaks about `ky/umd` (i haven't figured out the reasons yet)

decision - add these lines to package.json:
```json
"main": "./umd.js",
"module": "./index.js",
"types": "./index.d.ts",
```

explanation:
user will impor just `ky`. Node.js will see `main` field, will require `ky/umd` and won't break about esm. Bundlers like webpack and others will see `module` field and will bundle index.js (esm format).

Transpiling remains on the user's side

Maybe it partially related to https://github.com/sindresorhus/ky/issues/255

sholladay:
I would be fine with this if it actually worked everywhere, but I believe `module` is not as widely supported as you might think, including by Node.js itself and its new ESM loader.

We have what I think is a more forward thinking approach in PR #244 but unfortunately that work has stalled due to some configuration problem in our testing stack that I wasn't able to track down quickly. You could help by looking into that.

rifler:
> I would be fine with this if it actually worked everywhere, but I believe module is not as widely supported as you might think, including by Node.js itself and its new ESM loader.

"Module" widely supported by bundlers, first of all it is webpack and rollup.

But I think that this issue and #244 are not competitors and complementing each other:
`main` - for old versions of node & bundlers. It is CJS
`module` - for bundlers that supports this field, webpack < 5, rollup, etc. It is ESM
`exports` - for new versions of node and bundlers (webpack >= 5). It is dual mode (esm & cjs)

sholladay:
I believe that would break our CDN users who load Ky from `https://unpkg.com/ky`, for example. They would start receiving UMD instead of ESM, which I personally consider unacceptable for a browser oriented library. If `unpkg.com` preferred `module` over `main`, then it would be fine. But I don't think it does.

rifler:
@sholladay can you please provide an example? I'll check that

rifler:
@sholladay You are right, I fixed this behaviour in #284

sholladay:
There are a lot of other CDNs to test, though. I really think it would be better to "default" to ESM with `main` and then fix the Node.js case with `exports`. I'll play around with this and see how it goes, though.

rifler:
jsdelivr ok:
[...]

PR #284, opened 2020-10-05
add "main", "module" and "types" field

rifler:
fixes https://github.com/sindresorhus/ky/issues/283

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
 	"license": "MIT",
 	"repository": "sindresorhus/ky",
 	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
+	"main": "./umd.js",
+	"module": "./index.js",
+	"types": "./index.d.ts",
 	"author": {
 		"name": "Sindre Sorhus",
 		"email": "sindresorhus@gmail.com",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
 	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
 	"main": "./umd.js",
 	"module": "./index.js",
+	"browser": "./index.js",
+	"unpkg": "./index.js",
 	"types": "./index.d.ts",
 	"author": {
 		"name": "Sindre Sorhus",
--- a/test/browser.js
+++ b/test/browser.js
@@ -63,7 +63,7 @@ test('aborting a request', withPage, async (t, page) => {
 		controller.abort();
 		return request.catch(error_ => error_.toString());
 	}, server.url);
-	t.is(error, 'AbortError: The user aborted a request.');
+	t.is(error, 'AbortError: Failed to execute \'fetch\' on \'Window\': The user aborted a request.');
 
 	await server.close();
 });


sholladay:
CI failed. You can run the tests locally with `npm test`.

rifler:
> CI failed. You can run the tests locally with `npm test`.

Fixed, thanks!

But this test fails for me in `master` branch too, I think someone merge PR without passing CI tests

sholladay:
No, that error message comes from the `node-fetch` dependency. They probably changed it in a minor or patch version since the last time our CI ran. I'd argue that they broke SemVer rules. Ideally, they would roll back that change and release it as SemVer major. But oh well. Since it only affects our tests, I won't bother them about it.

rifler:
Oh, I see

maybe we should use package-lock.json for that kind of situations?
Builds must be reproducible over time

I can add package-lock in this PR

sholladay:
Sindre and I prefer to use lockfiles in apps but not in reusable libraries. There's a lot of discussion about it on Sindre's AMA if you want to dig into that more. Anyway, thanks for fixing it.

rifler:
> Sindre and I prefer to use lockfiles in apps but not in reusable libraries. There's a lot of discussion about it on Sindre's AMA if you want to dig into that more. Anyway, thanks for fixing it.

npm never publish lock file (and doesn't use it in dependency).

lock files are nessesary for libraries for reproducible build in CI and so, to prevent situations like this - test fails, even if I didn't touch any code

sholladay:
The short of it is that for libraries we prefer for CI to test the latest dependencies. After all, if Ky were to use a lockfile, It would only affect our CI, not end-users' apps. Thus the lockfile actually provides very little value for a library and can actually cause CI to be green even though a dependency had broken that would affect anyone installing Ky from scratch, which makes the lockfile actively harmful.

Lockfiles for apps, not libraries.

sindresorhus:
> Lockfiles for apps, not libraries.

Yup. https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

package.json
```diff
@@ -5,6 +5,11 @@
 	"license": "MIT",
 	"repository": "sindresorhus/ky",
 	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
+	"main": "./umd.js",
+	"module": "./index.js",
+	"browser": "./index.js",
+	"unpkg": "./index.js",
+	"types": "./index.d.ts",
```
sindresorhus, line 12: This is not needed. TS infers this automatically when named `index.d.ts`.

package.json
```diff
@@ -5,6 +5,11 @@
 	"license": "MIT",
 	"repository": "sindresorhus/ky",
 	"funding": "https://github.com/sindresorhus/ky?sponsor=1",
+	"main": "./umd.js",
```
sindresorhus, line 8: This will definitely break all the things. I think https://github.com/sindresorhus/ky/pull/244/files is a much better solution.

rifler: Ok! will it work for webpack 4 for example?

rifler: why it will break all the things?

package.json
rifler: Ok, I thought that explicit it better than implicit

sholladay:
The issue with the browser tests has been fixed on the master branch.

sindresorhus:
Closing as this is not going anywhere.

package.json
sindresorhus: Because `ky` is an ESM package so the "main" export should be just that.

PR #295, opened 2020-11-02
Fix Deno and React Native compatibility

### LINKED ISSUE #224: TypeScript issue with React Native + /// <reference lib="dom"/>
Hi there! I am consuming this library via [Draqula](https://github.com/vadimdemedes/draqula), and came across a thorny TypeScript issue that I'm struggling to work around.

Basically, the problem is: Draqula uses this library under the hood, and naturally exports some of the underlying `ky` types. This works great in the browser + DOM context. However, when working in React Native, the DOM library reference (`/// <reference lib="dom"/>`) in `index.d.ts`  wreaks havoc, because the DefinitelyTyped React Native types have a bunch of duplicate identifiers (like Request, FormData, etc). Because of that reference, despite my TypeScript compiler settings excluding DOM from the list of provided TypeScript libs, it gets pulled in anyway.

I'm cognizant of the fact that mine is a fringe use case - but is there any way I could convince you to _not_ include that reference at the top of the file? If you need it, you can always provide `libs` via `tsconfig.json` or the command line - but if you don't need it, then there's no way to escape it unless it isn't included as a reference. If there exists an alternative, I'd love to try that out as well. Thanks in advance!

### LINKED ISSUE #294: Ky's types don't work in Deno's Typescript
The types that Ky provides are in compatible with Deno's Typescript and cause duplicate identifier errors:

```ts
TS2300 [ERROR]: Duplicate identifier 'TextEncoder'.
declare class TextEncoder {
              ~~~~~~~~~~~
    at asset:///lib.deno.web.d.ts:179:15

    'TextEncoder' was also declared here.
    interface TextEncoder extends TextEncoderCommon {
              ~~~~~~~~~~~
        at asset:///lib.dom.d.ts:15504:11    and here.
    declare var TextEncoder: {
                ~~~~~~~~~~~
        at asset:///lib.dom.d.ts:15515:13

TS2300 [ERROR]: Duplicate identifier 'AbortController'.
declare class AbortController {
              ~~~~~~~~~~~~~~~
    at asset:///lib.deno.web.d.ts:193:15

    'AbortController' was also declared here.
    interface AbortController {
              ~~~~~~~~~~~~~~~
        at asset:///lib.dom.d.ts:1939:11    and here.
    declare var AbortController: {
                ~~~~~~~~~~~~~~~
        at asset:///lib.dom.d.ts:1950:13
 ```

Someone in Deno's Discord server suggested that removing `/// <reference lib="dom"/>` from index.d.ts would help.

Xeoth:
Removed the line in `index.d.ts` that caused errors when trying to use the library in Deno. This resolves #294 and fixes #224.

I have tested it on Deno 1.5.1 with Typescript 4.0.3 and it seems to work fine.

@sholladay mentioned that they want someone proficient in Typescript to analyze potential drawbacks from doing that, so I'm leaving it up to discussion since I'm far from an expert 😅

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,3 @@
-/// <reference lib="dom"/>
-
 type Primitive = null | undefined | string | number | boolean | symbol | bigint;
 
 type LiteralUnion<LiteralType extends BaseType, BaseType extends Primitive> =


sindresorhus:
It's there to show a friendly error if a user forgets to add the `dom` TS lib, but it seems it's not worth the trouble.

PR #301, opened 2020-11-15
Fix aborting when `onDownloadProgress` throws an exception

zqianem:
(See commit messages.)

For instance, the catch block in the [example in the README](https://github.com/sindresorhus/ky#cancellation) would not catch the AbortError if an onDownloadProgress callback were passed to ky.

I'm not entirely sure about the Chromium bug. With the patches applied in my project that uses ky, I could get Firefox to throw an AbortError, so I suspect https://bugs.chromium.org/p/chromium/issues/detail?id=817687 is not fully fixed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -468,7 +468,7 @@ class Ky {
 
 		return new globals.Response(
 			new globals.ReadableStream({
-				start(controller) {
+				async start(controller) {
 					const reader = response.body.getReader();
 
 					if (onDownloadProgress) {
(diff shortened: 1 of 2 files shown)


sindresorhus (approved):

szmarczak (approved):

sholladay (approved):

sholladay:
LGTM. One nice thing about this is that we could now allow `onDownloadProgress` to be an async function.

PR #307, opened 2020-12-13
Fix support for `ky.stop` as return type in `BeforeRetryHook` TypeScript type

### LINKED ISSUE #288: BeforeRetryHook type definitions
`BeforeRetryHook` has inline type definitions like so:

```ts
export type BeforeRetryHook = (options: {
	request: Request;
	response: Response;
	options: NormalizedOptions;
	error: Error;
	retryCount: number;
}) => void | Promise<void>;
```

Because of that, if I want to declare a `beforeRetry` hook as, for example, a class method, where I'm unable to type it using `BeforeRetryHook` type, I have to type the `options` object manually which is not nice:

```ts
  private async beforeRetry({ error, options }: { error: Error, options: NormalizedOptions }) {
    // ...
  }
```

We should move this inline thing to a stand-alone type, the only problem is to choose a name

UPD: also just noticed `BeforeRetryHook` has return type of `void | Promise<void>`, but it's possible to return symbol `ky.stop` to prevent retrying, but this is now awkward with typescript, because it only wants `void` or `Promise<void>`

rururux:
fixes #288

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,7 +17,7 @@ export type BeforeRetryHook = (options: {
 	options: NormalizedOptions;
 	error: Error;
 	retryCount: number;
-}) => void | Promise<void>;
+}) => symbol | void | Promise<symbol | void>;
 
 export type AfterResponseHook = (
 	request: Request,
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,7 +17,7 @@ export type BeforeRetryHook = (options: {
 	options: NormalizedOptions;
 	error: Error;
 	retryCount: number;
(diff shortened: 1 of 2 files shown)


sindresorhus:
It should be typed to the specific symbol. This also needs a type test.

rururux:
ok, What should I do next?
Sorry for inconveniencing you.

sindresorhus:
> What should I do next?

I wrote it in the last comment:

> This also needs a type test.

rururux:
I'm not confident, is this okay?

PR #308, opened 2020-12-14
Fix `BeforeRetryHook` and `NormalizedOptions` TypeScript type definitions

### LINKED ISSUE #288: BeforeRetryHook type definitions

### LINKED ISSUE #289: Stand-alone type definitions for NormalizedOptions
Now type definitions for `NormalizedOptions` use `Options` in some fields:

```ts
retry: Options['retry'];
prefixUrl: Options['prefixUrl'];
onDownloadProgress: Options['onDownloadProgress'];
```

For me personally it led to awkward code when trying to use `options.retry.statusCodes`, because TypeScript considers `options.retry` may be `number` or not defined at all, so I have to cast it to `RetryOptions`, and `RetryOptions['statusCode']` is declared optional, so I have to use `!.` which is better than casting, but still awkward. This is not supposed to be like that, because `ky` normalizes all retry options, so in `beforeRetry`, `options.retry` is guaranteed to be `RetryOptions` and `options.retry.statusCodes` is surely `number[]`. There is even a comment in type definitions `// Extended from custom `KyOptions`, but ensured to be set (not optional)`, but typescript doesn't think so

ialexi-bl:
Fixes #289 and fixes #288. The latter can be fixed by #307, but it doesn't include moving arguments that `BeforeRetryHook` receives to a separate object that can later be used to declare `beforeRetry` hooks outside of `ky.create` declaration or actual request.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,7 +20,7 @@ export type BeforeRetryOptions = {
 };
 export type BeforeRetryHook = (
 	options: BeforeRetryOptions
-) => void | Promise<void>;
+) => typeof ky.stop | void | Promise<typeof ky.stop | void>;
 
 export type AfterResponseHook = (
 	request: Request,
--- a/index.d.ts
+++ b/index.d.ts
@@ -11,7 +11,7 @@ export type BeforeRequestHook = (
 	options: NormalizedOptions,
 ) => Request | Response | void | Promise<Request | Response | void>;
 
-export type BeforeRetryOptions = {
+export type BeforeRetryState = {
 	request: Request;
 	response: Response;
 	options: NormalizedOptions;
@@ -19,7 +19,7 @@ export type BeforeRetryOptions = {
 	retryCount: number;
 };
 export type BeforeRetryHook = (
-	options: BeforeRetryOptions
+	options: BeforeRetryState
 ) => typeof ky.stop | void | Promise<typeof ky.stop | void>;
 
 export type AfterResponseHook = (
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,7 +13,7 @@ export type BeforeRequestHook = (
 
 export type BeforeRetryOptions = {
 	request: Request;
-	response:  Response;
+	response: Response;
 	options: NormalizedOptions;
 	error: Error;
 	retryCount: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -11,13 +11,16 @@ export type BeforeRequestHook = (
 	options: NormalizedOptions,
 ) => Request | Response | void | Promise<Request | Response | void>;
 
-export type BeforeRetryHook = (options: {
+export type BeforeRetryOptions = {
 	request: Request;
-	response: Response;
+	response:  Response;
 	options: NormalizedOptions;
 	error: Error;
 	retryCount: number;
-}) => typeof ky.stop | void | Promise<typeof ky.stop | void>;
+};
+export type BeforeRetryHook = (
+	options: BeforeRetryOptions
+) => void | Promise<void>;
 
 export type AfterResponseHook = (
 	request: Request,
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -11,13 +11,16 @@ export type BeforeRequestHook = (
 	options: NormalizedOptions,
 ) => Request | Response | void | Promise<Request | Response | void>;
 
-export type BeforeRetryHook = (options: {
+export type BeforeRetryOptions = {
 	request: Request;
-	response: Response;
+	response:  Response;
```
sindresorhus, line 16: Don't do unrelated changes.

sindresorhus:
This needs type tests.

https://github.com/sindresorhus/ky/blob/master/index.test-d.ts

sindresorhus:
Can you rebase from `master`?

ialexi-bl:
Eh, I'm not very familiar with rebases, I hope I didn't messed something up

index.d.ts
```diff
@@ -11,13 +11,16 @@ export type BeforeRequestHook = (
 	options: NormalizedOptions,
 ) => Request | Response | void | Promise<Request | Response | void>;
 
-export type BeforeRetryHook = (options: {
+export type BeforeRetryOptions = {
```
sindresorhus, line 14: I don't think "options" is the correct word here.

index.test-d.ts
```diff
@@ -48,7 +48,9 @@ ky(url, {
 			}
 		],
 		beforeRetry: [
-			({request, response, options, error, retryCount}) => {
+			(beforeRetryOptions) => {
+				const { request, response, options, error, retryCount } = beforeRetryOptions
```
sindresorhus, line 52:
```suggestion
				const {request, response, options, error, retryCount} = beforeRetryOptions
```

index.d.ts
ialexi-bl: What one then? Maybe Params?

sindresorhus: It's better to present solutions than asking questions. Try to think of some good names. I don't think `Params` is a good name. We try not to use abbreviations here.

ialexi-bl: Well, I can't think of a better name than "parameters". The properties of this object don't have anything in common so that there could be a sensible name for them as a whole, it's there just to reduce amount of arguments for `beforeRetry`. Params seems like a very common abbreviation, but it could be `BeforeRetryParameters`. If not this word, I don't know what one tbh

sholladay: I would name it `BeforeRetryState`.

If not `State`, then some other ideas...
 - `Context`
 - `Record`
 - `Variables` (or `Vars`)
 - `Properties` (or `Props`)
 - `Arguments` (or `Args`)

sindresorhus: > Well, I can't think of a better name tha "parameters".

Naming is indeed hard, but I don't think you tried your best here. All I asked was to come up with some alternatives.

sindresorhus: I prefer `Context`, but `State` would work too.

ialexi-bl: > Naming is indeed hard, but I don't think you tried your best here.

Well, sorry for not doing my best for someone else's library. I'll change it to `State`

sindresorhus:
Thanks for the PR :)

PR #309, opened 2020-12-14
Add `request` and `options` to `HTTPError`

### LINKED ISSUE #287: Include request object in HTTPError
I would suggest attaching request object to `HTTPError`. It would be useful to log request body, credentials policy etc. For now I have to use a custom http error, which leads to some problems with retry logic, so now I have to write a `beforeRetry` hook that checks whether request should be retried or not. It doesn't seem to do any damage to include request in `HTTPError`, so why not to do it? Also you could include `options` as well, because after normalization they could be used for logging as well

ialexi-bl:
Fixes #287

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -414,8 +414,10 @@ export interface ResponsePromise extends Promise<Response> {
 The error has a response property with the `Response` object.
 */
 declare class HTTPError extends Error {
-	constructor(response: Response);
+	constructor(response: Response, request: Request, options: NormalizedOptions);
 	response: Response;
+	request: Request;;
+	options: NormalizedOptions;;
 }
 
 /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -411,7 +411,7 @@ export interface ResponsePromise extends Promise<Response> {
 }
 
 /**
- Exposed for `instanceof` checks. The error has a `response` property with the `Response` object, `request` property with the `Request` object and `options` property with normalized options (either passed to `ky` when creating an instance with `ky.create()` or directly when performing the request).
+Exposed for `instanceof` checks. The error has a `response` property with the [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response), `request` property with the [`Request` object](https://developer.mozilla.org/en-US/docs/Web/API/Request), and `options` property with normalized options (either passed to `ky` when creating an instance with `ky.create()` or directly when performing the request).
 */
 declare class HTTPError extends Error {
 	constructor(response: Response, request: Request, options: NormalizedOptions);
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -414,8 +414,10 @@ export interface ResponsePromise extends Promise<Response> {
 The error has a response property with the `Response` object.
 */
 declare class HTTPError extends Error {
-	constructor(response: Response);
+	constructor(response: Response, request: Request, options: NormalizedOptions);
 	response: Response;
+	request: Request;;
+	options: NormalizedOptions;;
```
sindresorhus, line 420: Typo

index.test-d.ts
```diff
@@ -23,7 +23,13 @@ for (const method of requestMethods) {
 
 expectType<typeof ky>(ky.create({}));
 expectType<typeof ky>(ky.extend({}));
-expectType<InstanceType<typeof ky.HTTPError>>(new ky.HTTPError(new Response()));
+expectType<InstanceType<typeof ky.HTTPError>>(
+	new ky.HTTPError(
+		new Response(),
+		new Request("http://example.com"),
```
sindresorhus, line 29:
```suggestion
		new Request('https://example.com'),
```

sindresorhus:
You also need to update the TS doc comment and the readme.

sindresorhus (changes requested):

index.js
```diff
@@ -314,7 +316,7 @@ class Ky {
 			this._decorateResponse(response);
 
 			if (!response.ok && this._options.throwHttpErrors) {
-				throw new HTTPError(response);
+				throw new HTTPError(response, this.request, this._optoins);
```
dan-lee, line 319:
```suggestion
				throw new HTTPError(response, this.request, this. _options);
```

nielsdB97: Watch out, there's an added space there ☝️ 

```suggestion
				throw new HTTPError(response, this.request, this._options);
```

dan-lee: Thanks for spotting @nielsdB97 👀!  I edited the suggestion

ialexi-bl:
Why is it displaying "Changes requested"? I've added all of the suggestions in the last commit it seems

index.d.ts
```diff
@@ -411,13 +411,13 @@ export interface ResponsePromise extends Promise<Response> {
 }
 
 /**
-The error has a response property with the `Response` object.
+The error has a response, request and options properties that contain `Response`, `Request` and `NormalizedOptions` objects respectively
```
sindresorhus, line 414: Should end with a dot. And this text and the readme docs should be in sync.

sholladay:
`index.d.ts` needs to be synced with the readme, as Sindre mentioned, but otherwise LGTM.

PR #313, opened 2021-01-01
Make the package a module type

### LINKED ISSUE #283: add "main", "module" and "types" field

sindresorhus:
Fixes #283

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,11 @@
 		"email": "sindresorhus@gmail.com",
 		"url": "https://sindresorhus.com"
 	},
+	"type": "module",
+	"exports": {
+		"import": "./index.js",
+		"require": "./umd.js"
+	},
 	"engines": {
 		"node": ">=10"
 	},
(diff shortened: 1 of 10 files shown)


sholladay:
Wow, awesome!

I love that AVA is going full ESM.

package.json
```diff
@@ -69,13 +74,22 @@
...
-		]
+		],
+		"rules": {
+			"import/no-useless-path-segments": "off",
+			"unicorn/import-index": "off",
+			"import/extensions": "off"
+		}
 	},
 	"ava": {
+		"serial": true,
```
sholladay, line 85: Is this necessary? It will slow down testing and I think the ky tests can generally run concurrently.

sholladay (approved):

package.json
sindresorhus: Good point

PR #314, opened 2021-01-04
Improve docs for `ky.stop`

### LINKED ISSUE #290: Broken ky.stop

ialexi-bl:
Fixes #290. I would also like to add `console.warn` in shortcut methods, such as `.json()` and `.text()` that would display a warning when ky.stop is returned and these methods are used. Maybe a lazy solution to check if this `result` is not `undefined` could work:

https://github.com/sindresorhus/ky/blob/d9760294875b46263265fa4a4c470edfb3347295/index.js#L338

I can definitely add a check that it's not `undefined` to prevent those from throwing `TypeError` and to make them silently return `undefined` instead. The question is: is it guaranteed that if `result` is `undefined`, this was caused by `ky.stop`? If yes, I could also add `console.warn` in the same place where this check for `result !== undefined` is performed.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -256,9 +256,9 @@ Default: `[]`
 
 This hook enables you to modify the request right before retry. Ky will make no further changes to the request after this. The hook function receives an object with the normalized request and options, an error instance, and the retry count. You could, for example, modify `request.headers` here.
 
-If the request received a response, `error` will be an `HTTPError` and the `Response` object will be available at `error.response`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received. In that case, `error` will not be an instance of `HTTPError`, though.
+If the request received a response, the error will be of type `HTTPError` and the `Response` object will be available at `error.response`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received. In that case, the error will not be an instance of `HTTPError`.
 
-You can prevent Ky from retrying the request by throwing `error`. Ky will not handle it in any way, this error will be propagated to the request initiator. The rest of the `beforeRetry` hooks will not be called if you `throw error`. Alternatively, you can return the [`ky.stop`](#ky.stop) symbol to do the same thing but without throwing (this has some limitations, see `ky.stop` docs for details).
+You can prevent Ky from retrying the request by throwing an error. Ky will not handle it in any way and the error will be propagated to the request initiator. The rest of the `beforeRetry` hooks will not be called in this case. Alternatively, you can return the [`ky.stop`](#ky.stop) symbol to do the same thing but without propagating an error (this has some limitations, see `ky.stop` docs for details).
 
 ```js
 import ky from 'ky';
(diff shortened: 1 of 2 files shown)


sholladay (approved):

readme.md
```diff
@@ -256,7 +256,9 @@ Default: `[]`
 
 This hook enables you to modify the request right before retry. Ky will make no further changes to the request after this. The hook function receives an object with the normalized request and options, an error instance, and the retry count. You could, for example, modify `request.headers` here.
 
-If the request received a response, it will be available at `error.response`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received.
+If the request received a response, `error` will be an `HTTPError` and the `Response` object will be available at `error.response`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received. In that case, `error` will not be an instance of `HTTPError`, though.
```
sindresorhus, line 259:
```suggestion
If the request received a response, the the error will be of type `HTTPError` and the `Response` object will be available at `error.response`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received. In that case, the error will not be an instance of `HTTPError`.
```

readme.md
```diff
@@ -256,7 +256,9 @@ Default: `[]`
...
+
+You can prevent Ky from retrying the request by throwing `error`. Ky will not handle it in any way, this error will be propagated to the request initiator. The rest of the `beforeRetry` hooks will not be called if you `throw error`. Alternatively, you can return the [`ky.stop`](#ky.stop) symbol to do the same thing but without throwing (this has some limitations, see `ky.stop` docs for details).
```
sindresorhus, line 261:
```suggestion
You can prevent Ky from retrying the request by throwing the error. Ky will not handle it in any way and the error will be propagated to the request initiator. The rest of the `beforeRetry` hooks will not be called if you throw an error. Alternatively, you can return the [`ky.stop`](#ky.stop) symbol to do the same thing but without throwing (this has some limitations, see `ky.stop` docs for details).
```

readme.md
```diff
@@ -468,22 +470,29 @@ The error thrown when the request times out.
...
 		}
-	});
+	}
+
+	await ky.post('https://example.com', options);
+	
+	// Using .text() or other body methods is not suppported
+	const text = await ky('https://example.com', options).text();
 })();
 ```
```
sindresorhus, line 497: Why is this example even here if it's no longer using `ky.stop`?

readme.md
```diff
@@ -468,22 +470,29 @@ The error thrown when the request times out.
 
 A `Symbol` that can be returned by a `beforeRetry` hook to stop the retry. This will also short circuit the remaining `beforeRetry` hooks.
 
+Note: This aborts _successfully_ and Ky will return with an `undefined` response. If you use `ky.stop`, be sure to check for a response first before accessing any properties on it or use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). It is also incompatible with body methods, such as `.json()` or `.text()`, because there is no response to parse. In general, we recommend to `throw error` instead of `return ky.stop`, as that will abort _unsuccesfully_ and Ky will throw, which avoids these limitations.
```
sindresorhus, line 473:
```suggestion
Note: Returning this symbol makes Ky abort and return with an `undefined` response. Be sure to check for a response before accessing any properties on it or use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). It is also incompatible with body methods, such as `.json()` or `.text()`, because there is no response to parse. In general, we recommend throwing an error instead of returning this symbol, as that will cause Ky to abort and then throw, which avoids these limitations.
```

readme.md
```diff
@@ -468,22 +470,29 @@ The error thrown when the request times out.
...
+
+A valid use case for `ky.stop` is to prevent retries when making requests for side effects, where returned data is not important. For example, logging client activity to the server.
```
sindresorhus, line 475:
```suggestion
A valid use-case for `ky.stop` is to prevent retries when making requests for side effects, where the returned data is not important. For example, logging client activity to the server.
```

sindresorhus (changes requested):

sindresorhus:
index.d.ts should be in sync with the readme.

readme.md
ialexi-bl: Oops, seems like I removed `ky.stop` occasionally, I'll fix that

ialexi-bl:
I would also suggest to change tabs in code snippets in readme to spaces, because they are pretty hard to read with many levels of nesting, considering how wide tabs on github are. Maybe in a different PR though

sindresorhus:
> I would also suggest to change tabs in code snippets in readme to spaces, because they are pretty hard to read with many levels of nesting, considering how wide tabs on github are. Maybe in a different PR though

No thanks :)

PR #315, opened 2021-01-06
Require Node.js 12 and drop UMD build

sindresorhus:
I have decided to target Node.js 12 and drop the UMD build. Node.js 10 will be obsolete in a few months anyway.

This lets us go full ESM, which reduces the chances of [problems](https://twitter.com/bradleymeck/status/1346466891070664704) as some bundlers cannot handle `exports` in package.json.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,27 +2,6 @@
 
 const globals = {};
 
-const getGlobal = property => {
-	/* istanbul ignore next */
-	if (typeof self !== 'undefined' && self && property in self) {
-		return self;
-	}
-
-	/* istanbul ignore next */
-	if (typeof window !== 'undefined' && window && property in window) {
-		return window;
-	}
-
-	if (typeof global !== 'undefined' && global && property in global) {
-		return global;
-	}
-
-	/* istanbul ignore next */
-	if (typeof globalThis !== 'undefined' && globalThis) {
-		return globalThis;
-	}
-};
-
 const globalProperties = [
 	'Headers',
 	'Request',
@@ -36,9 +15,8 @@ const globalProperties = [
 for (const property of globalProperties) {
 	Object.defineProperty(globals, property, {
 		get() {
-			const globalObject = getGlobal(property);
-			const value = globalObject && globalObject[property];
-			return typeof value === 'function' ? value.bind(globalObject) : value;
+			const value = globalThis[property];
+			return typeof value === 'function' ? value.bind(globalThis) : value;
 		}
 	});
 }
--- a/index.js
+++ b/index.js
@@ -503,4 +503,6 @@ const createInstance = defaults => {
 	return ky;
 };
 
-export default createInstance();
+const ky = createInstance();
+
+export default ky;
--- a/test/browser.js
+++ b/test/browser.js
@@ -8,6 +8,14 @@ import withPage from './helpers/with-page.js';
 const test = ava.serial;
 const pBody = util.promisify(body);
 
+const kyScript = {
+	type: 'module',
+	content: `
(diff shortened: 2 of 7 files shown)


test/browser.js
```diff
@@ -8,6 +8,14 @@ import withPage from './helpers/with-page.js';
 const test = ava.serial;
 const pBody = util.promisify(body);
 
+const kyScript = {
+	type: 'module',
+	content: `
+		import ky from './index.js';
+		window.ky = ky;
```
sindresorhus, line 15: Any ideas why this doesn't work? I'm still getting:

```
  browser › headers are preserved when input is a Request and there are searchParams in the options

  Error {
    message: `Evaluation failed: TypeError: window.ky is not a function␊
        at __puppeteer_evaluation_script__:6:17`,
  }

  › __puppeteer_evaluation_script__:6:17
  › ExecutionContext._evaluateInternal (node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:217:19)
  › ExecutionContext.evaluate (node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:106:16)
  › file://test/browser.js:331:2
  › withPage (file://test/helpers/with-page.js:7:3)
```

in the browser tests.

sholladay: An educated guess... I think what's happening is that the browser is trying to load `/index.js` but there's no such route on the `server` in the tests. We didn't need a route prior to this PR because Puppeteer itself (on the Node.js side) was reading the file at `path` from the filesystem and injecting it inline as a string in a script tag.

test/browser.js
```diff
@@ -316,13 +324,15 @@ test('headers are preserved when input is a Request and there are searchParams i
 		response.end();
 	});
 
+	await page.setBypassCSP(true);
```
sholladay, line 327: Almost didn't notice this. What's it for?

index.js
```diff
@@ -2,27 +2,6 @@
 
 const globals = {};
 
-const getGlobal = property => {
```
sholladay, line 5: This could break some edge cases, including the scenario from https://github.com/sindresorhus/ky/pull/203 and https://github.com/sindresorhus/ky/issues/202. Basically, I've seen cases where some of the globals we rely on were actually defined on different objects (i.e. not true globals), particularly in mocked test environments.

That said, we're less reliant on these globals since the last time I had any issues and now that Node.js actually has `globalThis` and I can do `globalThis.window = globalThis`... maybe it won't be a problem? I'm not entirely sure. I'm 👍 on simplifying this, but let's definitely mark it as a breaking change.

test/browser.js
sindresorhus: Just forgot to remove it after testing something.

index.js
sindresorhus: Let's try and see. I would prefer for us to not have to have lots of workarounds for weird use-cases.

test/browser.js
sindresorhus: I managed to solve it. Really weird that the `import` call didn't throw though.

sindresorhus:
I'm skipping the browser tests on CI for now. They always succeed locally, but randomly fail with `Evaluation failed: TypeError: window.ky is not a function` on CI. Feels like some kind of internal race issue in Puppeteer. I wasted a lot of time trying to debug this.

rifler:
I'm sorry, but full esm means, that consumers should rewrite their code to esm too, otherwise it would be error `require of es modules is not supported`

szmarczak:
@rifler You need to use `import` instead of `require`.

PR #319, opened 2021-01-16
Simplify global variable usage

sholladay:
As of https://github.com/sindresorhus/ky/pull/315, Ky requires the fetch related APIs to be true globals on `globalThis`, as opposed to dynamically looking them up on various namespaces. Thus the list of globals and getters for them no longer serves a purpose.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,16 +1,14 @@
 /*! MIT License © Sindre Sorhus */
 
-const globals = globalThis;
-
 const isObject = value => value !== null && typeof value === 'object';
-const supportsAbortController = typeof globals.AbortController === 'function';
-const supportsStreams = typeof globals.ReadableStream === 'function';
-const supportsFormData = typeof globals.FormData === 'function';
+const supportsAbortController = typeof globalThis.AbortController === 'function';
+const supportsStreams = typeof globalThis.ReadableStream === 'function';
+const supportsFormData = typeof globalThis.FormData === 'function';
 
 const mergeHeaders = (source1, source2) => {
-	const result = new globals.Headers(source1 || {});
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -1,25 +1,6 @@
...
-
-for (const property of globalProperties) {
-	Object.defineProperty(globals, property, {
-		get() {
-			const value = globalThis[property];
-			return typeof value === 'function' ? value.bind(globalThis) : value;
-		}
-	});
-}
+const globals = globalThis;
```
sindresorhus, line 3: Can we drop this too and just use `globalThis` directly everywhere?

sholladay: Sure, done. I find it a bit lengthy and strange to look at. Maybe I just need to get used to it.

kirillgroshkov:
FYI: `globalThis` is supported on iOS 12.2+ (https://caniuse.com/mdn-javascript_builtins_globalthis), so it broke our app for iOS < 12.2 users.

For those who need a quick polyfill:

```typescript
;(window as any).globalThis = window
```

sholladay:
Hi @kirillgroshkov. This pull request has no user-facing changes. Also, it was merged but it has not been released yet in any version of Ky.

It sounds like you are talking about PR #315. But that was released in v0.26 where we made multiple breaking changes. So you would've had to explicitly upgrade to that version. You can stay at v0.25.1 if you need to.

kirillgroshkov:
> It sounds like you are talking about PR #315. But that was released in v0.26 where we made multiple breaking changes. So you would've had to explicitly upgrade to that version. You can stay at v0.25.1 if you need to.

Yes, you're right

PR #325, opened 2021-02-18
Don't mangle user-provided `searchParams` string

contra:
When providing a raw searchParams string, currently ky will pass it through `new URLSearchParams()` which has the result of mangling it in places where it is perfectly valid. A simple example is a querystring like `a&b=` (which is the output of `qs.stringify({ a: null, b: '' }, { strictNullHandling: true })`) is changed to `a=&b=` which semantically has a different meaning (a and b are both empty string).

This change is pretty simple, if a user provides a string don't mess with it - use it directly. This is a breaking change and should be released as a major if landed. I added one new test for a specific case with null values and also had to update other tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -232,7 +232,10 @@ class Ky {
 		this.request = new globalThis.Request(this._input, this._options);
 
 		if (this._options.searchParams) {
-			const searchParams = '?' + new URLSearchParams(this._options.searchParams).toString();
+			const textSearchParams = typeof this._options.searchParams === 'string' ?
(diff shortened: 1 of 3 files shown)


sholladay (approved):
The code looks good and I agree the behavior is arguably better. Though, by default all of the major parsers treat both `?foo` and `?foo=` as meaning that `foo` is an empty string and I'd strongly suggest not changing those defaults, otherwise it'll probably bite you at some point.

Issue #321, opened 2021-01-21
Any plan/interest to rewrite in TypeScript?


mgcrea:
Seeing that got is now 100% typescript, is there any plan or ongoing effort on porting this sister project to TypeScript?

sholladay:
I'm personally on board with this. It would certainly be nice for Deno users, as it would remove the need to manually import the types.

sindresorhus:
I'm not sure I want that. The Got rewrite to TS ended up being a nightmare. TS just so buggy.

sindresorhus:
> as it would remove the need to manually import the types.

Why would you need to manually import the types? Can Deno really not infer it automatically?

sholladay:
Deno imports modules directly from URLs. They have an excellent on-disk caching system so things get loaded from the file system when possible but they've made a design choice to not probe servers for `.d.ts` files for every new `.js` request. I get it, honestly. Those HTTP requests aren't free.

There _is_ an HTTP header that CDNs can use to explicitly point Deno to the TS types and it will then load them automatically. That's what the `?dts` query parameter on the Skypack CDN does, which we recommend to Deno users in the readme. So I wouldn't say this is a big deal. But it would be nice if Deno users could just load an `index.ts` file from any CDN and not have to worry about those details.

mgcrea:
> I'm not sure I want that. The Got rewrite to TS ended up being a nightmare. TS just so buggy.

I can't attest for the ongoing maintenance burden of got as a TS project but as someone that was pretty reluctant to migrate both personal projects and published libraries to TS, I really find that it ends up being a pretty big net positive in both development speed and overall robustness of the code.

Also since `ky` does not have any dependencies and `Fetch` related API must be pretty stable now. I don't think there would be much external TS-related burden.

Anyway, as maintainers it is your call to make. But if you reconsider it in the future, I could spend some time on it and give it a try and see how it goes.

sindresorhus:
There are multiple things:

- TS broke something in pretty much every minor release (from talking to people, this is a common occurrence).
- Node.js types are often outdated or incorrect. The situation might be better for DOM since it's now auto-generated.
- Tooling is annoying. We cannot move to ESM because AVA (test runner) doesn't yet support ESM with TS.
- Bloated package because of all the `d.ts` files.
- https://github.com/sindresorhus/got/issues/1137
- We have encountered a myriad of TS bugs where we end up on an old unresolved issue on the TS issue tracker. A pretty common occurrence.
[...]

PR #330, opened 2021-03-05
Move to TypeScript

mgcrea:
Following #321, I took some time today to attempt a port to TypeScript with minimal changes.

Work is not fully done yet (did not update the CI pipeline, hence the failed tests), but it does pass typescript validation.

There is a two places where I had to `@ts-expect-error`, related to dynamically assigned props as I'm not sure yet how I can port it to Typescript (could be done easily by assigning each props without the loop, see reviews). Open to ideas on this.

Also there is some remaining typing issues (existing in the current code but invisible) regarding the actual options passed around (see `InternalOptions` vs `NormalizedOptions`) so I had to add a few manual `as` casting as well that should be fixed (but could be done after the migration to keep changes to a minimum).

Also I had to refactor the Ky class API a bit (added a static `Ky.create`) that did return a `ResponsePromise` from its constructor instead of the expected `Ky` instance as it looks like TypeScript forbids class constructor return type overloads.

I tried my best to follow what was made for `got` and to not change anything in the code yet (I have quite a few changes that I'd like to propose next but that were out of scope).

Before I spend too much more time on this I'd love to get some feedback on the direction of this merge request and the probability of it being merged once ready.

I'm willing to change anything that you would not like (naming, etc.).

Thanks for the review!


sholladay:
Hi @mgcrea and thanks for your effort. I'll do a proper review once Sindre chimes in with whether he wants to move forward with this (given the comments on #321) and the tests are passing. I don't have strong feelings myself, except to say that - as you've encountered - some of the patterns we use in Ky don't seem to be fully supported by TypeScript and I dislike writing code differently than we would otherwise _just to please the compiler_. Especially when it comes to the public API, which seems to be the case here with the static `Ky.create` if I understand correctly.

That said, for Deno users like myself this would be awesome. The current approach to loading type definition files in Deno is kludgy and non-standard, so I'd love to remove the need for that.

mgcrea:
@sholladay thanks! If I'm not mistaken, `new Ky()` was an internal API so I don't think there is any major issue here. I understand your point about having to please the compiler, but in most cases I find the advices pretty sound and that do lead to simpler to understand and debug code.

Waiting for @sindresorhus blessings to continue with unit tests.

sholladay:
Correct, the Ky class is an internal concept. I thought that making `create()` a static method implied breaking the ability to call `create()` on an instance, but I can see in the code that you've kept that ability, so never mind on that.

sindresorhus:
I'm ok with moving forward with this as the code is complicated and it's nice having some type safety when refactoring or adding features.

sindresorhus:
What's the plan for testing? The output should be ESM.

> Tooling is annoying. We cannot move to ESM because AVA (test runner) doesn't yet support ESM with TS. - https://github.com/sindresorhus/ky/issues/321#issuecomment-765594679

mgcrea:
@sindresorhus Haven't really thought about it yet but I guess we could use a specific tsconfig build for tests (without ESM) along with [`@ava/typescript`](https://github.com/avajs/typescript), and still output multiple independent builds depending on what we want to support (amd/cjs/esm)?

I'm usually using [`ts-jest`](https://github.com/kulshekhar/ts-jest) in my projects where ts is a first-class citizen. So I could also easily port the testing setup to use jest but I guess you probably want to keep testing uniform among your projects and stick with ava.

Will try to have something working pretty soon.

sindresorhus:
> we could use a specific tsconfig build for tests (without ESM) along with @ava/typescript

I guess that might work.

> and still output multiple independent builds depending on what we want to support (amd/cjs/esm)?

ESM is the only production target.

mgcrea:
@sindresorhus I've just finished to fully migrate the tests to typescript. Was a bit painful to find the exact working setup to properly test the ESM build from non-ESM compiled unit-tests that nyc/ava can digest, but ended up finding something thanks to playwright debug mode.

I did bump all dev deps, so I had to fix a few linting-related issues in the code to have `xo`-related test pass, disabled a couple of eslint rules as well either related to typescript and to keep changes to a minimum.

Had to rewrite a bit the helpers part related to `createHttpServer` that was using an old non-ts module (took the helper directly from got) and had to add a couple of helpers to properly load the ESM build from chrome.

I went with named exports across the source code coupled with all-star-exports `index.js` files which is a setup I heavily use nowadays, however for an ESM-published lib, we might want to keep files to a minimum and explicitly import every dep. Can refactor that quickly if you want (or can be done later).

Finally had to add `.js` extensions everywhere so that the esm build properly works (since typescript don't want to have anything with import-related transforms).

If you find any issue blocking the merge don't hesitate to ping me and I'll do my best to address it / fix it in a timely manner.

source/core/Ky.ts
```diff
@@ -0,0 +1,322 @@
+/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
```
sindresorhus, line 1: Don't disable it for the whole file. Disable it where it needs to be, but keep in mind that most cases, the rule is correct.

package.json
```diff
@@ -10,17 +10,18 @@
...
-	"exports": "./index.js",
+	"exports": "./lib/index.js",
 	"engines": {
 		"node": ">=12"
 	},
 	"scripts": {
-		"test": "xo && ava && tsd"
+		"test": "xo && npm run build && nyc ava",
+		"release": "np",
+		"build": "del-cli dist/ lib/ && tsc && tsc -p tsconfig.lib.json",
```
sindresorhus, line 20: Use long-flag for `-p`.

.gitignore
```diff
@@ -1,4 +1,10 @@
-node_modules
+.eslintrc
+.nyc_output/
+.prettierrc
+.vscode/
```
sindresorhus, line 4: These are not project-related and should be in your own global gitignore file.

source/utils/merge.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	return deepMerge({}, ...sources);
+};
+
+export const mergeHeaders = (source1: KyHeadersInit = {}, source2: KyHeadersInit = {}) => {
+	const result = new globalThis.Headers(source1 as HeadersInit);
+	const isHeadersInstance = source2 instanceof globalThis.Headers;
+	const source = new globalThis.Headers(source2 as HeadersInit);
+
+	// eslint-disable-next-line unicorn/no-array-for-each
```
sindresorhus, line 19: Adhere to the linting.

package.json
```diff
@@ -67,31 +79,40 @@
 			"globalThis"
 		],
 		"rules": {
-			"import/no-useless-path-segments": "off",
-			"unicorn/import-index": "off",
-			"import/extensions": "off"
+			"@typescript-eslint/dot-notation": "off",
+			"@typescript-eslint/quotes": "off",
```
sindresorhus, line 83: Don't disable this rule.

package.json
```diff
@@ -67,31 +79,40 @@
...
+			"ava/no-import-test-files": "off",
+			"unicorn/filename-case": "off",
+			"unicorn/prevent-abbreviations": "off"
```
sindresorhus, line 86: Generally don't disable this. If it has false-positives, add inline `// eslint-disable..` comments.

sindresorhus (changes requested):

sindresorhus:
Commented on some nitpick. Will do a more in-depth review tomorrow.

sindresorhus:
> Another thing we should decide before publishing a release is the tsc compiler target, currently es2019 I guess would not work in some old browsers.

`es2019` is fine. It's what it currently is (Node.js 12). Ky only targets modern browsers.

mgcrea:
Looking at the code today I realised that during my tests I had dropped the `"type": "module"` from the package.json, unfortunately adding it back does break testing (since test are built with require but require is forbidden when type is module). So I had to hack something around by monkey-patching a copied package.json with commonjs module inside the dist folder.

Will be discarded when nyc/ava does support ESM.

sindresorhus:
Regarding testing, seems there's a solution with `ts-node`: https://github.com/avajs/ava/blob/main/docs/recipes/typescript.md#for-packages-with-type-module

mgcrea:
Just finished implementing the solution you linked and it drastically simplified the testing process as it no longer requires building the tests (so no more dist/ folder) and no package.json monkey-patching needed!

One downside however is that we can no longer test against NodeJS v12, but since it's a browser lib I'm not sure there is much benefit testing against a matrix of NodeJS versions (ideally we should test against a matrix of browsers).

sholladay:
> One downside however is that we can no longer test against NodeJS v12

We do support Node when combined with polyfills, but I'm fine with dropping Node 12 from the test matrix.

> ideally we should test against a matrix of browsers

That's what `test/browser.js` does. It uses Playwright to run some tests in a headless browser. It's currently set up to test against Chromium, but we could configure it to test Webkit and Firefox as well.

sindresorhus:
> but I'm fine with dropping Node 12 from the test matrix.

👍

package.json
```diff
@@ -11,16 +11,21 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
-	"exports": "./index.js",
+	"main": "./lib/index.js",
```
sindresorhus, line 14: This is moot. `exports` replaces `main`.

package.json
```diff
@@ -45,18 +50,29 @@
 		"node-fetch"
 	],
 	"devDependencies": {
+		"@ava/typescript": "^1.1.1",
```
sindresorhus, line 53: I think this is unused now.

package.json
```diff
@@ -45,18 +50,29 @@
...
+		"@sindresorhus/tsconfig": "^0.9.0",
```
sindresorhus, line 54: Upgrade to latest.

source/index.ts
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable @typescript-eslint/promise-function-async */
```
sindresorhus, line 1: Disable it inline when needed instead of for the whole file.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+	"compilerOptions": {
+		"target": "ES2019", // Node.js 12
+		"lib": [
+			"ES2019",
+			"DOM",
+			"DOM.Iterable",
+		],
+		"useDefineForClassFields": false,
+		"allowSyntheticDefaultImports": true,
+		"esModuleInterop": true,
```
sindresorhus, line 12: I don't think we need this now that the package is ESM.

sindresorhus (changes requested):

mgcrea:
@sindresorhus I've dropped the `node@12` from the GitHub action, is there anything else I can do to move this pr forward?

package.json
```diff
@@ -45,18 +49,28 @@
...
-		"delay": "^4.1.0",
-		"form-data": "^3.0.0",
-		"node-fetch": "^2.5.0",
-		"nyc": "^15.0.0",
-		"playwright-chromium": "^1.7.1",
-		"tsd": "^0.13.1",
-		"xo": "^0.25.3"
+		"del-cli": "^3.0.1",
+		"delay": "^5.0.0",
+		"express": "^4.17.1",
```
sholladay, line 64: From what I can tell, the reason you're using Express directly is so that you can add types to the server. But why not just add types to the `create-test-server` package? Then the whole community can more easily benefit from that work and we don't have to maintain it in Ky.

package.json
```diff
@@ -45,18 +49,28 @@
...
+		"form-data": "^4.0.0",
+		"node-fetch": "^2.6.1",
+		"pify": "^5.0.0",
+		"playwright-chromium": "^1.10.0",
+		"raw-body": "^2.4.1",
+		"requirejs": "^2.3.6",
```
sholladay, line 70: I can't tell what this is being used for. Also, RequireJS is legacy tech and unmaintained. It hasn't been updated in 3 years.

package.json
```diff
@@ -45,18 +49,28 @@
...
-		"playwright-chromium": "^1.7.1",
-		"tsd": "^0.13.1",
-		"xo": "^0.25.3"
+		"del-cli": "^3.0.1",
+		"delay": "^5.0.0",
+		"express": "^4.17.1",
+		"form-data": "^4.0.0",
+		"node-fetch": "^2.6.1",
+		"pify": "^5.0.0",
+		"playwright-chromium": "^1.10.0",
```
sholladay, line 68: Updating dependencies and stuff makes it harder to know what's strictly necessary vs an unrelated improvement. Would be nice to do it in a separate issue/PR, for clarity's sake.

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+	"compilerOptions": {
+		"target": "ES2019", // Node.js 12
+		"lib": [
+			"ES2019",
+			"DOM",
+			"DOM.Iterable",
+		],
+		"useDefineForClassFields": false,
+		"allowSyntheticDefaultImports": true,
+		"module": "ES2020"
```
sholladay, line 12: Looks like `@sindresorhus/tsconfig` has been updated recently to include most of this stuff by default.

tsconfig.lib.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "./tsconfig",
```
sholladay, line 2: I don't understand why we have two TS configs here. Seems like there should only be one.

sholladay (changes requested):

.gitignore
```diff
@@ -1,4 +1,7 @@
-node_modules
+.nyc_output/
+*.tsbuildinfo
+coverage/
+dist/
```
sindresorhus, line 4: Doesn't seem like this is used.

source/core/Ky.ts
```diff
@@ -0,0 +1,315 @@
...
+import {normalizeRequestMethod, normalizeRetryOptions} from '../utils/normalize.js';
+import {delay, timeout, TimeoutOptions} from '../utils/time.js';
+import {maxSafeTimeout, responseTypes, stop, supportsAbortController, supportsFormData, supportsStreams} from './constants.js';
+
+export class Ky {
+	public request: Request;
+	public abortController?: AbortController;
+	protected _retryCount = 0;
+	protected _input: Input;
+	protected _options: InternalOptions;
```
sindresorhus, line 16: Can we use JS private class fields here?

source/core/Ky.ts
```diff
@@ -0,0 +1,315 @@
...
+	// eslint-disable-next-line complexity
+	constructor(input: Input, options: Options = {}) {
+		this._retryCount = 0;
```
sindresorhus, line 19: This is moot. It's already set.

source/core/Ky.ts
```diff
@@ -0,0 +1,315 @@
...
+import type {Input, InternalOptions, NormalizedOptions, Options} from '../types/options.js';
+import {ResponsePromise} from '../types/response.js';
+import {deepMerge, mergeHeaders} from '../utils/merge.js';
+import {normalizeRequestMethod, normalizeRetryOptions} from '../utils/normalize.js';
+import {delay, timeout, TimeoutOptions} from '../utils/time.js';
+import {maxSafeTimeout, responseTypes, stop, supportsAbortController, supportsFormData, supportsStreams} from './constants.js';
+
+export class Ky {
+	public request: Request;
+	public abortController?: AbortController;
```
sindresorhus, line 13: I don't think we want to make this public.

source/core/Ky.ts
```diff
@@ -0,0 +1,315 @@
...
+
+				// @ts-expect-error not sure how to properly type this!
+				return response[type]();
+			};
+		}
+
+		return result as ResponsePromise;
+	}
+
+	_calculateRetryDelay(error: Error) {
```
sindresorhus, line 176: This should be marked as protected. Same with the other internal methods.

tsconfig.lib.json
```diff
@@ -0,0 +1,11 @@
...
+	"compilerOptions": {
+		"sourceMap": true,
+		"rootDir": "./source",
+		"outDir": "lib"
```
sindresorhus, line 6: I would prefer the output directory to be called `distribution`.

sindresorhus (changes requested):

sindresorhus:
@mgcrea Don't force push. It makes it impossible to review only what changed.

sindresorhus:
@mgcrea Don't force push. It makes it impossible to review only what changed.

sindresorhus:
@mgcrea Did you see @sholladay's feedback?

mgcrea:
Yes and I think I've answered every comment, to me there is nothing really blocking left (but I could be wrong). Point taken regarding force-pushing, won't do it again.

sholladay:
I don't see those responses to my comments, @mgcrea.

mgcrea:
Indeed it looks like I did not properly understand how the reviewing process works, all my comments were flagged as pending and I thought you could see them. I had to « submit review » to properly publish them...! 😅 Good to know. I also marked every conversation as resolved as an earlier attempt to publish so feel free to unresolve the point you want to have addressed / discussed.

sindresorhus:
For the remaining `@ts-expect-error`. I suggest we just leave it and open an issue about fixing `@ts-expect-error`. Maybe someone else knows how to correctly type those.

We can link to https://github.com/sindresorhus/ky/search?q=%22@ts-expect-error%22

sindresorhus:
@mgcrea Would you be able to fix the tests?

mgcrea:
@sindresorhus indeed, forgot to push a commit fixing the broken test due to the updated dist path, should be ok now.

sindresorhus:
@mgcrea Thanks for working on this 👍

mgcrea:
Thanks for merging such a big PR! Looking forward to contribute with smaller patches in the future!

PR #333, opened 2021-04-04
Improve HTTPError messages

sholladay:
This PR aims to provide a clearer error message to the user when an `HTTPError` occurs, by adding a description and showing both the status code and status text, when available.

Perhaps the best example of how this helps is where the response has an unusual status code, such as 0, but there is no status text, as is the case when HTTP/2 is being used.

Before:
```
HTTPError: 0
```

After:
```
HTTPError: Request failed with status 0
```

Happy to bikeshed the phrasing. Could change "failed with status" to "failed with response status", for example. Also on the fence about "an unknown error" vs "an unknown status" vs "an unknown response error", etc.

### DIFF
--- a/source/errors/HTTPError.ts
+++ b/source/errors/HTTPError.ts
@@ -6,12 +6,12 @@ export class HTTPError extends Error {
 	public options: NormalizedOptions;
 
 	constructor(response: Response, request: Request, options: NormalizedOptions) {
-		// Set the message to the status text, such as Unauthorized,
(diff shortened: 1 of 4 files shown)


sindresorhus:
> HTTPError: Request failed with status 0

Maybe: `HTTPError: Request failed with status code 0`?

sholladay:
> Maybe: `HTTPError: Request failed with status code 0`?

Done.

sholladay:
@sindresorhus I fixed the merge conflicts.

PR #335, opened 2021-04-28
Fix merging of searchParams in create()/extend()

### LINKED ISSUE #324: merging searchParams by ky.create() occurs unnecessary  headers  query string
I want to set default search params by `ky.create()` and set another one at execution-time.
But, this use case occurs  unnecessary  `headers=%5Bobject+Object%5D`  query string.

## minimal example

https://codepen.io/TasukuUno/pen/KKNgNae?editors=0010

```js
const instance = ky.create({
  searchParams: {
    // a: 1
  },
});

const res = await instance.get('/some/api', {
  throwHttpErrors: false,
  searchParams: {
    // b: 2
  },
});

// actual request url: /some/api?headers=%5Bobject+Object%5D
console.log(res.url);
```

edited: here's another example repository I made 3 months ago (and forgotten 😅  )
https://github.com/TasukuUno/repro-ky

sholladay:
Closes #324

This PR fixes how the `searchParams` option is handled in `ky.create()` and `ky.extend()` so that the user's expected search params are used. Previously, we were inadvertently adding a `headers` search param in some cases.

### DIFF
--- a/source/utils/merge.ts
+++ b/source/utils/merge.ts
@@ -52,7 +52,6 @@ export const deepMerge = <T>(...sources: Array<Partial<T> | undefined>): T => {
 				returnValue.headers = headers;
 			}
 		}
-
 	}
 
 	return returnValue;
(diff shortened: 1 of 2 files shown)


sindresorhus:
Unfortunate timing with your PRs. Would you be able to fix the merge conflict?

sholladay:
@sindresorhus fixed the merge conflicts. This and the other two PRs are ready for review.

PR #339, opened 2021-05-04
Remove some `@ts-expect-error`

### LINKED ISSUE #337: Fix TypeScript ignore directives
We have a lot of ignore directives now. We should try to fix the types:

https://github.com/sindresorhus/ky/search?q=%22@ts-expect-error%22&type=

Beraliv:
- [x] fixed in production code
- [x] fixed in tests
- [x] ts:check errors
- [x] tests
- [x] resolved conflicts

But I didn't remove every `@ts-expect-error` as sometimes it's needed (for error validation in test cases)

Fixes #337


source/types/hooks.ts
```diff
@@ -8,7 +8,6 @@ export type BeforeRequestHook = (
 
 export type BeforeRetryState = {
 	request: Request;
-	response: Response;
```
Beraliv, line 11: I'm not sure if response is needed

In docs it's not mentioned, in tests I didn't find a usage too

source/index.ts
```diff
@@ -6,14 +6,17 @@ import type {ky as KyInterface} from './types/ky.js';
...
 import {validateAndMerge} from './utils/merge.js';
 
+type Mutable<T> = {
+	-readonly[P in keyof T]: T[P]
+};
+
 const createInstance = (defaults?: Partial<Options>): KyInterface => {
 	// eslint-disable-next-line @typescript-eslint/promise-function-async
-	const ky = (input: Input, options?: Options) => Ky.create(input, validateAndMerge(defaults, options));
+	const ky: Partial<Mutable<KyInterface>> = (input: Input, options?: Options) => Ky.create(input, validateAndMerge(defaults, options));
```
Beraliv, line 15: So inside we declare it mutable and optional

So we can assign it later

source/index.ts
```diff
@@ -6,14 +6,17 @@ import type {ky as KyInterface} from './types/ky.js';
 import type {Input, Options} from './types/options.js';
 import {validateAndMerge} from './utils/merge.js';
 
+type Mutable<T> = {
+	-readonly[P in keyof T]: T[P]
+};
```
Beraliv, line 11: It's needed for `KyInterface` as it has `readonly stop` which needs to make mutable below (to be able to assign the value)

source/core/constants.ts
```diff
@@ -1,18 +1,23 @@
...
 export const supportsAbortController = typeof globalThis.AbortController === 'function';
 export const supportsStreams = typeof globalThis.ReadableStream === 'function';
 export const supportsFormData = typeof globalThis.FormData === 'function';
 
-export const requestMethods: readonly HTTPMethod[] = ['get', 'post', 'put', 'patch', 'head', 'delete'];
+export const requestMethods = ['get', 'post', 'put', 'patch', 'head', 'delete'] as const;
+
+type cases = [
+	Expect<Equal<typeof requestMethods[number], HTTPMethod>>,
+]
```
Beraliv, line 12: instead of declaring `requestMethods` as `readonly HTTPMethod[]`, you make it readonly as is

and then check here in `cases` that values are exactly the same

source/core/constants.ts
```diff
@@ -1,18 +1,23 @@
...
 
 export const responseTypes = {
 	json: 'application/json',
 	text: 'text/*',
 	formData: 'multipart/form-data',
 	arrayBuffer: '*/*',
 	blob: '*/*'
-};
+} as const;
```
Beraliv, line 20: this is needed to infer key and values for `Object.entries` later

source/core/Ky.ts
```diff
@@ -145,10 +149,9 @@ export class Ky {
 		};
 
 		const isRetriableMethod = ky._options.retry.methods.includes(ky.request.method.toLowerCase());
-		const result = isRetriableMethod ? ky._retry(fn) : fn();
+		const result = (isRetriableMethod ? ky._retry(fn) : fn()) as ResponsePromise;
 
-		for (const [type, mimeType] of Object.entries(responseTypes)) {
-			// @ts-expect-error not sure how to properly type this!
+		for (const [type, mimeType] of Object.entries(responseTypes) as ObjectEntries<typeof responseTypes>) {
```
Beraliv, line 154: the problem here is that `Object.entries` has incorrect `ReturnType`

So what I do:

1. I created `ObjectEntries` helper to infer it (however, it's possible to redefine it globally)
2. `result` as `ResponsePromise` now has all the methods for `type` so we can call it without issues

package.json
```diff
@@ -50,6 +50,7 @@
 	],
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "^1.0.1",
+		"@type-challenges/utils": "^0.1.1",
```
Beraliv, line 53: is needed for `Expect` and `Equal`

however, I can define it without dependency

test/browser.ts
```diff
@@ -157,9 +157,10 @@ test(
 				};
 			});
 		}, server.url);
-		// @ts-expect-error
+		if (typeof error !== 'object') {
+			throw new Error('Expected to have an object error');
+		}
```
Beraliv, line 162: it's validated with `throw` so it's narrowed the `object` now

test/fetch.ts
```diff
@@ -4,9 +4,11 @@ import ky from '../source/index.js';
 test.serial('relative URLs are passed to fetch unresolved', async t => {
 	const originalFetch = globalThis.fetch;
 	globalThis.fetch = async input => {
-		// @ts-expect-error @TODO
+		if (typeof input !== 'object') {
+			throw new Error('Expect to have an object request');
+		}
```
Beraliv, line 9: it's validated with `throw` so it's narrowed the `object` now

source/types/options.ts
```diff
@@ -71,7 +71,7 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 
 	Accepts any plain object or value, which will be `JSON.stringify()`'d and sent in the body with the correct header set.
 	*/
-	json?: unknown;
+	json?: Record<string, unknown>;
```
Beraliv, line 74: Correct me if it can be not only an object

test/hooks.ts
```diff
@@ -270,19 +271,19 @@ test('`afterResponse` hook is called with request, normalized options, and respo
...
-						// @ts-expect-error
 						async (request, options, response) => {
 							if (response.status === 403) {
 								// Retry request with valid token
 								return ky(request, {
 									...options,
 									json: {
-										// @ts-expect-error @TODO
-										...options.json,
+										...(options as Options).json,
```
Beraliv, line 280: we cannot use spread here if it's not an object

if it's valid only in test, let's cast it here

test/hooks.ts
```diff
@@ -270,19 +271,19 @@ test('`afterResponse` hook is called with request, normalized options, and respo
...
 										token: 'valid:token'
 									}
 								});
 							}
+
+							return undefined;
```
Beraliv, line 286: should return something in other branches

test/http-error.ts
```diff
@@ -1,18 +1,18 @@
...
+type Mutable<T> = {
+	-readonly[P in keyof T]: T[P]
+};
+
 function createFakeResponse({status, statusText}: {status?: number; statusText?: string}): Response {
 	// Start with a realistic fetch Response.
-	const response = {...new Response()};
+	const response: Partial<Mutable<Response>> = {...new Response()};
 
-	// Assign these values because the Response() constructor doesn't
```
Beraliv, line 8: There are 2 problems here:

1. it is `readonly`
2. it doesn't accept `undefined`

test/http-error.ts
```diff
@@ -1,18 +1,18 @@
 import test from 'ava';
 import {HTTPError} from '../source/index.js';
 
+type Mutable<T> = {
```
Beraliv, line 4: Let me know if I can move it to some place to reuse it (was used twice in the codebase)

Beraliv:
At the moment cannot run tests

Can you help me here @sindresorhus ?

source/core/Ky.ts
```diff
@@ -145,14 +149,14 @@ export class Ky {
 		};
 
 		const isRetriableMethod = ky._options.retry.methods.includes(ky.request.method.toLowerCase());
-		const result = isRetriableMethod ? ky._retry(fn) : fn();
+		const result = (isRetriableMethod ? ky._retry(fn) : fn()) as ResponsePromise;
```
sindresorhus, line 152: Sidenote: We should also try to reduce use of `as` as it just overrides the type no matter what.

source/types/options.ts
sindresorhus: JSON can have top-level array or primitive value.

test/hooks.ts
sindresorhus: The real issue is that the `options` parameter doesn't have the correct type.

test/http-error.ts
sindresorhus: You can move the type utilities to a `types` file in https://github.com/sindresorhus/ky/tree/main/source/utils

source/core/Ky.ts
```diff
@@ -145,14 +149,14 @@ export class Ky {
...
 
-		for (const [type, mimeType] of Object.entries(responseTypes)) {
-			// @ts-expect-error not sure how to properly type this!
+		for (const [type, mimeType] of Object.entries(responseTypes) as ObjectEntries<typeof responseTypes>) {
 			result[type] = async () => {
 				// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
 				ky.request.headers.set('accept', ky.request.headers.get('accept') || mimeType);
 
+				// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion, @typescript-eslint/non-nullable-type-assertion-style
```
sindresorhus, line 159: Why are you silencing the listing? If it's valid, it would be useful with a code comment about why.

sindresorhus (changes requested):

source/core/Ky.ts
Beraliv: No idea here, somehow `npm run test` throws an error here

source/core/Ky.ts
Beraliv: You anyway do that below, so I moved it here

Is it alright?

test/hooks.ts
Beraliv: What is the expected type here?

source/types/options.ts
Beraliv: So would it be ok to have `PrimitiveType | Record<string, unknown>`?

```ts
type PrimitiveType = number | string | boolean | undefined | null;
```

source/core/Ky.ts
sindresorhus: I mean didn't mean to apply that you added it. Was just saying in general, we should try to reduce the `as` whenever possible as they're often as bad as `@ts-expect-error`.

test/http-error.ts
Beraliv: Moved to source/utils ✅

source/utils/Mutable.ts
```diff
@@ -0,0 +1,3 @@
+export type Mutable<T> = {
```
sindresorhus, line 1: The file should be named `types.ts`. I don't want a separate file for each simple type utility.

Beraliv: yeah, that's working for me

Done ✅

source/core/Ky.ts
Beraliv: Okay, I understand that 👍

test/hooks.ts
Beraliv: Also, should it be connected to the `json`  we have in test?

In this case we need to have Generic type:

It will be `Options<Json extends BasicJson = BasicJson>` instead of `Options`

What do you think?

source/types/options.ts
Beraliv: Btw, what is the expected value after calling `.json()` if you pass primitives?

Here in test case I see that it's string instead of `boolean`:

```ts
test('POST JSON with falsey value', async t => {
	// #222
	const server = await createHttpTestServer({bodyParser: false});
		response.json(await parseRawBody(request));
	});

	const json = false;
	const responseJson = await ky.post(server.url, {json}).json();

	t.deepEqual(responseJson, json.toString());

	await server.close();
});
```

Can you please describe the rules so I will be on the same page?

test/hooks.ts
Beraliv: @sindresorhus I can suggest these changes regarding more stricter mapping of `Json` and the types in hooks and `json`

You can have a look at the changes (I didn't finish it, but you can already see the possible way) – https://github.com/Beraliv/ky/pull/1

wdyt?

source/types/options.ts
sindresorhus: > So would it be ok to have Primitive | Record<string, unknown>?

No, it used to be that, but it caused problems with `interface`s. `git blame` it and you'll find the relevant thread.

test/hooks.ts
sindresorhus: I think `...options.json as Record<string, unknown>` would be fine here. We know it's an object, so it's fine to force cast.

test/hooks.ts
```diff
@@ -340,8 +341,7 @@ test('beforeRetry hook is never called for the initial request', async t => {
 				hooks: {
 					beforeRetry: [
 						({options}) => {
-							// @ts-expect-error @TODO
-							options.headers!.set('unicorn', fixture);
+							(options.headers as Partial<Headers> | undefined)?.set?.('unicorn', fixture);
```
sindresorhus, line 344: Why is this needed? Is the `options.headers` type not correct?

test/hooks.ts
Beraliv: Do you think https://github.com/Beraliv/ky/pull/1 will be too much?

Beraliv:
@sindresorhus I go back to this MR at the end of this week

test/hooks.ts
sindresorhus: I think:

> Instead of ky.get(...).json<IncomingHttpHeaders>() we will have ky.get<IncomingHttpHeaders>(...).json()

is a good idea, but making it extend a JSON type is problematic for many reasons. TS is just not good enough to make it workable. https://github.com/sindresorhus/ky/pull/133#discussion_r282433150

test/hooks.ts
Beraliv: You need to narrow the type to work with it

`options.headers` here can be either of union type element: `Headers | string[][] | Record<string, string>;`

But `set` function exists only in `Headers`, that's why I cast it here

source/core/Ky.ts
Beraliv: Removed it 🔥

test/hooks.ts
Beraliv: Okay, let's try it in the future if interested

Beraliv:
@sindresorhus do you know how to solve test errors?

I have this error message:

```bash
TypeError [ERR_INVALID_RETURN_PROPERTY_VALUE]: Expected string to be returned for the "format" from the "loader getFormat" function but got type object.
```

it happens in several files:

```bash
test/retry.ts
test/browser.ts
test/main.ts
test/hooks.ts
test/headers.ts
test/prefix-url.ts
test/methods.ts
```

I use:

```bash
beraliv@Alexeys-MBP ky % node -v
v12.21.0
beraliv@Alexeys-MBP ky % npm -v
7.11.1
beraliv@Alexeys-MBP ky % yarn -v
1.22.10
```

Also have another error:

```bash
fetch.ts › fetch option takes a custom fetch function

  Planned for 12 assertions, but got 6.
```

I'm just not familiar with `ava` and how it's configured with `node`
Maybe I need something to change?

source/core/Ky.ts
Beraliv: This way it removes one `@ts-expect-error` so that's better than before

I will keep it ✅

test/hooks.ts
sindresorhus: I think `NormalizedOptions` needs to ensure headers are only `Headers`. We should present simple unified types to the user.

sindresorhus:
> do you know how to solve test errors?

Upgrade `ts-node` to latest version.

source/core/Ky.ts
lukaw3d: Sorry to give you merge conflicts. I added `entry as [keyof typeof responseTypes, string]` in 
https://github.com/sindresorhus/ky/pull/355/commits/3e75d74d7b0eb6385d19d6ff1d080eb0f0e7834a#diff-1d28e040cea335f5a22aa9c3f678f5d00b353e3ce966dc0978a91fe56f240b04R151-L152

Beraliv: Good, I will update the PR on the weekend, should be fine 🔥

Beraliv:
> > do you know how to solve test errors?
> 
> Upgrade `ts-node` to latest version.

Updated 👍

sindresorhus:
You need to fix the merge conflicts.

Beraliv:
@sindresorhus rebased 👍

sindresorhus:
https://github.com/sindresorhus/ky/pull/339#discussion_r626105488 was not done, so I moved it back in https://github.com/sindresorhus/ky/pull/339/commits/72de135b55a39f5088dcf9b9ea608b41a50dc695. It has to be `unknown`.

sindresorhus:
Thanks :)

Beraliv:
> [#339 (comment)](https://github.com/sindresorhus/ky/pull/339#discussion_r626105488) was not done, so I moved it back in [72de135](https://github.com/sindresorhus/ky/commit/72de135b55a39f5088dcf9b9ea608b41a50dc695). It has to be `unknown`.

Ah, right, thank you!

Beraliv:
@sindresorhus thank you for the help during development, that was an interesting experience

Issue #340, opened 2021-05-05
Module not found: Can't resolve 'ky' with version 0.28


LoiKos:
Upgrading to ky@0.28, i get `Module not found: Can't resolve 'ky'`.

When i open vs code and look at the 'ky' module it refers to `index.d.ts` and load typescript definition instead of index.js.

I use create-react-app i don't know if that help you but it works perfectly fine with ky@0.27

sholladay:
Hi @LoiKos and thanks for reporting this. I was able to reproduce the issue relatively easily with some educated guesses. That said, in the future, please provide reproduction steps or a sample repository. There are people out there like me who don't use CRA.

Here are the reproduction steps:

1. `npx create-react app foo`
2. `cd foo && npm install ky`
3. Open `src/index.js` and add `import ky from 'ky'`
4. `npm start`

I was able to fix the problem by modifying `node_modules/ky/package.json` to use `main` instead of `exports`, which is slightly confusing because Ky v0.27.0 also uses `exports`, albeit with a different value, and somehow that works. Perhaps the CRA tooling only has partial or buggy support for `exports`? The only difference that stands out to me is that Ky v0.28 puts everything in a `distribution` directory.

Arguably, this is likely a problem with CRA itself. But at the same time, maybe we should consider going back to `main` for ecosystem compatibility.

LoiKos:
Yeah sorry for missing information. I was able to reproduce in codeSandbox : https://codesandbox.io/s/vigilant-hooks-mw7xl?file=/src/App.tsx 
If someone need to see the problem in action

Eric-Dunaway:
having the same problem with a vue-cli app

Todo:
will add steps to reproduce

sholladay:
@LoiKos and @Eric-Dunaway could you please also file an issue on the repos for those tools?

 - CRA: https://github.com/facebook/create-react-app/issues/new/choose
 - Vue CLI: https://github.com/vuejs/vue-cli/issues/new/choose

Once that's done, I'll create a branch of Ky that you can use as a workaround.

719media:
This is a blocker for me to upgrade to the latest ky 0.28.0 (using vue-cli). I don't know the ins and outs of npm package deployment, but when I use over 80+ dependencies directly in my application with vue-cli, and this is the first time such a problem has ever happened to me with any library, seems like perhaps having other vendors "fix" this issue (even if they are in the wrong) may not be the best solution? Just a humble suggestion :)

Repro:
[...]

PR #342, opened 2021-05-07
Restore `main` field in package.json

mgcrea:
`ky@0.28.0` can't be used in react-native apps anymore due to a missing main field in `package.json`.

Metro does provide configuration options to support extra "main" fields but adding `"exports"` there breaks as other packages start to use this field instead of their `"main"` (not sure why, as it is expected to be an ordered check, maybe a metro-related bug of the `"exports"` key handling when in object form). Haven't found a way to make it work except manually patching the ky's package.json.

```
error: Error: While trying to resolve module `ky` from file `~/foo/src/services/pumalocal/client.ts`, the package `~/foo/node_modules/ky/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`~/foo/node_modules/ky/index`. Indeed, none of these files exist:
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
+	"main": "./distribution/index.js",
 	"exports": "./distribution/index.js",
 	"types": "./distribution/index.d.ts",
 	"engines": {


sholladay:
Issue #340 mentions other tools that are also breaking due to lack of `main`.

I'm personally in favor of merging this fix. Though, I'd really like to understand the root cause first. Ky is a valid package and Node's `require()` function can handle our usage of `exports`, so what are these tools doing that's causing problems?

mgcrea:
> I'm personally in favor of merging this fix. Though, I'd really like to understand the root cause first. Ky is a valid package and Node's `require()` function can handle our usage of `exports`, so what are these tools doing that's causing problems?

I think webpack or metro are designed to actively check specific keys in your package.json in a specific, user-configurable order, (eg. `['browser', 'module', 'main']`), one use-case was to enable tree-shaking with the non-official `"module"` key before the `"exports"` field was spec'ed.

sholladay:
Sure, I am aware of that behavior and it's a good point that the problem likely stems from how they detect which key to use. However, I would expect that once they find the `exports` key, they would pass its value to `require()`, which should work. Perhaps the implementation is more complicated than that, though.

sholladay:
As I mentioned in https://github.com/sindresorhus/ky/issues/340#issuecomment-845308284, importing Ky from the unpkg CDN is also broken and I believe it's the same root cause, meaning this will likely fix that, too. It would seem that a lot of the ecosystem is just implementing their own algorithms for finding the main entry point instead of using Node's `require.resolve()` or anything even remotely standardized.

sholladay (approved):

lpellegr:
I experienced the same issue using the last version of ky with a Typescript web app and ESM imports. This PR fixes the issue.

mgcrea:
@sindresorhus any chance you could merge & release quickly this change as it is breaking many setups? Thanks!

sindresorhus:
Sorry for the slow reply. Somehow I got accidentally unsubscribed to the Ky repo.

https://github.com/sindresorhus/ky/releases/tag/v0.28.2

PR #344, opened 2021-05-13
Fix `searchParams` option TypeScript type

### LINKED ISSUE #343: Discrepancy between doc and types for searchparams options in v 0.28
Following the recent refactoring to TS, the types for `PlainOptions.searchParams` became more restrictive than what is in the doc: https://github.com/sindresorhus/ky#searchparams

For us, our codebase fails to type-check when using boolean values in 0.28 (but was passing before):
```ts
ky.get(url, {
  // this fails in v0.28
  searchParams: { val: true },
  })
```

I'm happy to go ahead and update all the places where the compilation fails if it is an API design decision but before I wanted to check that it was intentional or wasn't an oversight?

mgcrea:
I narrowed the type to fix a TypeScript error when the option was passed to the URLSearchParams constructor, but obviously it was a mistake as we don't want any breaking changes. We might want to reconsider in the future to be more compliant with the official types.

I've added a manual `as` to the constructor as a fix.

---

Fixes #343

### DIFF
--- a/source/types/options.ts
+++ b/source/types/options.ts
@@ -2,7 +2,11 @@ import type {LiteralUnion, Required} from './common.js';
 import type {Hooks} from './hooks.js';
 import type {RetryOptions} from './retry.js';
 
+// eslint-disable-next-line unicorn/prevent-abbreviations
 export type SearchParamsInit = string | string[][] | Record<string, string> | URLSearchParams | undefined;
+
+// eslint-disable-next-line unicorn/prevent-abbreviations
+export type SearchParamsOption = SearchParamsInit | Record<string, string | number | boolean> | Array<Array<string | number | boolean>>;
 
 export type HTTPMethod = 'get' | 'post' | 'put' | 'patch' | 'head' | 'delete';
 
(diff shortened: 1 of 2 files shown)


source/types/options.ts
```diff
@@ -2,6 +2,8 @@ import type {LiteralUnion, Required} from './common.js';
 import type {Hooks} from './hooks.js';
 import type {RetryOptions} from './retry.js';
 
+export type SearchParamsInit = string | string[][] | Record<string, string> | URLSearchParams | undefined;
```
sindresorhus, line 5: Isn't this type already available in the DOM lib?

source/types/options.ts
```diff
@@ -99,7 +101,7 @@ export interface Options extends Omit<RequestInit, 'headers'> {
 
 	Accepts any value supported by [`URLSearchParams()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams).
 	*/
-	searchParams?: string[][] | Record<string, string> | string | URLSearchParams;
+	searchParams?: SearchParamsInit | Record<string, string | number | boolean> | Array<Array<string | number | boolean>>;
```
sindresorhus, line 104: The whole type here should be exported as a named type `SearchParamsOption`.

sindresorhus (changes requested):

source/types/options.ts
mgcrea: Done.

source/types/options.ts
mgcrea: Haven't found it: https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L16108

PR #351, opened 2021-05-25
Fix incompatibility with New Relic

### LINKED ISSUE #345: 1ms delay before fetch causing issues with New Relic SPA logger
I found a finicky bug where the implementation detail of ky waiting 1ms before fetch (https://github.com/sindresorhus/ky/pull/118) is actually not allowing New Relic Pro+SPA to log AJAX requests. 

I am not sure of the full detail of why this may be happening - but this is how i have verified it's being caused by 1ms setTimeout.

- I tried using native `fetch` and `wretch` library instead of `ky`. New Relic was able to pick up all AJAX requests for both scenarios.
- I tried using `axios` that relies on XHR requests. New Relic was also able to pick this up.
- I used `ky` but manually commented out the `await delay(1);` from ky inside `node_modules`, and it started working (granted, the `.json` function fails, but the request was getting picked up)

Do you guys have any idea why this may be happening?

If so, do you have suggestions on how to get around this issue?

lukeupup:
Fix #345

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -105,54 +105,43 @@ export class Ky {
 				throw new RangeError(`The \`timeout\` option cannot be greater than ${maxSafeTimeout}`);
 			}
 
-			return new Promise<Response>((resolve, reject) => {
-				// Fix #345: We must keep setTimeout() here instead of using delay(), because new relic cannot handle parallel promise correctly.
-				setTimeout(async () => {
-					try {
-						let response = await ky._fetch();
-						for (const hook of ky._options.hooks.afterResponse) {
-						// eslint-disable-next-line no-await-in-loop
-							const modifiedResponse = await hook(
-								ky.request,
-								ky._options as NormalizedOptions,
-								ky._decorateResponse(response.clone())
-							);
+			await Promise.resolve();
+			let response = await ky._fetch();
 
-							if (modifiedResponse instanceof globalThis.Response) {
-								response = modifiedResponse;
-							}
-						}
+			for (const hook of ky._options.hooks.afterResponse) {
+				// eslint-disable-next-line no-await-in-loop
+				const modifiedResponse = await hook(
+					ky.request,
+					ky._options as NormalizedOptions,
+					ky._decorateResponse(response.clone())
+				);
 
-						ky._decorateResponse(response);
+				if (modifiedResponse instanceof globalThis.Response) {
+					response = modifiedResponse;
+				}
+			}
 
-						if (!response.ok && ky._options.throwHttpErrors) {
-							reject(new HTTPError(response, ky.request, (ky._options as unknown) as NormalizedOptions));
-							return;
-						}
+			ky._decorateResponse(response);
 
-						// If `onDownloadProgress` is passed, it uses the stream API internally
-						/* istanbul ignore next */
-						if (ky._options.onDownloadProgress) {
-							if (typeof ky._options.onDownloadProgress !== 'function') {
-								reject(new TypeError('The `onDownloadProgress` option must be a function'));
-								return;
-							}
+			if (!response.ok && ky._options.throwHttpErrors) {
+				throw new HTTPError(response, ky.request, (ky._options as unknown) as NormalizedOptions);
+			}
 
-							if (!supportsStreams) {
-								reject(new Error('Streams are not supported in your environment. `ReadableStream` is missing.'));
-								return;
-							}
+			// If `onDownloadProgress` is passed, it uses the stream API internally
+			/* istanbul ignore next */
+			if (ky._options.onDownloadProgress) {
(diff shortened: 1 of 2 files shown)


source/core/Ky.ts
```diff
@@ -105,43 +105,54 @@ export class Ky {
...
+							resolve(ky._stream(response.clone(), ky._options.onDownloadProgress));
+							return;
+						}
 
-			return response;
+						resolve(response);
+					} catch (error: unknown) {
+						reject(error);
+					}
+				}, 1);
```
lukeupup, line 154: Didn't change 1 to 0, though 0 looks also work

source/core/Ky.ts
```diff
@@ -105,43 +105,54 @@ export class Ky {
...
-					ky._decorateResponse(response.clone())
-				);
-
-				if (modifiedResponse instanceof globalThis.Response) {
-					response = modifiedResponse;
-				}
-			}
+			return new Promise<Response>((resolve, reject) => {
+				// Fix #345: We must keep setTimeout() here instead of using delay(), because new relic cannot handle parallel promise correctly.
+				setTimeout(async () => {
```
lukeupup, line 110: The major change is just remove the `delay(1)` and perform the delay by a setTimeout.

sholladay (changes requested):
I don't think you need to wrap the `for` loop in the new promise, just the fetch.

sindresorhus:
Please report this to New Relic and then update the code comment with a bug ID.

source/core/Ky.ts
```diff
@@ -105,43 +105,54 @@ export class Ky {
...
-					ky._options as NormalizedOptions,
-					ky._decorateResponse(response.clone())
-				);
-
-				if (modifiedResponse instanceof globalThis.Response) {
-					response = modifiedResponse;
-				}
-			}
+			return new Promise<Response>((resolve, reject) => {
+				// Fix #345: We must keep setTimeout() here instead of using delay(), because new relic cannot handle parallel promise correctly.
```
sindresorhus, line 109: Not sure what you mean by "parallel promise". Would be great if you could be more specific.

sindresorhus:
Can you check if `queueMicrotask()` works?

source/core/Ky.ts
sindresorhus: I don't remember why it's `1` and not `0`.

@sholladay Do you?

sindresorhus: If `0` is fine, we can just remove the argument altogether.

sholladay:
@lukeupup I pushed some changes to make the code a bit cleaner. I'm hoping it will still work but have no way to test it. Can you please check?

lukeupup:
> @lukeupup I pushed some changes to make the code a bit cleaner. I'm hoping it will still work but have no way to test it. Can you please check?

@sholladay Amazing, it works indeed on our environment! Although I don't really understand how new relic treat `delay` and `Promise.resolve` differently... Not sure if @smblee can see this thread, but it would be better to invite him to have some test as well. If it also works, I think we can solve the problem as you've submitted. Thanks a lot!

source/core/Ky.ts
szmarczak: @sindresorhus https://github.com/sindresorhus/ky/pull/118

szmarczak: I'm not sure if Chrome replaces `0` with `1` or no. In that case it wouldn't matter.

szmarczak:
@lukeupup In case you missed this: https://github.com/sindresorhus/ky/pull/351#issuecomment-847874309

smblee:
Just catching up on this thread - we switched to axios for the time being but i'll give this a shot!

sholladay:
My theory is that the bug has something to do with the fact that `delay()` first schedules a callback in the task queue and then after that fires it schedules a callback in the job (microtask) queue. Doing one or the other is normal but combined it's pretty unusual so I figure that may be a situation that New Relic's code isn't handling well. It happens because `delay()` is using _both_ `setTimeout()` and a promise.

 - https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/
 - https://careersjs.com/magazine/javascript-job-queue-microtask/

I don't see a clean way to use `queueMicrotask()` here as the code is async. We'd have to wrap `queueMicrotask()` in a promise, which would effectively queue two microtasks. Seems pointless. That said, I updated the code to use `await Promise.resolve()`, which is very similar to `queueMicrotask()`.

BTW, I also updated AVA to reduce the amount of noisy console logs during testing.

I think this PR is ready to merge, assuming it works for @smblee.

lukeupup:
@sholladay This looks convincing. So let's look forward to the verification result from @smblee . Is there anything I should do before we can merge the code?

smblee:
I can confirm that this definitely addresses it! Thanks!

PR #353, opened 2021-05-31
Fix missing hook method types for TypeScript

Qwerios:
Since version 0.28 the types exported by ky seem to have been reduced. Our codebase relies on ky hooks and we used to be able to import the hook method types from ky in TypeScript.

```typescript
import { AfterResponseHook, BeforeRequestHook } from 'ky';
```

This PR adds the missing exports for the hook methods back to the index including the new retry hook method.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -28,6 +28,6 @@ const ky = createInstance();
 export default ky;
 
 export {Options};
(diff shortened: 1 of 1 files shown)


Qwerios:
Thank you very much for the rapid response, merge and release

Issue #135, opened 2019-05-15
afterResponse hook causes response Promise to hang


exogen:
Here's a simple repro: https://github.com/exogen/ky-afterResponse-repro

This doesn't seem to happen for every request; I suspect it might have something to do with the response body size. I tried a couple small public JSON API demos before I landed on a GraphQL endpoint that reproduced the behavior.

Under certain conditions, simply including an `afterResponse` hook causes the response Promise to hang. The hook doesn't have to do anything; it can be a no-op function or just return the same cloned response. Removing the hook causes it to start working. Switching to `node-fetch` also works fine.

I wonder if the hook causes the response to be cloned and therefore the Promise as well, making one of them never resolve? Just a thought.

sindresorhus:
Duplicate of https://github.com/sindresorhus/ky-universal/issues/8

sindresorhus:
Can you post your issue description above as a comment on https://github.com/sindresorhus/ky-universal/issues/8?

Issue #31, opened 2018-09-17
Ky promise not resolved with large responses when using methods


taylorjdawson:

taylorjdawson:
### The Quest continues...

It looks like the `ky(<url>).json()` is called that `ky` makes 2 requests? Is that correct? 

`ky(<url>)` calls `_fetch()` here: 

https://github.com/sindresorhus/ky/blob/81bbf7b6f6eed796b2f9c711ee96f5f298367818/index.js#L111-L113) 

...and then `.json()` calls `_fetch()` here:

https://github.com/sindresorhus/ky/blob/81bbf7b6f6eed796b2f9c711ee96f5f298367818/index.js#L117-L119

Sorry if I am missing something I am more at intermediate and this stuff is pretty advanced 😉

sindresorhus:
> It looks like the ky(<url>).json() is called that ky makes 2 requests? Is that correct?

You can check that yourself by adding a console.log statement inside `_fetch()` and then do a request or run a single test that does a request.

> ...and then .json() calls _fetch() here:

Notice the `_retryCount` check and that it's inside `this._retry`. It only calls `_fetch()` again if it's retrying.

sindresorhus:
Can you try setting the `retry` option to `0` and see if the problem still exists?

sholladay:
Hi @taylorjdawson! I tried to reproduce this problem using your gist, but all of those checks succeeded for me. I did tweak your gist a tiny bit to account for the fact that I'm running it in a real browser and I took the `ky` source code from https://unpkg.com/ky and assigned it to a variable in the console instead of importing it in a file (mostly out of laziness). So there are some differences, but maybe those differences will help you narrow down the problem.

Here is the full code I ran that you can just copy-paste into your browser console.
<details>
<summary>Click to see code snippet</summary>

```js
const isObject = value => value !== null && typeof value === 'object';

const deepMerge = (...sources) => {
	let returnValue = {};

	for (const source of sources) {
		if (Array.isArray(source)) {
			if (!(Array.isArray(returnValue))) {
				returnValue = [];
			}

			returnValue = [...returnValue, ...source];
		} else if (isObject(source)) {
			for (let [key, value] of Object.entries(source)) {
				if (isObject(value) && Reflect.has(returnValue, key)) {
					value = deepMerge(returnValue[key], value);
				}

				returnValue = {...returnValue, [key]: value};
			}
		}
	}

	return returnValue;
};

const requestMethods = [
	'get',
[...]

PR #356, opened 2021-06-02
Add `avoidCloningResponse` option (useful for large responses)

lukaw3d:
This mitigates recurring issues (https://github.com/sindresorhus/ky/issues/135, https://github.com/sindresorhus/ky/issues/31, https://github.com/sindresorhus/ky-universal/issues/18, https://github.com/sindresorhus/ky-universal/issues/8) caused by node-fetch clone limitations. Included test reproduces large responses getting stuck by default.

##### avoidCloningResponse

Type: `boolean`\
Default: `false`

Don't clone response before `hooks.afterResponse`, `onDownloadProgress`, and method shortcuts.

Useful to reduce browser memory consumption when fetching large responses. And in NodeJS, node-fetch doesn't get stuck if response is larger than highWaterMark.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -303,6 +303,17 @@ Throw an `HTTPError` when, after following redirects, the response has a non-2xx
 
 Setting this to `false` may be useful if you are checking for resource availability and are expecting error responses.
 
+
+##### avoidCloningResponse
+
+Type: `boolean`\
+Default: `false`
+
+Don't clone response before `hooks.afterResponse`, `onDownloadProgress`, and method shortcuts.
+
+Useful to reduce browser memory consumption when fetching large responses. And in NodeJS, node-fetch doesn't get stuck if response is larger than highWaterMark.
+
+
 ##### onDownloadProgress
 
 Type: `Function`
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -36,6 +36,7 @@ export class Ky {
 			prefixUrl: String(options.prefixUrl || ''),
 			retry: normalizeRetryOptions(options.retry),
 			throwHttpErrors: options.throwHttpErrors !== false,
+			avoidCloningResponse: options.avoidCloningResponse ?? false,
 			timeout: typeof options.timeout === 'undefined' ? 10000 : options.timeout,
 			fetch: options.fetch ?? globalThis.fetch.bind(globalThis)
 		};
@@ -114,7 +115,7 @@ export class Ky {
 				const modifiedResponse = await hook(
 					ky.request,
 					ky._options as NormalizedOptions,
-					ky._decorateResponse(response.clone())
+					ky._options.avoidCloningResponse ? response : ky._decorateResponse(response.clone())
 				);
 
 				if (modifiedResponse instanceof globalThis.Response) {
@@ -139,14 +140,17 @@ export class Ky {
 					throw new Error('Streams are not supported in your environment. `ReadableStream` is missing.');
 				}
 
-				return ky._stream(response.clone(), ky._options.onDownloadProgress);
+				return ky._stream(
+					ky._options.avoidCloningResponse ? response : response.clone(),
+					ky._options.onDownloadProgress
+				);
 			}
 
 			return response;
 		};
 
 		const isRetriableMethod = ky._options.retry.methods.includes(ky.request.method.toLowerCase());
-		const result = isRetriableMethod ? ky._retry(fn) : fn();
+		const result = (isRetriableMethod ? ky._retry(fn) : fn()) as ResponsePromise;
 
 		for (const [type, mimeType] of Object.entries(responseTypes)) {
 			// @ts-expect-error not sure how to properly type this!
@@ -154,7 +158,7 @@ export class Ky {
 				// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
 				ky.request.headers.set('accept', ky.request.headers.get('accept') || mimeType);
 
-				const response = ((await result) as Response).clone();
+				const response = ky._options.avoidCloningResponse ? await result : (await result).clone();
 
 				if (type === 'json') {
 					if (response.status === 204) {
@@ -171,7 +175,7 @@ export class Ky {
 			};
 		}
 
-		return result as ResponsePromise;
+		return result;
 	}
 
 	protected _calculateRetryDelay(error: Error) {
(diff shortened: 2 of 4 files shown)


sindresorhus:
> Useful to reduce browser memory consumption when fetching large responses.

Are you sure this is correct? I would have assumed browsers use copy-on-write, so it only doubles if the copy is modified.

sindresorhus:
> Included test reproduces large responses getting stuck by default.

Did you see https://github.com/sindresorhus/ky-universal#clone-hangs-with-a-large-response-in-node---what-should-i-do ?

sindresorhus:
And did you see [this](https://github.com/sindresorhus/ky-universal/issues/8#issuecomment-509789901) comment?

lukaw3d:
> > Useful to reduce browser memory consumption when fetching large responses.
> 
> Are you sure this is correct? I would have assumed browsers use copy-on-write, so it only doubles if the copy is modified.

I had a streaming usecase in mind. So the main data reader doesn't hold it all in memory, but any clones must - until read.

To reproduce, serve 100mb file:
```sh
dd if=/dev/urandom of=100mb bs=1024 count=102400
npx http-server
open http://127.0.0.1:8080/
```
run in browser console:
```js
response = await fetch('http://127.0.0.1:8080/100mb');
// clone = response.clone();
reader = response.body.getReader();
receivedLength = 0;
while(true) {
  const {done, value} = await reader.read();
  if (done) break;
  receivedLength += value.length;
}
console.log(receivedLength);
// Tab memory footprint 25MB => 29MB without clone
// Tab memory footprint 21MB => 177MB with clone
```

---

> > Included test reproduces large responses getting stuck by default.
> 
> Did you see https://github.com/sindresorhus/ky-universal#clone-hangs-with-a-large-response-in-node---what-should-i-do ?

Yes; I assume that is meant to be:
```diff
  const response = await ky('https://example.com', {
    // 20 MB
-   highWaterMark: 1000 * 1000 * 20
+   fetch: (url, opts) => fetch(url, { highWaterMark: 1000 * 1000 * 20, ...opts }),
  });
```
That makes it work for responses up to ~20MB (my test is 50MB just to be sure), but gets stuck if response is larger than highWaterMark. So to stream-read 1GB file, I'd need to set highWaterMark at 1GB so unused clone can buffer it all up and not pause the main pipe.

---

> And did you see [this](https://github.com/sindresorhus/ky-universal/issues/8#issuecomment-509789901) comment?

Avoiding methods that `response.clone()` internally seems too shaky. And since removing `.clone()` internally removes some safety, making it optional seemed like a better idea

sholladay:
Personally, I'm against adding this option. In my view, cloning the response is an implementation detail. If the clone method doesn't work well, then we either shouldn't use it or we should advocate for upstream changes that will work better. Namely, the browser can and should share memory between the original and cloned responses. And `node-fetch` should fix the hanging bug. I'm not a fan of adding an option to control memory usage from JavaScript land, and I suspect few people would use it. I also don't like that this option only applies to the response. The same issues presumably apply to the request.

lukaw3d:
> If the clone method doesn't work well

When downloading a massive file to multiple clones, I think the only ways it can work are:
- If all clones are being exhausted in parallel it works. Browser downloads a chunk and all clones consume it.
- If one clone starts reading with a delay, it works. Browser downloads a chunk and most clones consume it, but one buffers up to highWaterMark. Then either
  - [**node-fetch behavior**] browser pauses downloading. After delay the clone consumes highWaterMark-buffer and browser resumes downloading.
  - [**browse/chrome behavior**] browser continues and buffers unread chunks (above highWaterMark). After delay the clone consumes highWaterMark-buffer and consumes browser-buffer.
- If one clone never starts reading. Then either:
  - [**node-fetch behavior**] browser waits at highWaterMark in case clone later starts reading
  - [**browse/chrome behavior**] browser buffers the whole massive file in case clone later starts reading
  - [alternative] browser just continues reading and if clone doesn't start reading before highWaterMark it will miss chunks

Not great options, but only an issue if size > highWaterMark

---

> I also don't like that this option only applies to the response. The same issues presumably apply to the request.

Oh, good point! `request.clone()` is always called. Here's what I found with node-fetch

- Uploading a stream-inside-FormData: doesn't clone, works.
  ```js
  test('large FormData requests dont get stuck when cloned', async t => {
    const server = await createHttpTestServer();
    server.post('/', async (_request, response) => {
      response.end();
    });
    t.timeout(10000);
    const formData = new FormData();
    formData.append('file', fs.createReadStream('./test/100mb', {encoding: 'binary', highWaterMark: 1024 * 1024}));
    await t.notThrowsAsync(
      ky.post(server.url, {body : formData as any}).text()
    );
    await server.close();
  });
  ```
  ```js
  this._options.fetch(this.request.clone());
  // `new Request(` and `request.clone()` call `clone(input)` https://github.com/node-fetch/node-fetch/blob/ffef5e3c2322e8493dd75120b1123b01b106ab23/src/request.js#L62
  // but clone ignores FormData https://github.com/node-fetch/node-fetch/blob/ffef5e3c2322e8493dd75120b1123b01b106ab23/src/body.js#L250-L252
  // so doesn't get stuck, because there is no unread pipe clone.

  this._options.fetch(this.request.clone());
  await this._options.fetch(this.request.clone());
  // doesn't get stuck, reading from same stream instance in parallel.

  await this._options.fetch(this.request.clone());
  await this._options.fetch(this.request.clone());
  // gets stuck without error, I assume because stream already exhausted.
  ```

- Uploading a stream: clones, gets stuck.
  ```js
  test.failing('large stream requests get stuck if cloned', async t => {
    const server = await createHttpTestServer();
    server.post('/', async (_request, response) => {
      response.end();
    });

    t.timeout(10000);
    const stream = fs.createReadStream('./test/100mb', {encoding: 'binary', highWaterMark: 1024 * 1024});
    await t.notThrowsAsync(
      ky.post(server.url, {
        body: stream as any,
        headers: {
          'Content-Type': 'application/octet-stream'
        }
      }).text()
    );
    await server.close();
  });
  ```
  ```js
  this._options.fetch(this.request.clone);
  // and even
  this._options.fetch(this.request);
  // gets stuck because `new Request(` immediately clones https://github.com/node-fetch/node-fetch/blob/ffef5e3c2322e8493dd75120b1123b01b106ab23/src/request.js#L62
  // and gets stuck at highWaterMark. Would only work if all pipe clones were exhausted in parallel.
  ```

sindresorhus:
I also have this hope that https://github.com/Ethan-Arrowood/undici-fetch will soon be stable and we can use it in `ky-universal` instead of node-fetch.

PR #358, opened 2021-06-07
Update instructions for use without a bundler

sholladay:
This PR aims to fix the usage instructions and CDN import URL for those without a bundler, which PR #330 broke.

CDN users need to import `distribution/index.js` instead of `index.js`.

I also changed the CDN host from jsdelivr to unpkg, which isn't strictly necessary so long as the path is updated, but jsdelivr seems to be buggy. It's incorrectly caching `index.js` from previous versions and returning it, even though that file doesn't exist anymore.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -537,11 +537,11 @@ Either use a test runner that can run in the browser, like Mocha, or use [AVA](h
 
 #### How do I use this without a bundler like Webpack?
 
-Upload the [`index.js`](index.js) file in this repo somewhere, for example, to your website server, or use a CDN version. Then import the file.
+Make sure your code is running as an ES module, for example by using a `<script type="module">` tag in your HTML document. Then Ky can be imported directly by that module without a bundler or other tools.
 
 ```html
 <script type="module">
-import ky from 'https://cdn.jsdelivr.net/npm/ky@latest/index.js';
+import ky from 'https://unpkg.com/ky/distribution/index.js';
 
 const json = await ky('https://jsonplaceholder.typicode.com/todos/1').json();
 
--- a/readme.md
+++ b/readme.md
@@ -537,7 +537,7 @@ Either use a test runner that can run in the browser, like Mocha, or use [AVA](h
 
 #### How do I use this without a bundler like Webpack?
 
-Make sure your code is running as an ES module, for example by using a `<script type="module">` tag in your HTML document. Then Ky can be imported directly by that module without a bundler or other tools.
+Make sure your code is running as a JavaScript module (ESM), for example by using a `<script type="module">` tag in your HTML document. Then Ky can be imported directly by that module without a bundler or other tools.
 
 ```html
 <script type="module">


readme.md
```diff
@@ -537,11 +537,11 @@ Either use a test runner that can run in the browser, like Mocha, or use [AVA](h
 
 #### How do I use this without a bundler like Webpack?
 
-Upload the [`index.js`](index.js) file in this repo somewhere, for example, to your website server, or use a CDN version. Then import the file.
+Make sure your code is running as an ES module, for example by using a `<script type="module">` tag in your HTML document. Then Ky can be imported directly by that module without a bundler or other tools.
```
sindresorhus, line 540: I don't think `ES module` is commonly used. How about this?

```suggestion
Make sure your code is running as a JavaScript module (ESM), for example by using a `<script type="module">` tag in your HTML document. Then Ky can be imported directly by that module without a bundler or other tools.
```

MartinKolarik:
Hi @sholladay, Martin from jsDelivr here.

> It's incorrectly caching index.js from previous versions and returning it, even though that file doesn't exist anymore.

This is actually a feature - when people use links without a specified version, they tend to break over time as files move around. We're able to detect this and serve the latest available version *where the file exists*, to prevent breaking existing websites. This is one of many production use-focused features unpkg doesn't have 😉

szmarczak:
> without a specified version

@MartinKolarik The specified version here is `latest`.

> We're able to detect this and serve the latest available version where the file exists, to prevent breaking existing websites.

But this is still breaking. Example: `main.js` uses `hello.js` but a new version got out: `main.js` was renamed to `index.js` and `hello.js` was rewritten so it is not compatible with the old version. Since it's pointing to the latest file available, `main.js` will use the newer `hello.js` file which is breaking.

Please correct me if I'm wrong.

MartinKolarik:
> The specified version here is latest.

Yes, maybe a better word would be "specific" instead of "specified". We simply consider all versions that match the requested range.

> But this is still breaking. Example: main.js uses hello.js but a new version got out: main.js was renamed to index.js and hello.js was rewritten so it is not compatible with the old version. Since it's pointing to the latest file available, main.js will use the newer hello.js file which is breaking.

That's correct - this fallback doesn't fix every case but fixes many of them (particularly when the file used in the browser is bundled - there's just one file == no room for the described problem).

MartinKolarik:
Now that I think about it, we probably could respond with a redirect to the older version instead, that would solve the dependency problem and also make it more obvious what's going on.

szmarczak:
> we probably could respond with a redirect to the older version instead

Well, it depends on whether the clients accept redirects or not. Another solution would be to accept a search param to disable cache. E.g. `?cache=0`

sholladay:
What if the removed file has security vulnerabilities or was removed for other important reasons? The CDN should respect updates.

This reminds me of how if you ask the Skypack CDN for a specific version of React (and only React), they'll just ignore it and serve you the latest version (v17 at time of writing). Try it: https://cdn.skypack.dev/react@16

They, too, [justify that as a feature](https://github.com/skypackjs/skypack-cdn/issues/88). But it has caused a lot of problems for myself and others because React 17 doesn't work in Deno yet. As a result, one day my [Pogo](https://github.com/sholladay/pogo) web server framework and all apps built with it just stopped working, even though I had pinned my dependencies. I had to switch to the jspm CDN to fix it.

I strongly believe that we as a community need to break the habit of creating software with such magical behavior. It's the author's responsibility to follow semver, come up with a migration strategy between versions, and maintain backward compatibility where possible. And it's the user's responsibility to choose a specific version or version range if they want stability instead of updates. Aside from that, there are more generalized solutions to the broken links problem, including systems like [IPFS](https://ipfs.io/) and sites like the [Wayback Mahine](https://archive.org/web/).

PR #383, opened 2021-10-12
Export KyInstance

piitaya:
To avoid using `typeof ky`, we can use the `KyInstance` type.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -27,6 +27,8 @@ const ky = createInstance();
 
 export default ky;
 
+export type {KyInterface};
+
 export {
 	Options,
 	NormalizedOptions,


sindresorhus:
How is this better than just doing `typeof ky` which is shorter and doesn't require another import?

piitaya:
Some people (like me 😅) like to use build-in library types instead of inferring them but no problem if you don't think it's necessary 🙂

sindresorhus:
I think it's better to limit the exported types to a minimum. A lot of random types can overwhelm the user. Since the user can already use `typeof`, I'm going to pass on this.

ZachHaber:
I'd like you to reconsider exporting the KyInstance type. Typescript complains about the lack of KyInstance export when using `"composite":true` in the tsconfig:

```
src/index.ts:3:14 - error TS2742: The inferred type of 'api' cannot be named without a reference 
to './node_modules/ky/distribution/types/ky.js'. This is likely not portable. A type annotation 
is necessary.

3 export const api = ky.extend({});
```

It's possible to work around this by using `export const api: typeof ky = ky.extend({})`. Unfortunately, that fix is very much so non-intuitive when you are first encountering this error, and typescript itself doesn't make it easy to figure out the root cause of the error as the incremental builds appear to hide errors.

sindresorhus:
That sounds more like a TypeScript bug if it only happens with `"composite":true`. Can you confirm that if you add this line to the `ky` package in your `node_modules`, it starts working for you?

ZachHaber:
Yes, it starts working when I add the export:
image
image

> That sounds more like a TypeScript bug if it only happens with `"composite":true`.


Honestly, I started looking into seeing if it was a bug in VSCode at first because I wasn't seeing the error in the console. While I was trying to set up a reproducible test case, I found out typescript itself was also showing the error. Eventually I figured out that tsconfig.tsbuildinfo swallows the error and doesn't display it again until something significant changes, or you delete the file.

I started looking into if it's an issue with TS, and I was reading the documentation on composite which states 

> All implementation files must be matched by an [include](https://www.typescriptlang.org/tsconfig#include) pattern or listed in the [files](https://www.typescriptlang.org/tsconfig#files) array. If this constraint is violated, tsc will inform you which files weren’t specified.

That lead me to here finding this PR though one of the weirdly similar duplicates of this one. At that point, I tried changing the `index.d.ts` file and realized that, yup, the error disappears the moment I add in the export.

The real bug in typescript is probably that the incremental builds don't display the error when re-running 🤦‍♂️

sindresorhus:
https://github.com/sindresorhus/ky/releases/tag/v1.1.3

PR #384, opened 2021-10-12
Fix types for TypeScript 4.4

piitaya:
The catch error type is `unknown` instead of `any` with Typescript 4.4. https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables


piitaya:
You right, `const ms = Math.min(this._calculateRetryDelay(error), maxSafeTimeout);` act as a typeguard for the unknown error so we can safety use throw an `Error` in the hook.

PR #396, opened 2021-11-05
Fix source map generation

zengguirong:


sholladay (approved):
LGTM. I'm curious, is it common for TypeScript libraries to include inline sources? It can significantly increase bundle size. Though, i care more about DX than bundle size.

zengguirong:
> LGTM. I'm curious, is it common for TypeScript libraries to include inline sources? It can significantly increase bundle size. Though, i care more about DX than bundle size.

The inline `sourcesContent` is append to the `.map` files instead of the `.js` files. I think it's not important to increase the size of these source map files.

When source map is provided but missing sourceConent or source files, in traditional bundler like `Webpack`, it will fallback to the uglified file itself(I don't understand how it realize) and it is fairly acceptable. But with `Vite`, I can't entirely debug with the source because the `sourceContent` isn't provided in the `.map` file, and the browser will try to get the source files declared in `file` field which are also missing.

According to my experience，there are usually  3 ways to deal with source map:
1. No source map, when you files are not uglified, usually using by bundlers, such as [antd/es](https://www.jsdelivr.com/package/npm/antd?path=es%2Fimage)
2. Source map with sourceContent, such as [lit-html](https://www.jsdelivr.com/package/npm/lit-html)
3. Source map with source files, such as [underscore](https://www.jsdelivr.com/package/npm/underscore) (the source folder is `modules` which is included in the published package)

sholladay:
You're right, I mistakenly thought it would inline the source map and the source content, but those are separate options. Cool, this looks like the right approach to me.

PR #416, opened 2022-01-04
Json return empty string id server responds with a 202 status code

ionutcirja:


source/core/Ky.ts
```diff
@@ -70,7 +70,7 @@ export class Ky {
 				const response = (await result).clone();
 
 				if (type === 'json') {
-					if (response.status === 204) {
+					if (response.status === 202 || response.status === 204) {
```
sholladay, line 73: If I'm not mistaken, in the case of a 202 response, this will work as expected if the response does not contain a body, but it will drop the body if one is present.

test/main.ts
```diff
@@ -266,6 +266,22 @@ test('.json() with 204 response and empty body', async t => {
 	await server.close();
 });
 
+test('.json() with 202 response and empty body', async t => {
```
sholladay, line 269: Can you also add tests for 202s with a body, one with valid JSON and one with invalid JSON (e.g. plain text)?

sholladay (changes requested):

sindresorhus:
@ionutcirja Friendly bump

PR #418, opened 2022-01-24
feat: export KyInstance as type

charlie632:
sometimes you want to have a KyInstance declared somewhere, like:

```ts
import ky from 'ky'

class Requester {
  api: typeof KyInstance

  constructor() {
     this.api = ky.create({...});
  }
}
```

but sometimes you only want the type reference, not the actual `ky` module imported, so this:

```ts
import type { KyInstance } from 'ky'
```

might help


sindresorhus:
Duplicate of #383

sindresorhus:
Does this work?

```ts
import type ky from 'ky';

//...

typeof ky
```

charlie632:
Fair enough. Thanks!

PR #429, opened 2022-03-02
Allow overriding `Content-Type` header for JSON requests

### LINKED ISSUE #369: Allow custom content-type header to be set
For some reason Amazon REST api's require the content-type to be set to "application/x-amz-json-1.1" instead of the standard "application/json"

Currently the content-type header is set automatically for json data overriding any headers passed in to the request.

It would be useful to first check if the header is set before overriding it:
```
if (!this.request.headers.set('content-type')) {
    this.request.headers.set('content-type', 'application/json');
}
```

sholladay:
Closes #369

This PR adds logic to ensure that `options.headers` takes precedence over our own default for the `Content-Type` header when the `json` option is used. This allows the user to send a JSON request with a content type other than `application/json`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -527,19 +527,18 @@ const response = await ky.post(url, {body: searchParams});
 
 ### Setting a custom `Content-Type`
 
-Ky automatically sets an appropriate
-[Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
-header for each request based on the data in the request body. However, some
-APIs require custom, non-standard content types, such as
-`application/x-amz-json-1.1`. Using the
-`headers` option, you can manually override the content type.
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -525,6 +525,27 @@ searchParams.set('drink', 'icetea');
...
+
+Ky automatically sets an appropriate
+[Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
+header for each request based on the data in the request body. However, some
+APIs require custom, non-standard content types, such as
+`application/x-amz-json-1.1`. Using the
+`headers` option, you can manually override the content type.
+
+```js
+import ky from "ky";
```
sindresorhus, line 538: Use single-quotes

sholladay:
Fixed the formatting issues. Need to work on my VS Code auto formatting settings...

Issue #390, opened 2021-10-26
Several retry problems


xmedeko:
Retry algorithm  has several problems:

* [x] Retry `delay(ms)` wait is not abortable by `this._options.signal` , see https://github.com/sindresorhus/ky/blob/2049affe2ca939f53350fb9b7354c3ff8d2e1c1f/source/core/Ky.ts#L228
  * Check [sindresorhus/delay](https://github.com/sindresorhus/delay) 😉 
* [x] If `Retry-After` is 0, i.e. "retry immediately", then the request is not retried at all. (The problem is that 0 returned from `_calculateRetryDelay` is treaded as "stop retry").
* [x] IMO if `Retry-After` is above `maxRetryAfter`, then it should be capped to `maxRetryAfter` and not that the retry is stopped. See https://github.com/sindresorhus/ky/blob/2049affe2ca939f53350fb9b7354c3ff8d2e1c1f/source/core/Ky.ts#L194

ruscon:
Perhaps it can be closed?

PR #433, opened 2022-03-24
Fix abortions between retries

Richienb:
#390


Richienb:
I still need to modify the tests before this is ready

Richienb:
Before, abortions would essentially be postponed until `window.fetch` would run, at which point it would throw. But now, the abortion is fast-tracking before the delay between retries completes. This means we'll need to throw our own error which causes a mismatch in the unit tests.

Should we replicate the fetch abortion error that was previously thrown or continue rolling our own custom error message?

sindresorhus:
I think we should try to replicate it. See how we did it here: https://github.com/sindresorhus/p-retry/pull/59/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R38-R40

source/errors/DOMException.ts
```diff
@@ -0,0 +1,10 @@
+// TODO: Use `??` instead of '||' when targeting Node.js 14
+const DOMException = globalThis.DOMException || Error;
+
+export default DOMException;
```
Richienb, line 4: The other files export a class but this time it isn't actually used outside of this file. Not sure if the function name is specific enough either. What do you think?

sindresorhus: It's fine 👍

source/utils/p-defer.ts
```diff
@@ -0,0 +1,48 @@
+// https://github.com/sindresorhus/p-defer/tree/f83ba3bb302f67d6e86395142167a9e49cb0b8f9
+
+export interface DeferredPromise<ValueType> {
+	/**
+	The deferred promise.
+	*/
+	promise: Promise<ValueType>;
+
+	/**
+	Resolves the promise with a value or the result of another promise.
```
sindresorhus, line 10:
```suggestion
	Resolves the promise with a value or the result of another promise.

```

sindresorhus:
Can you add a TODO comment about using `.signalThrowIfAborted` when targeting Node.js 18: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted

sindresorhus:
This is still marked as draft, is there any pending work?

Richienb:
I don't think so but the build is failing and I don't know why

sindresorhus:
```
Uncaught exception in test/browser.ts
[23](https://github.com/sindresorhus/ky/runs/6109551543?check_suite_focus=true#step:5:23)

[24](https://github.com/sindresorhus/ky/runs/6109551543?check_suite_focus=true#step:5:24)
  test/helpers/with-page.ts(2,32): error TS2614: Module '"ava"' has no exported member 'UntitledMacro'. Did you mean to use 'import UntitledMacro from "ava"' instead?
```

Follow what the error says?

sindresorhus:
There's a new error now: https://github.com/sindresorhus/ky/runs/7428874132?check_suite_focus=true

Richienb:
Should be working now, try rerunning CI

sindresorhus:
Thanks :)

PR #437, opened 2022-04-20
Fixes #312 Add typed json response

### LINKED ISSUE #312: [Feature request] Typing JSON response using TS generics
I'm considering a move from `axios` to `ky` in order to leverage the Fetch API, which seems to be relatively easy because of the clear APIs and documentation for them in ky. 🙂 
However I'm running into a limitation of the types, where I would like to be able to pass a generic for the Response type.

This is currently [possible in Axios](https://github.com/axios/axios/blob/59ab559386273a185be18857a12ab0305b753e50/index.d.ts#L140) and I would like to keep the same syntax, rather than having to resort to typecasting (i.e. `await response.json() as MyResponse`).

Is this something you would consider adding support for as well?

TeraNovaLP:
Hey,

to fix the issue, I implemented an axios like solution with an extended response object. With this solution, there should be no breaking changes either.

Fixes #312


sindresorhus:
This needs a type test to prove that your change actually resolves the issue.

TeraNovaLP:
@sindresorhus I'm not quite sure how to do this. I don't see a way to automatically check the type when they are gone at runtime. Sorry if I overlook something right now. 😅

TeraNovaLP:
@sindresorhus Sorry to bother you, but I can't continue without an answer to my previous question.

sindresorhus:
You should test at compile-time. The Ky tests are in TypeScript, so you can use type assertions. For example, using https://github.com/mmkal/ts/tree/master/packages/expect-type

PR #441, opened 2022-06-08
Fix Response details are lost when using onDownloadProgress

### LINKED ISSUE #394: Response details are lost when using onDownloadProgress
As discussed in https://github.com/sindresorhus/ky/discussions/393, using the `onDownloadProgress` callback causes Ky to return an incorrect response, which always has a status code of 200 and without the received response headers. This bug has existed since the introduction of `onDownloadProgress` in PR https://github.com/sindresorhus/ky/pull/34.

Looking at the code, it's easy to see that we're creating a new `Response` without passing through any headers or status code, etc.

https://github.com/sindresorhus/ky/blob/2049affe2ca939f53350fb9b7354c3ff8d2e1c1f/source/core/Ky.ts#L275-L309

promer94:
close #394

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -167,7 +167,7 @@ export class Ky {
 			// To provide correct form boundary, Content-Type header should be deleted each time when new Request instantiated from another one
 			if (
 				((supportsFormData && this._options.body instanceof globalThis.FormData)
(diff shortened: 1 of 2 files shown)


promer94:
Screen Shot 2022-06-10 at 19 04 15


Not sure why test failed at node 14

PR #451, opened 2022-09-01
fix(#450): Duplex Request param needed in Chrome@105

### LINKED ISSUE #450: Chrome@105 Failed to construct Request The duplex member must be specified
Chrome@105 Failed to construct Request The duplex member must be specified.

> From Chromium 105, you can start a request before you have the whole body available by using the Streams API.

https://developer.chrome.com/articles/fetch-streaming-requests/
screenshot_998

Tested on Ky@0.28.5 and lates Ky@0.31.1

It seems that 
```
duplex: 'half'
```

option is missing.

karol-f:
Fixes #450

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -156,6 +156,7 @@ export class Ky {
 		this.request = new globalThis.Request(this._input as RequestInfo, this._options as RequestInit);
 
 		if (supportsStreams) {
+			// @ts-expect-error - Types are outdated.
 			this.request.duplex = 'half';
 		}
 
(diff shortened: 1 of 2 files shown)


jkhaui (approved):

jkhaui:
Tested on version 105.0.0.0 and older versions of Chrome, looks to have fixed the issue with [requests with stream bodies](https://developer.chrome.com/articles/fetch-streaming-requests/#half-duplex).

wwan5803 (approved):
Tested on chrome Version 105.0.5195.52 (Official Build) (x86_64), working as expected

PR #454, opened 2022-09-14
Add retry.backoffLimit option

### LINKED ISSUE #389: Customize retry backoff algorithm
Current backoff  algorithm is `0.3 * (2 ** (retry - 1)) * 1000` that does not suit for all. E.g. I may like to change some coefficient or add a random jitter (see [AWS Exponential Backoff And Jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)). Solution may be simple - just allow custom function to calculate the delay.

fvanwijk:
I saw in the source code that the retry backoff function is hardcoded and exponential. However, I wanted to customize it.
Fortunately I also found the open issue addressing this limitation.

To start simple, I added a `backoffLimit` so that you can cap the delay that is exponentially growing. After 5 or so retries the delay becomes so large that the user will give up anyway.
Setting the limit to for example 1000 in combination with 60 retries will change the behavior into polling for 1 minute (my use case).

I had some trouble testing the delay because the calculate function is private and I didn't find a way to extract this information from the `ky` instance or response.
So I did some perf measurements to test the delays.

If you are happy with the PR so far, I will implement the `calculateDelay` option too.

Fixes #389

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "editor.formatOnSave": false
-}
--- a/readme.md
+++ b/readme.md
@@ -169,6 +169,7 @@ Default:
 - `methods`: `get` `put` `head` `delete` `options` `trace`
 - `statusCodes`: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`500`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)
 - `maxRetryAfter`: `undefined`
+- `backoffLimit`: `undefined`
 
 An object representing `limit`, `methods`, `statusCodes` and `maxRetryAfter` fields for maximum retry count, allowed methods, allowed status codes and maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time.
 
@@ -176,7 +177,7 @@ If `retry` is a number, it will be used as `limit` and other defaults will remai
 
 If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`. If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
-Delays between retries is calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1).
+Delays between retries is calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1). If `backoffLimit` is set, it will maximize the delay to that number so that it won't increase further after many retries.
 
 Retries are not triggered following a [timeout](#timeout).
 
@@ -187,7 +188,8 @@ const json = await ky('https://example.com', {
 	retry: {
 		limit: 10,
 		methods: ['get'],
-		statusCodes: [413]
+		statusCodes: [413],
+		backoffLimit: 3000
 	}
 }).json();
 ```
--- a/readme.md
+++ b/readme.md
@@ -177,8 +177,9 @@ If `retry` is a number, it will be used as `limit` and other defaults will remai
 
 If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`. If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
-By default, the delay is calculated with `0.3 * (2 ** (attemptCount - 1)) * 1000`. The delay increases exponentially.
+The `backoffLimit` option is the upper limit of the delay per retry in milliseconds.
 To clamp the delay, set `backoffLimit` to 1000, for example.
+By default, the delay is calculated with `0.3 * (2 ** (attemptCount - 1)) * 1000`. The delay increases exponentially.
 
 Retries are not triggered following a [timeout](#timeout).
 
--- a/readme.md
+++ b/readme.md
@@ -177,7 +177,8 @@ If `retry` is a number, it will be used as `limit` and other defaults will remai
 
 If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`. If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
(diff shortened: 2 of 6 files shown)


source/types/retry.ts
```diff
@@ -33,4 +33,16 @@ export interface RetryOptions {
 	@default Infinity
 	*/
 	maxRetryAfter?: number;
+
+	/**
+	The upper limit of the `computedValue`.
```
sindresorhus, line 38: This should describe the option in simpler higher-level terms without mentioning `computedValue`.

source/types/retry.ts
```diff
@@ -33,4 +33,16 @@ export interface RetryOptions {
...
+
+	By default, the computedValue is calculated in the following way:
+
+	0.3 * (2 ** (attemptCount - 1)) * 1000
+
+	The delay increases exponentially.
+	In order to prevent this, you can set this value to a fixed value, such as 1000.
+	 */
```
sindresorhus, line 46:
```suggestion
	*/
```

sindresorhus (changes requested):

sindresorhus:
CI is failing

source/types/retry.ts
fvanwijk: I copied from `got` a bit too literally as there is no `computedValue` field in the `RetryOptions` type, while in [got there is one](https://github.com/sindresorhus/got/blob/f6d1fb71b5e107ef4a6e3b5d5dfa2a80ec3f30a5/documentation/7-retry.md#calculatedelay).

I now rephrashed the comment.

fvanwijk:
Thanks for the review.

I must say that testing the delay of the retries a bit hard. You probably also know this, because the delay wasn't initially tested anyway.
Currently the test depends on a quick execution time, but if you run the test on a slow machine it might fail. In my opinion no test is better than a flaky test, so we should delete it.
Or are there ways to test the delay that I don't know of?

sindresorhus:
Maybe check `process.env.CI` and increase the number a bit then?

sindresorhus:
Maybe check `process.env.CI` and increase the number a bit then?

sindresorhus:
You also need to add the docs to the readme.

readme.md
```diff
@@ -169,14 +169,15 @@ Default:
...
+- `backoffLimit`: `undefined`
 
 An object representing `limit`, `methods`, `statusCodes` and `maxRetryAfter` fields for maximum retry count, allowed methods, allowed status codes and maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time.
 
 If `retry` is a number, it will be used as `limit` and other defaults will remain in place.
 
 If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`. If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
-Delays between retries is calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1).
+Delays between retries is calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1). If `backoffLimit` is set, it will maximize the delay to that number so that it won't increase further after many retries.
```
sindresorhus, line 180: I don't think "maximise" is the correct term here. Maybe "clamp".

source/types/retry.ts
```diff
@@ -33,4 +33,19 @@ export interface RetryOptions {
...
+
+	```
+	0.3 * (2 ** (attemptCount - 1)) * 1000
+	```
+
+	The delay increases exponentially.
+
+	In order to prevent this, you can set this value to a fixed value, such as 1000.
+	*/
+	backoffLimit?: number;
```
sindresorhus, line 50: The docs for TS and readme should be in sync as much as possible. It makes it hard to maintain if they differ significantly.

sindresorhus (changes requested):

readme.md
fvanwijk: Thought the same :)

fvanwijk (commented):
I updated the tests to use the Node 14 Performance API (instead of LTS).

By the way, it would be nice if CI runs automatically on push. Now the feedback cycle is pretty slow.

sindresorhus:
> By the way, it would be nice if CI runs automatically on push.

I agree, but GitHub forces me to push this button each time.

Screenshot 2022-10-20 at 17 31 22

sholladay:
> GitHub forces me to push this button each time

That can be changed in the repository settings under Actions > General

image

sindresorhus:
@sholladay I'm aware, and I did for this repo. But I have 1000+ repos. I cannot do it manually on all of them.

source/types/retry.ts
```diff
@@ -33,4 +33,19 @@ export interface RetryOptions {
 	@default Infinity
 	*/
 	maxRetryAfter?: number;
+
+	/**
+	The upper limit of the delay per retry in milliseconds.
```
sindresorhus, line 38: This sentence still needs to be added to the readme.

source/types/retry.ts
```diff
@@ -33,4 +33,19 @@ export interface RetryOptions {
...
+
+	By default, the delay is calculated in the following way:
+
+	```
+	0.3 * (2 ** (attemptCount - 1)) * 1000
+	```
+
+	The delay increases exponentially.
+
+  To clamp the delay, set `backoffLimit` to 1000, for example.
```
sindresorhus, line 48: Incorrect indentation and I would put it on line 39.

test/retry.ts
```diff
@@ -440,3 +442,60 @@ test('throws when retry.statusCodes is not an array', async t => {
...
+	// Register observer that asserts on duration when a measurement is performed
+	const obs = new PerformanceObserver(items => {
+		const measurements = items.getEntries();
+
+		const duration = measurements[0].duration ?? Number.NaN;
+		const expectedDuration = {default: 300 + 600 + 1200 + 2400, custom: 300 + 600 + 1000 + 1000}[measurements[0].name] ?? Number.NaN;
+
+		t.true(Math.abs(duration - expectedDuration) < allowedOffset, `Duration of ${duration}ms is not close to expected duration ${expectedDuration}ms`); // Allow for 300ms difference
+
+		if(measurements[0].name === 'custom') {
```
sindresorhus, line 473:
```suggestion
		if (measurements[0].name === 'custom') {
```

sindresorhus (changes requested):

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
```
sindresorhus, line 3: Don't add unrelated things.

source/types/retry.ts
```diff
@@ -33,4 +33,20 @@ export interface RetryOptions {
...
+
+	 By default, the delay is calculated in the following way:
+
+	 ```
+	 0.3 * (2 ** (attemptCount - 1)) * 1000
+	 ```
+
+	 The delay increases exponentially.
+
+	 @default Infinity
```
sindresorhus, line 49: Incorrect indentation.

fvanwijk:
It looks like the new test that depends on timing makes the test suite flaky. I doubt whether it is a good idea to keep te test. Removing it altogether might be the right thing to do.

.vscode/settings.json
fvanwijk: Oops, had to ignore the file. Now you know why the formatting issues keep hunting me ;)

fvanwijk:
@sindresorhus In the mean time there haven't been any activity for about a month. Any suggestions how to follow up? Changes needed or complete the PR?

sindresorhus:
Thanks. Sorry for the delay. I must have missed the notification.

fvanwijk:
This PR had flaky tests, so I increased the test timeout offsets. Maybe it is fixed in #484, so we could decrease (set to 0?) the [allowedOffset](https://github.com/sindresorhus/ky/pull/454/files#diff-057f1cac975e0d13462dc0b8af0a28b3435beb1d59652a11b20a180026546b70R471).

Issue #375, opened 2021-08-27
Accept empty responses on `json` response type


szmarczak:
Got: https://github.com/sindresorhus/got/blob/295979693bb179b32d11107367c57c31cb05bf78/source/core/response.ts#L140-L142

### Discussed in https://github.com/sindresorhus/ky/discussions/374

<div type='discussions-op-text'>

<sup>Originally posted by **cruzms** August 27, 2021</sup>

Hi i'm getting `SyntaxError: JSON Parse error: Unexpected EOF`, after receiving an empty response from server with a status 200.
The only way to changing this behavior is changing responses from server to no content(204) ?
Loggin my response with afterResponse hook i got this
``` json
{
  "type":"default",
  "status":200,
  "ok":true,
  "statusText":"",
  "headers":{
    "map":{
      "content-length":"0",
      "connection":"keep-alive",
      "date":"Thu, 26 Aug 2021 22:37:33 GMT",
      "keep-alive":"timeout=5"
    }
  },
  "url":"http://localhost/test",
  "bodyUsed":false,
  "_bodyInit":{
    "_data":{
      "size":0,
      "offset":0,
      "blobId":"c4a5f299-c8f6-46bf-84e3-4ed7a981b5b2",
      "__collector":{
      }
    }
  },
  "_bodyBlob":{
    "_data":{
      "size":0,
      "offset":0,
      "blobId":"c4a5f299-c8f6-46bf-84e3-4ed7a981b5b2",
      "__collector":{
      }
    }
  }
}
```</div>

matei-radu:
According to the HTTP/1.1 specifications  ([RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3)), `200 OK` responses have a payload, while `204 No Content` don't. With this in mind, I believe ky is behaving correctly.

However, I can also understand that many APIs and developers do not respect this distinction (or maybe they're lazy :P) so, spec complaint or not, maybe we need this feature anyway.

szmarczak:
You missed this part:

> an origin server MAY generate a payload body of zero length.
> If no payload is desired, an origin server ought to send 20

ought to =/= must

matei-radu:
I stand corrected, thanks for digging that up.

matei-radu:
I stand corrected, thanks for digging that up.

sindresorhus:
[...]

PR #459, opened 2022-10-12
Parse empty response bodies without throwing

### LINKED ISSUE #415: Parse empty response bodies without throwing
In case of an 202 there is no way for the HTTP to later send an asynchronous response indicating the outcome of processing the request. 
When trying to call the json method we will get Unexpected end of JSON input.

### LINKED ISSUE #430: Json return empty string if server responds with a 202 status code
https://github.com/sindresorhus/ky/pull/416

lemonadern:
## Summary
Following the discussion in this Issue( #415 ), I changed the `.json()` behavior.
When response body is empty, `.json()` now returns an empty string (`''`) instead of throwing an exception(`FetchError: Invalid JSON`).
>  I was thinking in terms of whether it's reasonable for a given status code to have an empty body. But I'm sure there are plenty of implementations out there that return an empty body with a 200 or other success codes. Maybe we should just allow that for convenience sake.
https://github.com/sindresorhus/ky/issues/415#issuecomment-1007626969

> When Content-Length is 0 or is missing, body must be returned as an empty string.
https://github.com/sindresorhus/ky/issues/415#issuecomment-1228718284

## Issues
Close #415 
Close #430 

### Added
- json() with empty response body
- json() with Invalid JSON response body ( plain text )
### Removed
- .json() with 200 response and empty body

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -96,7 +96,7 @@ The `input` and `options` are the same as [`fetch`](https://developer.mozilla.or
 - The `credentials` option is `same-origin` by default, which is the default in the spec too, but not all browsers have caught up yet.
 - Adds some more options. See below.
 
-Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, an appropriate `Accept` header will be set depending on the body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range of `200...299`. Also, `.json()` will return an empty string if the response status is `204` instead of throwing a parse error due to an empty body.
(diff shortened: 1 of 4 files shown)


lemonadern:
Another related issue? #375

sindresorhus:
This behavior also needs to be documented.

lemonadern:
I think this CI failure is flaky one. My source code passed these tests on my local machine and forked environment.
https://github.com/sindresorhus/ky/actions/runs/3321613592/jobs/5507852198#step:5:19

lemonadern:
I think this CI failure is flaky one. My source code passed these tests on my local machine.
https://github.com/sindresorhus/ky/actions/runs/3321613592/jobs/5507852198#step:5:19

readme.md
```diff
@@ -96,7 +96,7 @@ The `input` and `options` are the same as [`fetch`](https://developer.mozilla.or
...
+Returns a [`Response` object](https://developer.mozilla.org/en-US/docs/Web/API/Response) with [`Body` methods](https://developer.mozilla.org/en-US/docs/Web/API/Body#Methods) added for convenience. So you can, for example, call `ky.get(input).json()` directly without having to await the `Response` first. When called like that, an appropriate `Accept` header will be set depending on the body method used. Unlike the `Body` methods of `window.Fetch`; these will throw an `HTTPError` if the response status is not in the range of `200...299`. Also, `.json()` will return an empty string if body is empty or the response status is `204` instead of throwing a parse error due to an empty body.
```
sindresorhus, line 99: You also need to document this in the TS doc comment.

lemonadern: OK, I'll add the same one to the TS doc.

Issue #463, opened 2022-10-31
In v0.32.0, `.json()` returns an empty string if "Transfer-Encoding: chunked" header is present (and thus "Content-Length: [...]" header is absent)


davidrunger:
This bug was introduced in https://github.com/sindresorhus/ky/pull/459 (cc @lemonadern, for your interest).

davidrunger:
^ there is linked a PR (https://github.com/sindresorhus/ky/pull/464) that should fix this bug

Nickhoyer:
@davidrunger @sindresorhus Compressed Cloudflare responses come without Content-Length and without Transfer-Encoding

https://community.cloudflare.com/t/no-content-length-header-when-response-is-compressed/134475

davidrunger:
@Nickhoyer Good catch. That is a problem.

My PR #464 fixed the bug that was failing my app's test suite... and so now that v0.32.1 has passed my test suite, it's in my app out in production behind Cloudflare, and I can confirm that I am indeed seeing this error (`Uncaught (in promise) TypeError: t.forEach is not a function`) on the client-side in production now, because Cloudflare is indeed providing neither a `Content-Length` nor a `Transfer-Encoding` header.

It would probably be good to come up with a solution ASAP, before too many people ship this "occurs only in production (behind Cloudflare)" bug to users. 😬

davidrunger:
> It seems to me that a request can provide a content-encoding header for a stream instead of a transfer-encoding header for chunks

So basically the idea would be to add another conditional, then?, i.e.:

```js
const contentLength = response.headers.get('Content-Length');
const contentEncoding = response.headers.get('Content-Encoding');
const transferEncoding = response.headers.get('Transfer-Encoding');

const responseIsEmpty =
  (contentLength === null || contentLength === '0') &&
    contentEncoding === null &&
    transferEncoding !== 'chunked'

if (responseIsEmpty) {
  return '';
}
```

Are we confident that would have the desired effect?, i.e. will `Content-Encoding` only be present if the response is not empty?

If we're not confident of that, then the best other solution that I can think of in the meantime would be to get rid of this code entirely for now (i.e. revert both #464 and #459).

davidrunger:
[...]

PR #464, opened 2022-10-31
Don't return empty string for .json() if Transfer-Encoding is chunked

davidrunger:
fixes sindresorhus/ky#463

Here is some relevant documentation about the `Transfer-Encoding` header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding , specifically when a `chunked` value is set for that header:

> `chunked`: Data is sent in a series of chunks. The [`Content-Length`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) header **is omitted in this case** [...] [emphasis added]

This is in contradiction to an implicit assumption seemingly made in https://github.com/sindresorhus/ky/pull/459 that a `Content-Length` header will always be set to some non-zero value if there is content in the response body.

The logic change in this PR returns a blank string for `.json()` only if the `Content-Length` header is `'0'` or missing (per #459) **AND** if there is not a `Transfer-Encoding` header with a value of `chunked`.

The test added in this PR sets a `Transfer-Encoding: chunked` header (and omits a `Content-Length` header) via the following setup, which writes to the response in chunks rather than all at once:

```js
server.get('/', async (request, response) => {
  response.write('[');
  response.write('"one",');
  response.write('"two"');
  response.end(']');
});
```

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -92,7 +92,8 @@ export class Ky {
 					}
 
 					const contentLength = response.headers.get('Content-Length');
-					if (contentLength === null || contentLength === '0') {
+					const transferEncoding = response.headers.get('Transfer-Encoding');
+					if ((contentLength === null || contentLength === '0') && transferEncoding !== 'chunked') {
 						return '';
 					}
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for fixing :)

PR #465, opened 2022-10-31
Verify absent Content-Encoding header before returning "" for json

### LINKED ISSUE #463: In v0.32.0, `.json()` returns an empty string if "Transfer-Encoding: chunked" header is present (and thus "Content-Length: [...]" header is absent)

davidrunger:
fixes #463 (second attempt)

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -92,8 +92,15 @@ export class Ky {
 					}
 
 					const contentLength = response.headers.get('Content-Length');
+					const contentEncoding = response.headers.get('Content-Encoding');
 					const transferEncoding = response.headers.get('Transfer-Encoding');
-					if ((contentLength === null || contentLength === '0') && transferEncoding !== 'chunked') {
+
+					const responseIsEmpty
+					= (contentLength === null || contentLength === '0')
+						&& contentEncoding === null
+						&& transferEncoding !== 'chunked';
+
+					if (responseIsEmpty) {
 						return '';
 					}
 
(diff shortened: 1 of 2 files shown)


davidrunger:
(Not yet ready for merge; need to get linter/tests passing.)

davidrunger:
The linter and tests (apart from the "1 known failure") are now passing. (However, I still have not added a new test covering this change, and might never.)

davidrunger:
I'm currently deploying this branch to production for my application to (hopefully) confirm that it resolves the bug that is introduced by Cloudflare's header alteration.

davidrunger:
Having deployed this branch in my app to production (behind Cloudflare) just now, I can confirm that this does indeed fix the bug (returning an empty string `""` instead of the parsed JSON response body) that emerges as a result of Cloudflare's header alterations as surfaced by @Nickhoyer [here](https://github.com/sindresorhus/ky/issues/463#issuecomment-1297227055).

davidrunger:
I finally managed to add a test.

davidrunger:
I'm going to close this. For the reasons mentioned [here](https://github.com/sindresorhus/ky/issues/463#issuecomment-1297395043) and [here](https://github.com/sindresorhus/ky/issues/463#issuecomment-1297468129), I don't think that continuing to try to analyze the response headers in order to figure out whether the response has body content is going to be a reliable or maintainable solution.

PR #484, opened 2023-01-12
Fix flaky browser tests

cristobal:
The browser tests were flaky due to the reason that the `server.close()` call was slow sometimes due to open connection kept alive, which is default server behaviour.  Resulting in false pending tests assertiions that are actually finished but never finish and times out due to that the server is not closed within the expected timeout.

## What's done
* Added middleware that ensures that the connections are closed after response is sent (@see stackoverflow [here](https://stackoverflow.com/a/13554590/620479))
* Added `beforeEach`  and `afterEach` methods that setups and closes server for each test.
* All the tests are also now serial which is needed since we only have one server that is setup and teardown per test as mentioned above (also playwright is not that good with parallell running browsers either way).
* Enabled the github workflow matrix to also run for node versions 16 and 18.
* Fixed import of `UntitledMacro` that has been renamed to `Implementation`.

## NOTES:
* Tried to avoid adding the `beforeEach` and `afterEach` methods that setups and closes the server. Unless done this way the browser tests  keep beeing flaky 🤷🏽 .

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,8 @@ jobs:
       matrix:
         node-version:
           - 14
+          - 16
+          - 18
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,9 +10,9 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for fixing that 👍

cristobal:
> Thanks for fixing that 👍

Thanks for creating ky 👋🏽

PR #499, opened 2023-04-08
importing index.ts in readme.md deno example

guy-borderless:
for deno I think it's better to use the more idiomatic index.ts export from https://deno.land

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -110,7 +110,7 @@ console.log(json);
 If you are using [Deno](https://github.com/denoland/deno), import Ky from a URL. For example, using a CDN:
 
 ```js
-import ky from 'https://esm.sh/ky';
+import ky from "https://deno.land/x/ky/source/index.ts";
 ```
 
 ## API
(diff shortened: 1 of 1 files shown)


sindresorhus:
Did you test that this works?

readme.md
```diff
@@ -110,7 +110,7 @@ console.log(json);
 If you are using [Deno](https://github.com/denoland/deno), import Ky from a URL. For example, using a CDN:
 
 ```js
-import ky from 'https://esm.sh/ky';
+import ky from "https://deno.land/x/ky/source/index.ts";
```
sindresorhus, line 113: Single-quotes

I would also prefer to keep the esm.sh example below the new Deno.land one.

guy-borderless: > Did you test that this works?

😬

sindresorhus: I don't know. I don't use Deno.

guy-borderless: ok thanks, it doesn't work.

sholladay: At some point, they changed the backend architecture of deno.land so that in order to publish new releases of Ky on there, we have to set up a webhook on our repository, whereas before they had some magic to watch for GitHub releases automatically.

We just need to set up this webhook and then importing from deno.land will work again.

PR #529, opened 2023-09-19
Export type KyInstance

adematte:
Thanks for this great lib 🙏 

This is a small PR to export KyInstance to improve typescript support. 

This is especially useful when using `ky.create(...)` to init a ky instance with new defaults in a class constructor. Without KyInstance type being exported it's impossible to type a class property holding the ky instance.

```ts
class Foo {
    #kyClient:KyInstance // can't be typed without exporting KyInstance

    constructor() {
        this.#kyClient = ky.create({
            headers: { 'PRIVATE-TOKEN': 'XXXXXXX' }
        })
    }
 // etc...
}
```


sindresorhus:
Duplicate of https://github.com/sindresorhus/ky/pull/383

adematte:
> Duplicate of #383

makes perfect sense 👌

Issue #509, opened 2023-05-31
Support retry delay with jitter?


rileytomasek:
Thanks for creating ky! I used it extensively in my libraries and projects.

I'm running into an issue where I'm making a lot of calls in parallel to an API with a rate limit and the 429 retries are all happening at the same time and triggering the rate limit again. I've tested with a forked version of Ky and adding a random jitter to the retry time makes things much better.

I'm happy to contribute a PR, but wanted to see which, if any, solutions you are open to:
1. Add a random jitter (and potentially a way to control it) to the [retry calculation here](https://github.com/sindresorhus/ky/blob/780e0321a2010697c64b2f8391e99d3d447228b3/source/core/Ky.ts#L239)
2. Allow passing a custom function to replace `BACKOFF_FACTOR * (2 ** (this._retryCount - 1)) * 1000`
3. Make it possible to do this from the `beforeRetry` hook?

For more context, what I tested was calculating a jitter factor like `JITTER_FACTOR * randomBetween(1, -1)` and then multiplying the existing retry time by that. It could be configured so that the default value is 1, so there is no change to existing retry time calculation.

sindresorhus:
Yes, I'm open to all of this. 👍

sindresorhus:
@rileytomasek Only number 2 of these requests have been done.

rileytomasek:
@sindresorhus can’t number 1 be accomplished by adding jitter to the delay function? I’m not familiar with the beforeRetry hook to know if you could do it there, but is it required if you have 1 and 2?

sindresorhus:
It can yes, but I think it could still be useful to have it built-in.

sholladay:
Any thoughts on how to do `3`?

We could allow `beforeRetry` to return the ms to delay. But that means the hooks must run before the delay rather than after it, which is arguably a breaking change and IMO less intuitive.

sindresorhus:
> We could allow beforeRetry to return the ms to delay. But that means the hooks must run before the delay rather than after it, which is arguably a breaking change and IMO less intuitive.

It could potentially apply to the next retry instead. You don't really need a jitter for the first retry.

Alternatively, skip this and only do a jitter function. Something like:

```js
retry: {
	delay: 1000,
	jitter: delay => delay * (0.8 + Math.random() * 0.4)
}
```

PR #533, opened 2023-09-29
Add retry.delay option to control the time between retries

rclarey:
Related to option (2) to #509 

### Changes
- add `delay` function to `RetryOptions` object
- add test for using `delay`

### DIFF
--- a/source/types/options.ts
+++ b/source/types/options.ts
@@ -138,7 +138,7 @@ export interface Options extends Omit<RequestInit, 'headers'> { // eslint-disabl
 
 	If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`. If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
-	Delays between retries is calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1).
+	By default delays between retries are calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1), however this can be changed by passing a `delay` function.
 
 	Retries are not triggered following a timeout.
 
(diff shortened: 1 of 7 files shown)


rclarey:
@sindresorhus @szmarczak @sholladay
Sorry, can you please take a look when you have a chance

source/types/options.ts
```diff
@@ -138,7 +138,7 @@ export interface Options extends Omit<RequestInit, 'headers'> { // eslint-disabl
 
 	If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`. If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
-	Delays between retries is calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1).
+	By default delays between retries are calculated with the function `0.3 * (2 ** (retry - 1)) * 1000`, where `retry` is the attempt number (starts from 1), however this can be changed by passing a `delay` function.
```
sindresorhus, line 141:
```suggestion
	By default, delays between retries are calculated with the function `0.3 * (2 ** (attemptCount - 1)) * 1000`, where `attemptCount` is the attempt number (starts from 1), however this can be changed by passing a `delay` function.
```

test/retry.ts
```diff
@@ -499,3 +499,46 @@ test('respect maximum backoff', async t => {
...
+			limit: retryCount,
+			delay: n => 100 * (n + 1),
+		},
+	}).text(), fixture);
+	performance.mark('end');
+
+	performance.measure('linear', 'start', 'end');
+
+	await server.close();
+});
```
sindresorhus, line 544: I feel like the performance mesurement code could be abstracted a bit so we don't have so much duplication between the tests.

sindresorhus (changes requested):

sindresorhus:
You need to document it in the readme too.

TommySorensen:
FYI: @sindresorhus 
After the changes in this PR, then i'm seeing warnings in NextJS like this on all my endpoints.
```
 ⚠ fetch for https://myspeficisomeurl.com on /cart specified "cache: default" and "revalidate: 60", only one should be specified.
```

I'm using ky pretty much basic. But it seems like that KY internally is adding the cache to the fetcher with value of default.
```js
const contentApi = ky.create({
  prefixUrl: `${process.env.CONTENT_API_URL}/api`,
  retry: 0,
});


contentApi
    .get(`getSomething/${someId}`, {
      next: { revalidate: 60 },
    })
    .json();
```

Issue #531, opened 2023-09-23
Ky does not pass custom options down to the native fetch


thexpand:
When I try to pass down custom options to the native fetch, e.g. Next.js 13's `{ next: { revalidate: 0 } }`, these options don't get passed to the `fetch` function. Here's the issue created in Next.js's repo, but the problem is that actually ky doesn't pass the custom options:
https://github.com/vercel/next.js/issues/48519

References about how to use the options in the Next.js docs:
- https://nextjs.org/docs/app/building-your-application/caching#fetch-optionsnextrevalidate
- https://nextjs.org/docs/app/building-your-application/caching#fetch-optionsnexttags-and-revalidatetag

kylewardnz:
I'm not sure if this is related, but I'm also using Next 13.5.2 and a simple post request:
```ts
const data = await ky
    .post(endpoint, {
      json: {
        some: 'json',
      },
    })
    .json<SomeResponseType>();
```

results in:
```
Request with GET/HEAD method cannot have body.
```

Manually specifying `{ method: "POST" }` doesn't work either, so perhaps it's just not passing options at all?

sindresorhus:
This may because Ky is not passing options directly to `fetch(…, options)`, but instead constructs a `Request` and passes the options in there. My guess would be that Next.js does not correctly handle that.

sindresorhus:
Actually, yeah, this is the reason: https://github.com/sindresorhus/ky/issues/516#issuecomment-1674303251

spaceemotion:
I just came across the same issue when i tried to set the request priority. is this something ky should natively support (i can open issue for that, if needed) or left for after the option passthrough has been fixed?

Issue #516, opened 2023-07-13
Full compatibility with Bun


doroved:
Hi, it looks like ky is working in Bun ([bun.sh](https://bun.sh)), you need support for Bun fetch types, for example Bun implements the ability to use a proxy in a request
```js
const response = await fetch(targetUrl, {
    proxy: `http://${proxyUsername}:${proxyPassword}@${proxyHost}:${proxyPort}`,
  })
```
But it doesn't work through ky. Please adapt the library for Bun
```js
const response = await ky(targetUrl, {
    proxy: `http://${proxyUsername}:${proxyPassword}@${proxyHost}:${proxyPort}`,
  })
```

sindresorhus:
Ky just passes the option through to `Fetch`:

> The input and options are the same as [fetch](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch), with some exceptions:

sholladay:
Just a hunch... does it work if you don't use .json()? Ky does a bit of extra processing on the options if you use .json().

sholladay:
What I should have said is to use response.json() instead of the ky().json() shortcut. They are different.

sholladay:
Your code example does this:

```js
const json = await ky(targetUrl, { proxy }).json();
console. log(json);
```

^^ that is what we refer to as the "json shortcut". It's a shortcut because you are calling `.json()` on a promise, which is normally not possible. Promises, including the one returned by `fetch()`, do not have a `.json()` method. But `ky()` returns a special promise that has an added `.json()` method for convenience.

Compare that with `fetch()`, notice how you first have to `await` its promise for a response before calling `.json()` on the response and then you have to `await` that too, since `json()` returns another promise.

```js
const response = await fetch(targetUrl, { proxy });
console. log(await response.json());
```

Ky's json shortcut is optional. You can still the await the response and call its `.json()` method, just as you do with `fetch()`. Here's what it looks like using Ky without the json shortcut:

```js
const response = await ky(targetUrl, { proxy });
console. log(await response.json());
```

In theory, there should not be much difference between all of the above constructs. But Ky does do a bit of extra work when you use the json shortcut, including adding an `Accept` header to the request.

I would like to know if using Ky without the json shortcut changes the result of your test, so that we can rule that out as a cause.

sholladay:
Okay, I think I know what's happening here.

[...]

PR #536, opened 2023-10-17
Pass unknown options to fetch

kdelmonte:
This PR is implements option number 2 of [this comment](https://github.com/sindresorhus/ky/issues/516#issuecomment-1674303251) by @sholladay.

> Ky starts passing unknown options directly to fetch instead of relying on the Request class to do it for us. This is the right thing for Ky to do but might prove to be a little tricky. We would need to either pass all options to fetch, which could inadvertently override some of our options processing logic that's done with the Request class, or have some kind of filtering system to only pass unknown options to fetch, which would presumably involve some kind of hardcoded list of known options (which would have to include all options that the Request class understands) and that doesn't sound fun.

Fixes:
- #531
- #516


source/core/Ky.ts
```diff
@@ -294,11 +295,13 @@ export class Ky {
 			}
 		}
 
+		const requestInit = findUnknownOptions(this.request, this._options);
```
sholladay, line 298: Maybe name this `nonRequestOptions` or `unhandledOptions`?

sholladay (approved):
Great implementation. It's about as simple as you could make it, given the silly dance we have to do here.

I was originally thinking we would maintain a list of known request options to use for filtering, which I really did not want to do. But I see that the request class exposes most (all?) of its constructor options as instance properties and you are using that to avoid hardcoding the list. Nice work.

Technically, we could allow the Ky options to be passed on to `fetch()` and it likely wouldn't cause any trouble, at least in browsers. Although, there could be environments where `fetch()` throws for unexpected options. So doing it this way is certainly safer. 👍

kdelmonte:
This PR caused #539.

kdelmonte:
@doroved , please create a separate issue and provide all the details of what you think needs to be done.

loxK:
> This PR caused #539 which is being resolved with #540.

A couple of hours of debugging an app to finally find out that regression....

Issue #539, opened 2023-10-26
1.1.1 breaks multipart form boundary


etfz:
Multipart boundary value in `Content-Type` header does not match that of the body. Works in 1.1.0.

```js
let formData = new FormData()
formData.append('name', '5')
return ky.post(`/api/user`, {
	body: formData,
})
```

```
POST /api/user HTTP/2
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/118.0
Content-Type: multipart/form-data; boundary=---------------------------1565952755359404003007232716
Content-Length: 174

-----------------------------350127340439298904923059468489
Content-Disposition: form-data; name="name"

5
-----------------------------350127340439298904923059468489--
```

kdelmonte:
I'll take a look.

spaceemotion:
Came across the same issue when a user reported file uploads no longer working in Firefox (which is via S3). Chrome still works however (and all other Chromium alternatives).

sholladay:
So this issue only happens on Chromium?

We have a [test for this](https://github.com/sindresorhus/ky/blob/dd713b6212b185366009d2429347796e6bae97e1/test/browser.ts#L325-L355) that is currently passing.

At the moment, our browser tests only run against Chromium. However, we have the ability to test other browsers as well. We should probably add Firefox [here](https://github.com/sindresorhus/ky/blob/dd713b6212b185366009d2429347796e6bae97e1/test/helpers/with-page.ts#L10-L12). Just need to replace `playwright-chromium` with `playwright` and then `import { firefox } from 'playwright'`.

kdelmonte:
Thanks for the guidance @sholladay. I can confirm this test indeed fails along with 4 others when using `import { firefox } from 'playwright'`.

<img width="856" alt="image" src="https://github.com/sindresorhus/ky/assets/900997/472febc1-546f-44df-a185-1b970e990cbb">

This test passes in 1.1.0 although I did find that some others do not:

<img width="856" alt="image" src="https://github.com/sindresorhus/ky/assets/900997/65c3109b-c756-4bda-bdec-058dbee86eff">

I am investigating now.

sindresorhus:
https://github.com/sindresorhus/ky/releases/tag/v1.1.2

PR #540, opened 2023-10-27
Fix multipart boundary mismatch

kdelmonte:
Turns out that the read-only `body` property of the Request interface [is not [yet] supported by Firefox](https://developer.mozilla.org/en-US/docs/Web/API/Request/body#browser_compatibility) which causes #539.

Changes:

- Add functionality to run same test in multiple browsers including Firefox (⚠️ slows down CI by about 1 minute; ~15% slower)
- Check [Request constructor options](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#options) explicitly.

Fixes: #539 (introduced in #536)

### DIFF
--- a/source/core/constants.ts
+++ b/source/core/constants.ts
@@ -1,5 +1,6 @@
 import type {Expect, Equal} from '@type-challenges/utils';
 import {type HttpMethod, type KyOptionsRegistry} from '../types/options.js';
+import {type RequestInitRegistry} from '../types/request.js';
 
 export const supportsRequestStreams = (() => {
 	let duplexAccessed = false;
@@ -59,19 +60,20 @@ export const kyOptionKeys: KyOptionsRegistry = {
 	fetch: true,
 };
 
-export const requestOptions = new Set([
-	'options',
-	'method',
-	'headers',
-	'body',
-	'mode',
-	'credentials',
-	'cache',
-	'redirect',
-	'referrer',
-	'referrerPolicy',
-	'integrity',
-	'keepalive',
-	'signal',
-	'priority',
-]);
+export const requestOptionsRegistry: RequestInitRegistry = {
+	method: true,
+	headers: true,
+	body: true,
+	mode: true,
+	credentials: true,
+	cache: true,
+	redirect: true,
+	referrer: true,
+	referrerPolicy: true,
+	integrity: true,
+	keepalive: true,
+	signal: true,
+	window: true,
+	dispatcher: true,
+	duplex: true,
+};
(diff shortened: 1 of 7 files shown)


kdelmonte:
@sholladay you were correct [in your comment](https://github.com/sindresorhus/ky/pull/536#pullrequestreview-1694206514) when you questioned if all options were being exposed as instance properties. Please let me know what you think about this change.

source/core/constants.ts
```diff
@@ -58,3 +58,20 @@ export const kyOptionKeys: KyOptionsRegistry = {
 	onDownloadProgress: true,
 	fetch: true,
 };
+
+export const requestOptions = new Set([
```
sindresorhus, line 62: Is there any other way than having a hard-coded list here? We'll have to keep this up to date and if we miss something, it could cause problems.

kdelmonte: I really want to avoid this too. Let me give it some thought.

kdelmonte: Maybe a similar solution to the `KyOptionsRegistry` can be employed here to defer the list keeping to Typescript and undici. I will try this.

kdelmonte: So there are two RequestInit interfaces that are relevant here: the one exported from node's `undici-types` and the one exported from Typescript's `lib.dom.d.ts`.


### Common fields present in both interfaces:
- `method`
- `keepalive`
- `headers`
- `body`
- `redirect`
- `integrity`
- `signal`
- `credentials`
- `mode`
- `referrer`
- `referrerPolicy`
- `window`

### Unique to `undici-types`:
- `dispatcher`
- `duplex`

### Unique to Typescript's `lib.dom.d.ts`:
- `cache`


### Standard Request Options [as specified in MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#options)

#### Present in both `undici-types` and `lib.dom.d.ts`:
- `method`
- `headers`
- `body`
- `mode`
- `credentials`
- `redirect`
- `referrer`
- `referrerPolicy`
- `integrity`
- `keepalive`
- `signal`

#### Present only in `lib.dom.d.ts`:
- `cache`

#### Missing in both `undici-types` and `lib.dom.d.ts`:
- `priority` (⚠️ this field is still considered experimental but it is already supported in Chromium based browsers)

With all of this being said, I think we can leave `priority` out and let it be detected as an unknown property since it will still be passed down to fetch.

Please let me know what you think of this approach @sindresorhus and @sholladay.

sindresorhus: Yeah, that sounds like a good approach.

sindresorhus:
Thanks for the quick fix 👍

PR #543, opened 2023-11-12
Make types compatible with both DOM and Node.js

alecmev:
Otherwise TypeScript fails with `skipLibCheck: false`, `exactOptionalPropertyTypes: true` and no `DOM`:

```
Error: node_modules/ky/distribution/types/options.d.ts(226,18): error TS2430: Interface 'NormalizedOptions' incorrectly extends interface 'RequestInit'.
  Types of property 'credentials' are incompatible.
    Type 'RequestCredentials | undefined' is not assignable to type 'RequestCredentials'.
      Type 'undefined' is not assignable to type 'RequestCredentials'.
Error: node_modules/ky/distribution/types/options.d.ts(16,29): error TS2552: Cannot find name 'HeadersInit'. Did you mean 'KyHeadersInit'?
Error: node_modules/ky/distribution/types/options.d.ts(160,21): error TS2552: Cannot find name 'RequestInfo'. Did you mean 'RequestInit'?
Error: node_modules/ky/distribution/types/options.d.ts(226,18): error TS2430: Interface 'NormalizedOptions' incorrectly extends interface 'RequestInit'.
  Types of property 'method' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

Would add a test, but it doesn't seem like there are any type tests. How do you want to go about this?

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -175,7 +175,7 @@ export class Ky {
 			this._options.duplex = 'half';
 		}
 
-		this.request = new globalThis.Request(this._input as RequestInfo, this._options as RequestInit);
+		this.request = new globalThis.Request(this._input, this._options);
 
 		if (this._options.searchParams) {
 			// eslint-disable-next-line unicorn/prevent-abbreviations
--- a/source/types/options.ts
+++ b/source/types/options.ts
@@ -22,7 +22,8 @@ export type DownloadProgress = {
 	totalBytes: number;
 };
 
-export type KyHeadersInit = HeadersInit | Record<string, string | undefined>;
+// Not HeadersInit directly because @types/node doesn't export it
+export type KyHeadersInit = NonNullable<RequestInit['headers']> | Record<string, string | undefined>;
 
 /**
 Custom Ky options
@@ -177,7 +178,7 @@ export type KyOptions = {
 	const json = await ky('https://example.com', {fetch}).json();
 	```
 	*/
-	fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
+	fetch?: (input: Input, init?: RequestInit) => Promise<Response>;
 };
 
 /**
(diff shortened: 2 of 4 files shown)


source/core/Ky.ts
```diff
@@ -175,7 +175,7 @@ export class Ky {
 			this._options.duplex = 'half';
 		}
 
-		this.request = new globalThis.Request(this._input as RequestInfo, this._options as RequestInit);
```
alecmev, line 178: Doesn't appear to be needed?

source/types/options.ts
```diff
@@ -177,7 +178,7 @@ export type KyOptions = {
 	const json = await ky('https://example.com', {fetch}).json();
 	```
 	*/
-	fetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
```
alecmev, line 180: This is breaking, but actually both [`@types/node`](https://github.com/nodejs/undici/blob/91ca13ccd065d50c47279e9fdac0c152464baee9/types/fetch.d.ts#L12) and [`lib.dom.d.ts`](https://github.com/microsoft/TypeScript/blob/ca7a3af5e607a6c8589f5f600d6c4960a8bc07d4/src/lib/dom.generated.d.ts#L25921) define it this way.

source/types/options.ts
```diff
@@ -251,8 +252,8 @@ Normalized options passed to the `fetch` call and the `beforeRequest` hooks.
 */
 export interface NormalizedOptions extends RequestInit { // eslint-disable-line @typescript-eslint/consistent-type-definitions -- This must stay an interface so that it can be extended outside of Ky for use in `ky.create`.
 	// Extended from `RequestInit`, but ensured to be set (not optional).
-	method: RequestInit['method'];
-	credentials: RequestInit['credentials'];
+	method: NonNullable<RequestInit['method']>;
+	credentials: NonNullable<RequestInit['credentials']>;
```
alecmev, line 256: `exactOptionalPropertyTypes: true` fails on this, as optional properties get resolved as `Foo | undefined`, and that isn't compatible with `RequestInit`.

sindresorhus:
I'm unable to publish this until https://github.com/sindresorhus/ky/issues/544 is figured out.

alecmev:
That's alright, thanks for taking a look!

PR #547, opened 2023-12-06
Fix #528 (retry count 1 less than retry limit)

### LINKED ISSUE #528: retry is wrongly counting
I wrote a library that can use ky or got as an HTTP client.

I wanted to verify retries are consistent. To be sure about hat, I wrote a little server that counts retries:

```js
let retryCount = -1
let last 

const server = require('http').createServer((req, res) => {
  if (Date.now() - last > 5000) retryCount = -1
  ++retryCount
  res.statusCode = 500
  console.log({ retryCount })
  last = Date.now()
  res.end(String(retryCount))
})

server.listen(4000, () => {
  console.log('> Listening at http://localhost:4000')
})
```

Then, I setup my code with `{ retry: 3 }`

Using `got`:

```js
{ retryCount: 0 } // first request
{ retryCount: 1 }
{ retryCount: 2 }
{ retryCount: 3 }
```

Using `ky`:

```js
{ retryCount: 0 } // first request
{ retryCount: 1 }
{ retryCount: 2 }
```

As you can see, ky is counting the first request as retry while retry should be counter _after_ the first request is performed.

I'm opening this issue since it's unsure if that is intentional; My fix was to count +1 in the case of ky:
https://github.com/microlinkhq/mql/pull/130/

MunyuShizumi:
Fixes #528 (actual retry count being 1 less than the specified retry limit).
This could be seen as a breaking change for projects using the +1 workaround.

The [main issue](https://github.com/sindresorhus/ky/blob/3cf50649ac2a9d5230ebcff94fa58b63a77781dc/source/core/Ky.ts#L209-L211) was `_retryCount` first being incremented, then being checked if it's less than `_options.retry.limit`, resulting in an off-by-one which is fixed with a less-or-equal comparison instead.

Re-enabled retry count test, and fixed other misconfigured tests which either explicitly failed or had silent bugs related to this. All tests pass with Node v18.18.0 (v18.18.2 breaks a number of tests, and Ava can't use custom loaders with v18.19 or v20+ making testing impossible).
EDIT: CI seems to fail since it doesn't extend `tsconfig.json` properly - I instead pointed it explicitly at `./node_modules/@sindresorhus/tsconfig/tsconfig.json` as a quickfix, but messing with the base config seems beyond the scope of this PR.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -208,7 +208,7 @@ export class Ky {
 	protected _calculateRetryDelay(error: unknown) {
 		this._retryCount++;
 
-		if (this._retryCount < this._options.retry.limit && !(error instanceof TimeoutError)) {
+		if (this._retryCount <= this._options.retry.limit && !(error instanceof TimeoutError)) {
 			if (error instanceof HTTPError) {
 				if (!this._options.retry.statusCodes.includes(error.response.status)) {
 					return 0;
--- a/test/browser.ts
+++ b/test/browser.ts
@@ -479,7 +479,7 @@ browserTest('retry with body', [chromium, webkit], async (t: ExecutionContext, p
 		page.evaluate(async (url: string) => window.ky(`${url}/test`, {
 			body: 'foo',
 			method: 'PUT',
-			retry: 2,
+			retry: 1,
 		}), server.url),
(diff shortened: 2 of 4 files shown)


sholladay (approved):
LGTM, aside from the CI failure.

PR #557, opened 2024-01-16
fix: Add a condition to verify that the `accept` of the request header starts with `'text'`

### LINKED ISSUE #556: [BUG] When Response is a string(not valid JSON), `json` method occurs error
- Example

```ts
const instance = ky.create({});

// URL returns string 'OK'
instance.get(URL).text();
```

- Error Message

```
VM21639:1 Uncaught (in promise) SyntaxError: Unexpected token 'O', "OK" is not valid JSON
```

Same as  when use`JSON.parse('OK')`

- How to solve

https://github.com/sindresorhus/ky/blob/main/source/core/Ky.ts#L81

A try-catch block may be needed when the type is `json`.
Now, it seems that the code has been written to ensure that the API response value is JSON.
If an error occurs while call `json` method, it would be good to return empty object.

- Temporary solution

```ts
instance.get(URL, { parseJson: (text) => text }).text();
```

alstn2468:
Fixed #556 

## [Example]

```ts
const instance = ky.create({});

// URL returns string 'OK'
instance.get(URL).text();
```

## [Error Message]

```
VM21639:1 Uncaught (in promise) SyntaxError: Unexpected token 'O', "OK" is not valid JSON
```

Same as  when use`JSON.parse('OK')`

## [Temporary solution]

```ts
instance.get(URL, { parseJson: (text) => text }).text();
```

## [Description]

For `json` type, if there is no JSON type of response, an exception occurs when invoking the `JSON.parse` function.
The `text` method of `Response` has been modified to return a string as it is in the response.
I think we can discuss which values we should return (whether we return empty strings like 204)
I think it's an error to have a parsing error when the user didn't directly call the `json` method.

2024.01.17
It was modified from using the try catch block to confirm the `accpet` of the request header.
If the type is `json` and you can't parse it with JSON and you don't want to process it, I'd like your opinion.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -101,7 +101,7 @@ export class Ky {
 						return options.parseJson(await response.text());
 					}
 
-					if (response.headers.get('accept')?.startsWith('text')) {
+					if (ky.request.headers.get('accept')?.startsWith('text')) {
 						return response.text();
 					}
 				}
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -103,9 +103,9 @@ export class Ky {
 
 					// If error causes when call JSON.parse
 					try {
-						return response[type]();
+						return (await response.json());
 					} catch {
-						return await response.text()
+						return response.text();
 					}
 				}
 
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -101,10 +101,7 @@ export class Ky {
 						return options.parseJson(await response.text());
 					}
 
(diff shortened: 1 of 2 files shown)


alstn2468:
@sindresorhus I think I need to add a test code for `Ky` instance, so it would be nice if you could give me your opinion on which one to add.

alstn2468:
Oh I just remembered something, and I think this issuse might be a problem with hooks applied to instance.
I'll check a little more and close the PR.

PR #559, opened 2024-02-22
fix: support cloudflare workers

### LINKED ISSUE #366: The `credentials` field on `RequestInitializerDict` is not implemented; Cloudflare Workers issue.
I am experiencing the same issue with credentials that prevents `ky` from being used with Cloudflare Workers that `redaxios` is also [experiencing here](https://github.com/developit/redaxios/issues/64).

riderx:
This PR comes to fix the Cloudflare workers issue mentioned here with the solution:
https://github.com/sindresorhus/ky/issues/366#issuecomment-1670435774

Fixes #366

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -118,9 +118,10 @@ export class Ky {
 	// eslint-disable-next-line complexity
 	constructor(input: Input, options: Options = {}) {
 		this._input = input;
+		const isCredentialsSupported = 'credentials' in Request.prototype;
 		this._options = {
 			// TODO: credentials can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
-			credentials: (this._input as Request).credentials || 'same-origin',
+			credentials: isCredentialsSupported ? ((this._input as Request).credentials || 'same-origin') : undefined,
 			...options,
 			headers: mergeHeaders((this._input as Request).headers, options.headers),
 			hooks: deepMerge<Required<Hooks>>(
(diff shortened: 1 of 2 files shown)


sindresorhus:
CI is failing

riderx:
@sindresorhus fixed :)

source/core/Ky.ts
```diff
@@ -118,9 +118,10 @@ export class Ky {
 	// eslint-disable-next-line complexity
 	constructor(input: Input, options: Options = {}) {
 		this._input = input;
+		const isCredentialsSupported = 'credentials' in Request.prototype;
 		this._options = {
 			// TODO: credentials can be removed when the spec change is implemented in all browsers. Context: https://www.chromestatus.com/feature/4539473312350208
-			credentials: (this._input as Request).credentials || 'same-origin',
+			credentials: isCredentialsSupported ? ((this._input as Request).credentials || 'same-origin') : 'same-origin',
```
fregante, line 124: I don’t think this is gonna work, `_options.credentials` should not be set at all, according to the error.

Also the comment is very outdated and we can probably use:

```suggestion
			credentials: isCredentialsSupported ? (this._input as Request).credentials : undefined,
```

… as long as CloudFlare accepts `{credentials: undefined}` and doesn't actually require `{}`

fregante (changes requested):

riderx:
Thanks @fregante i did update it.
I had to change the typing to make this solution work, and added a comment.
I did try and `{credentials: undefined}` is the required way for Cloudflare.
Since isCredentialsSupported is not reused, I wonder if that shouldn't be better, directly in the ternary?

fregante (approved):

PR #561, opened 2024-02-26
Allow input to start with slash by removing head slashes

jdavidferreira:
- Allow `input` param to start with slash (`/`) by removing head slashes.
- Add missing `undici-types` dev dependency.


package.json
```diff
@@ -72,6 +72,7 @@
 		"raw-body": "^2.5.2",
 		"tsx": "^4.7.0",
 		"typescript": "^5.2.2",
+		"undici-types": "^6.6.2",
```
sindresorhus, line 75: Where is this used?

sindresorhus:
From the docs:

> Leading slashes in input are disallowed when using this option to enforce consistency and avoid confusion about how the input URL is handled, given that input will not follow the normal URL resolution rules when prefixUrl is being used, which changes the meaning of a leading slash.

package.json
jdavidferreira: It is used here: https://github.com/sindresorhus/ky/blob/cb89702ca488717a8b2021522f5ffe05a3c01a99/source/types/request.ts#L1.

I got a warning about the library not being found.

jdavidferreira:
> From the docs:
> 
> > Leading slashes in input are disallowed when using this option to enforce consistency and avoid confusion about how the input URL is handled, given that input will not follow the normal URL resolution rules when prefixUrl is being used, which changes the meaning of a leading slash.

I see. Could it be something "discouraged", rather than something "disallowed"? Could this change break things in the library?

package.json
jdavidferreira: I removed the dependency since it is no needed from [v1.2.2](https://github.com/sindresorhus/ky/compare/v1.2.1...v1.2.2).

sholladay:
There's been a lot of discussion on this before in the issues and previous PRs. It's easily the most controversial part of Ky. Basically, it's an opinionated choice meant to make it abundantly clear that the `input` URL will be appended to the `prefixUrl` and, more specifically, not resolved against the origin.

FWIW, my stance on this has softened because of how many people complain about it. I'd like to find a way to satisfy everyone.

fregante:
While I agree with the relative-vs-root argument, most REST API docs show the endpoints as `/something` so ky should probably follow suit and maybe even _require_ that.


- https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28
- https://www.reddit.com/dev/api/
- https://open-api.netlify.com/#tag/user/operation/getCurrentUser
- https://www.twilio.com/docs/sendgrid/api-reference/mail-send/mail-send
- https://developer.x.com/en/docs/twitter-api/tweets/timelines/api-reference/get-users-id-tweets
- https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_limits.htm

Those I found that don't follow the format, it's just because they show the full URL (Google APIs, Microsoft APIs)

sholladay:
See PR #606 for my proposed solution.

faradaytrs:
is there any progress on this?

sholladay:
No but it's very much on our radar. I still have a bit of work to do on #606 before it's finalized. When it is ready, I want to discuss which approach is preferred and hopefully merge one or the other. I like having both PRs to look at so the discussion is less abstract.

bertho-zero:
I prefer this PR to #606, just remove that `if` which doesn't really provide the protection it thinks it does.

sindresorhus:
Closing in favor of #606

PR #565, opened 2024-03-21
Fix `exactOptionalPropertyTypes`, fix TypeScript 5.4

alecmev:
https://github.com/sindresorhus/ky/pull/559 undid the fix in https://github.com/sindresorhus/ky/pull/543. Added `exactOptionalPropertyTypes` to `compilerOptions` to prevent this from happening again, seems pretty low-impact. It should be okay to pass `undefined` to `credentials` and other options, but this is an upstream problem (see `RequestInit` in `undici-types/fetch.d.ts`).

The build was failing on TS 5.4 because there's a `priority` in `RequestInit` now.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -118,9 +118,13 @@ export class Ky {
 	// eslint-disable-next-line complexity
 	constructor(input: Input, options: Options = {}) {
 		this._input = input;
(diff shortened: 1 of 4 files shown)


alecmev:
Thanks!

PR #578, opened 2024-04-20
Fix request types for use with Node

jliuhtonen:
Screenshot 2024-04-20 at 21 59 11

When using Ky with Node, TypeScript will run into compilation error regarding types required by `UndiciRequestInit`. This is because `@types/node` does not yet provide these types, and they are not imported from `undici-types` either. The latter would not even be possible to keep Ky usable with browsers.

This happened to work in the project, because `@sindresorhus/tsconfig` includes `DOM` in the libs section. In most Node applications, the DOM lib should not be there and clashes with Node types. One way to circumvent this is to add `skipLibCheck: true` to `tsconfig.json`, but this is not a good solution.

This commit adds the needed types until `@types/node` imports the missing types and adds them to globals. After that, all "Undici" types from the request types should be removed.

### DIFF
--- a/source/types/request.ts
+++ b/source/types/request.ts
@@ -1,15 +1,49 @@
+type UndiciHeadersInit =
+	| string[][]
+	| Record<string, string | readonly string[]>
+	| Headers;
+
+type UndiciBodyInit =
+	| ArrayBuffer
+	| AsyncIterable<Uint8Array>
+	| Blob
+	| FormData
+	| Iterable<Uint8Array>
+	| NodeJS.ArrayBufferView
+	| URLSearchParams
+	// eslint-disable-next-line @typescript-eslint/ban-types
+	| null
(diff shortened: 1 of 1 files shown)


jliuhtonen:
Here is [an example](https://github.com/jliuhtonen/blu-hawaii/pull/20) of a project not compiling because of the issue fixed in this PR.

sindresorhus:
> This commit adds the needed types until @types/node imports the missing types and adds them to globals. After that, all "Undici" types from the request types should be removed.

Can you open an issue on the TS types repo and add a code comment with the link? I like to know that it will eventually be resolved upstream.

jliuhtonen:
@sindresorhus done, I opened a discussion here: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/69408

Code comment also added.

PR #579, opened 2024-04-21
feat: add support for `stringifyJson`

Reverier-Xu:
When sending an HTTP request, this option allows the user to specify how the request data should be serialized as a string.

For example, it will be useful when users want to change the default serialization behavior using `JSON.stringify(_, replacer)`.

```typescript
import ky from 'ky';
import { DateTime } from 'luxon';

const json = await ky('https://example.com', {
  stringifyJson: data => JSON.stringify(data, (key, value) => {
    if (key.endsWith('_at')) {
      return DateTime.fromISO(value).toSeconds();
    }
    return value;
  })
}).json();
```

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -203,7 +203,7 @@ export class Ky {
 		}
 
 		if (this._options.json !== undefined) {
-			this._options.body = JSON.stringify(this._options.json);
+			this._options.body = this._options.stringifyJson ? this._options.stringifyJson(this._options.json) : JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', this._options.headers.get('content-type') ?? 'application/json');
 			this.request = new globalThis.Request(this.request, {body: this._options.body});
 		}
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -203,7 +203,7 @@ export class Ky {
 		}
 
 		if (this._options.json !== undefined) {
-			this._options.body = this._options.stringifyJson ? this._options.stringifyJson(this._options.json) : JSON.stringify(this._options.json);
+			this._options.body = this._options.stringifyJson?.(this._options.json) ?? JSON.stringify(this._options.json);
 			this.request.headers.set('content-type', this._options.headers.get('content-type') ?? 'application/json');
 			this.request = new globalThis.Request(this.request, {body: this._options.body});
 		}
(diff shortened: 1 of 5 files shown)


source/core/Ky.ts
```diff
@@ -203,7 +203,7 @@ export class Ky {
 		}
 
 		if (this._options.json !== undefined) {
-			this._options.body = JSON.stringify(this._options.json);
+			this._options.body = this._options.stringifyJson ? this._options.stringifyJson(this._options.json) : JSON.stringify(this._options.json);
```
sindresorhus, line 206:
```suggestion
			this._options.body = this._options.stringifyJson?.(this._options.json) ?? JSON.stringify(this._options.json);
```

test/main.ts
```diff
@@ -733,3 +733,24 @@ test('parseJson option with promise.json() shortcut', async t => {
 
 	await server.close();
 });
+
+test('stringifyJson option with request.json()', async t => {
+	const server = await createHttpTestServer({bodyParser: false});
+	server.post('/', async (request, response) => {
+		const body = await parseRawBody(request);
+		t.is(body, '{"data":{"hello":"world"},"extra":"extraValue"}');
```
sindresorhus, line 741: Would be more readable as an object passed to `JSON.stringify`.

sindresorhus: Just move the `JSON.stringify({
			data,
			extra: 'extraValue',
		})` part to the top and use it in this check too.

sindresorhus (changes requested):

Reverier-Xu:
Done. 😉

test/main.ts
```diff
@@ -733,3 +733,26 @@ test('parseJson option with promise.json() shortcut', async t => {
...
+		const body = await parseRawBody(request);
+		t.is(body, JSON.stringify(result));
+		response.end();
+	});
+
+	await ky.post(server.url, {
+		stringifyJson: data => JSON.stringify({
+			data,
+			extra: 'extraValue',
+		}),
```
sindresorhus, line 753: I think you misunderstood. `JSON.stringify({
			data,
			extra: 'extraValue',
		})` should be used both here and in 745.

Reverier-Xu: No, the `JSON.stringify` here is to add fields to any additional request data, not just the data in the example, so it's not appropriate to unify the two.

sindresorhus, line 753:
```suggestion
	const json = {hello: 'world'};
	const extra = 'extraValue';

	server.post('/', async (request, response) => {
		const body = await parseRawBody(request);
		t.is(body, JSON.stringify({data: json, extra}));
		response.end();
	});

	await ky.post(server.url, {
		stringifyJson: data => JSON.stringify({data, extra}),
```

PR #582, opened 2024-04-25
Improve detection of streaming request support (#581)

### LINKED ISSUE #581: Error: unsupported BodyInit type
#530 

## runtime
QQ浏览器(ios) QQ(ios)

## screenshot
[test url](https://rvq9nh-5173.csb.app/)
微信图片_20240425154625

[codesandbox](https://codesandbox.io/p/github/edram/issue-ky-581/main)
[repo](https://github.com/edram/issue-ky-581/blob/main/src/main.ts#L14)

edram:
Closes #581

### DIFF
--- a/source/core/constants.ts
+++ b/source/core/constants.ts
@@ -20,7 +20,7 @@ export const supportsRequestStreams = (() => {
 				},
 			}).headers.has('Content-Type');
 		} catch (error) {
-			// IOS QQBrowser throw "unsupported BodyInit type" Error (see issue #581)
+			// QQBrowser on iOS throws "unsupported BodyInit type" error (see issue #581)
 			if (error instanceof Error && error.message === 'unsupported BodyInit type') {
 				return false;
 			}
--- a/source/core/constants.ts
+++ b/source/core/constants.ts
@@ -9,15 +9,19 @@ export const supportsRequestStreams = (() => {
 	const supportsRequest = typeof globalThis.Request === 'function';
 
 	if (supportsReadableStream && supportsRequest) {
-		hasContentType = new globalThis.Request('https://empty.invalid', {
-			body: new globalThis.ReadableStream(),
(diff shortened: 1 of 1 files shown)


source/core/constants.ts
```diff
@@ -9,15 +9,19 @@ export const supportsRequestStreams = (() => {
...
+				body: new globalThis.ReadableStream(),
+				method: 'POST',
+				// @ts-expect-error - Types are outdated.
+				get duplex() {
+					duplexAccessed = true;
+					return 'half';
+				},
+			}).headers.has('Content-Type');
+		} catch {
+			// IOS QQBrowser throw "unsupported BodyInit type" Error (see issue #581)
```
sholladay, line 23: Sorry, but we cannot silently ignore all errors from the `Request` constructor. There are many genuine, useful errors that it can potentially throw. Maybe you could explain what you are trying to accomplish (I couldn't understand from the linked issue). If you think we should be ignoring a particular error in certain circumstances, then we will need to properly detect those conditions and the specific error.

edram: Based on my understanding, this code is supposed to determine whether the runtime supports streaming requests. However, the [code](https://github.com/edram/issue-ky-581/blob/main/src/main.ts#L14) above throws `Error: unsupported BodyInit type` in the QQ app or QQ browser on iOS. This indicates that these two runtimes do not support streaming requests. The check for support should not throw an error. Given the large number of QQ users, the latest version of ky in my project can use is version 0.31.1.

image

sholladay: Okay, thanks. I understand now. I have reopened the PR. Can you please update your code so it re-throws the error if `error.message !== 'unsupported BodyInit type'`?

edram: Thank you for your response. I have updated my code. 🫰

PR #591, opened 2024-06-02
Fix `onDownloadProgress` test commented out in test/browser.ts

komura-c:
Hello, I noticed a test that was commented out in browser.ts and fixed it.
When I uncommented it, I got a `ReferenceError: __name is not undefined` on playweight at the following location
```
const decodeUtf8 = (array: Uint8Array) => String.fromCodePoint(...array);
```
On playweight, it is executed as follows
```
const decodeUtf8 = __name((array) => String.fromCodePoint(...array), "decodeUtf8")
```
I found out that this was due to an optimization option `keepNames` in vite.
refs: vite/pull/2376#issuecomment-805285480, esbuild/issues/2605#issue-1404600925

In this repository, it seems to be happening only within playweight. I don't think there is any need to make it a function except for readability, so I fixed it so that it doesn't create a function.
Just to be sure, I tried it `onDownloadProgress` in a vite+react environment and the same thing did not happen, so it seems to be a good fix for testing only.

### DIFF
--- a/test/browser.ts
+++ b/test/browser.ts
@@ -3,6 +3,7 @@ import busboy from 'busboy';
 import express from 'express';
 import {chromium, webkit, type Page} from 'playwright';
 import type ky from '../source/index.js';
+import type {DownloadProgress} from '../source/index.js';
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks :)

Issue #592, opened 2024-06-19
Set TimeoutError#cause to be Request or URL


arty-name:
[TimeoutError now includes the Request as `.request` property](https://github.com/sindresorhus/ky/issues/267). 

JS errors also have a standard property `.cause` which is widely supported by various tooling. Would it make sense for TimeoutError to set this property in addition to (or even instead of) the `.request`? The value could be the whole Request or just its URL. 

If not added into the core, this could be [a use case for `beforeTimeout` hook](https://github.com/sindresorhus/ky/issues/508).

arty-name:
Actually the whole Request seems to choke my tooling at least (Vite/Astro), so the URL could be a better option.

arty-name:
Thank you everyone for the quick solution! :heart:

PR #593, opened 2024-06-19
Add `cause` to `TimeoutError`

arty-name:
Fixes https://github.com/sindresorhus/ky/issues/592

This is the simplest approach we could take

### DIFF
--- a/source/errors/TimeoutError.ts
+++ b/source/errors/TimeoutError.ts
@@ -2,7 +2,7 @@ export class TimeoutError extends Error {
 	public request: Request;
 
 	constructor(request: Request) {
-		super('Request timed out', { cause: request });
+		super('Request timed out', { cause: request.url });
 		this.name = 'TimeoutError';
 		this.request = request;
 	}
--- a/source/errors/TimeoutError.ts
+++ b/source/errors/TimeoutError.ts
@@ -2,7 +2,7 @@ export class TimeoutError extends Error {
 	public request: Request;
 
 	constructor(request: Request) {
-		super('Request timed out');
+		super('Request timed out', { cause: request });
 		this.name = 'TimeoutError';
 		this.request = request;
 	}


arty-name:
The CI fails on TypeScript not knowing of this property. Locally I see this defined in `lib.es2022.error.d.ts`, so apparently this was added in ES 2022, which is [indirectly confirmed by MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#browser_compatibility). Apparently Ky uses a lower version of JS, which is reasonable.

I believe that this second parameter to the Error constructor will be simply ignored by older runtimes, so there should be no danger in adding it. 

What would be your preferred way of handling this error in Ky project?

sindresorhus:
I think it would be more correct to attach the URL as a property on the error. The URL is not really the "cause" of the error.

arty-name:
Thank you for the quick response!

I see your point, that’s fair enough.  I understand that you wouldn’t want to have this in the core.

Personally, I’ll still use `cause` for practical reasons. Do you think that this use case justifies adding the `beforeTimeout` hook?

sindresorhus:
> Personally, I’ll still use cause for practical reason: only cause is supported by tooling.

Not sure what you mean by this, but adding properties to errors has always been the way to add metadata.

arty-name:
It was, but only the `cause` gets special treatment by e.g. browser developer tools and Vite/Astro developer overlay. And I am not sure how normative this is, but [MDN suggests to use it for metadata](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause).
image
image
image

sholladay:
The URL alone doesn't seem like a good use of `cause`.

I would be okay with either or both of the following:
1. Move `error.request` to `error.cause` to conform to the expectations of the broader ecosystem and make the metadata more discoverable.
2. Add the request method and URL directly to the timeout error message for simple debugging cases

sindresorhus:
> Move error.request to error.cause to conform to the expectations of the broader ecosystem and make the metadata more discoverable

While it is acceptable to use `cause` for non-errors, I don't think the community conventions for `cause` are established enough yet for me to want to do this.

> Add the request method and URL directly to the timeout error message for simple debugging cases

Let's do this.

sholladay:
Closing in favor of PR #595 

@arty-name what do you think of that solution?

arty-name:
@sholladay This solution will also resolve my issue, thank you!

PR #595, opened 2024-06-23
Add request method and URL to error messages

### LINKED ISSUE #592: Set TimeoutError#cause to be Request or URL

sholladay:
Closes #592

This PR aims to make the error messages from Ky's `TimeoutError` and `HTTPError` more helpful by including the request method and URL.

Before:

```
TimeoutError: Request timed out
```
```
HTTPError: Request failed with status code 400 Bad Request
```

After:

```
TimeoutError: Request timed out: POST https://localhost:3000/
```
```
HTTPError: Request failed with status code 400 Bad Request: POST https://localhost:3000/
```

### DIFF
--- a/source/errors/HTTPError.ts
+++ b/source/errors/HTTPError.ts
@@ -12,7 +12,7 @@ export class HTTPError extends Error {
 		const status = `${code} ${title}`.trim();
 		const reason = status ? `status code ${status}` : 'an unknown error';
 
-		super(`Request failed with ${reason}`);
+		super(`Request failed with ${reason}: ${request.method} ${request.url}`);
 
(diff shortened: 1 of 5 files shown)


arty-name:
I guess, too long URLs are not a real concern here

PR #599, opened 2024-06-25
Allow explicitly unsetting hooks on `.extend()`

### LINKED ISSUE #408: Allow extend() to add/remove hooks
Two Code Snippets

```ts
import ky from 'ky';

// Root Ky
const rootKy = ky.create({
  prefixUrl: '/api',
  timeout: false,
  
  hooks: {
    beforeRequest: [pruneRequestPayload],
    beforeRetry: [logError],
  },
});

// Ky which serializes request and response
export const kyWithSerializer = rootKy.extend({
  hooks: {
    beforeRequest: [requestToSnakeCase],
    afterResponse: [responseToCamelCase],
  },
});
```

Expectation: `kyWithSerializer` to also execute `pruneRequestPayload` in beforeRequest hook and should have `beforeRetry` present too
Actual: `beforeRequest` hook array is completely replaced with new array

Kenneth-Sills:
Ref: #408

The implementation here is pretty simple: just give hook merging some custom logic, so it either merges OR resets to an empty array. Matches what was discussed in the related ticket.

Documentation has been updated to describe this behavior as well.

The dedicated test is just a modified copy of the existing `ky.extend()` one. Admittedly, I did modify a couple of other tests to cover the edge cases instead of building brand-new tests. No reason other than noticing they were already close to what I needed.. and I'm being lazy. 😅


sholladay:
I left my thoughts on this approach here: https://github.com/sindresorhus/ky/issues/408#issuecomment-2188241786

TL;DR: I don't like it. It continues down a path that is too convoluted and magical for my taste.

That aside, the implementation in this PR looks good to me.

Kenneth-Sills:
@sholladay This should be ready for re-review, following [your latest comment](https://github.com/sindresorhus/ky/issues/408#issuecomment-2239919656) in the related ticket. Thanks!

Issue #291, opened 2020-10-21
prefixUrl is unneededly applied to absolute urls


ux-engineer:
I'm having a problem with `prefixUrl` option:
```
const ky = Ky.extend({
  prefixUrl: API_BASE_URL,
});
```
Our backend API is returning absolute download paths which already include the base url. In those cases we end up with request urls having prefixed twice with the base url.

Could we have `prefixUrl` option's behavior altered so that it checks if the url already starts with this value, and if so it would not prefix it there?

I tried to create a beforeRequest hook to modify request.url, but it seems the hook cannot modify that property. That approach would have averted having to check and modify urls in possibly multiple parts of the codebase.

sholladay:
We could, but my concern is it will then be confusing if it still applies the prefix to input URLs like `../meow.jpg`, which is another case you probably don't want it. More likely, you want some type of URL resolution, and Ky currently does not provide any mechanism for that when using `prefixUrl`. It's just a simple prefix.

In the past, I've proposed adding a `baseUrl` option that would be akin to the HTML `<base>` tag, which would resolve the input URL against a base URL. Sindre thought it would be confusing to have both options, and it turns out a lot of people wanted the prefix behavior so that `/meow.jpg` and `meow.jpg` with a prefix would load the same URL, whereas those are very different URLs when resolving against a base. So that's how we ended up with the current system.

You're probably thinking this makes sense but is still a bit kludgy and I agree. One thought I've been kicking around is to let `beforeRequest` return a URL instance. Internally, t's a little tricky to update the request URL without messing up the request object, unfortunately. But it should be doable.

Open to other suggestions!

haggen:
@ux-engineer @sholladay What if we keep the current behavior of `prefixUrl`, but also add support for a new option `resolveUrl` value, to which we delegate the URL resolution process? This would helpful to me as well. I was thinking something of the sort:

```js
const api = ky.extend({
  prefixUrl: "https://api",

  resolveUrl(url, prefix) {
    return url.startsWith("http") ? url : prefix + url;
  }
});

api.get("foo"); //=> GET https://api/foo
api.get("http://bar"); //=> GET https://bar
```

sholladay:
A resolver function is a decent idea, although to me this is just a more verbose way to invoke the browser's URL resolution logic, compared to a `baseUrl` option.

@sindresorhus what if we add `baseUrl` but make it mutually exclusive with `prefixUrl`?

sindresorhus:
[...]

PR #606, opened 2024-07-04
Add baseUrl option, rename prefixUrl, and allow leading slashes in input

sholladay:
This PR aims to improve the flexibility of input URLs and options for resolving them prior to the request.

- `prefixUrl` is renamed to `startPath`
- - As before, when this option is used, it is prepended to `input` before any other processing takes place
- - `startPath` can still be a full URL if needed, however the name is meant to signal its primary use case is a simple base path without a domain, e.g. `/api`
- A new `baseUrl` option is added
- - When this option is used, `input` (including any `startPath`) is resolved against the `baseUrl` according to standard URL resolution rules
- - `baseUrl` is similar to the HTML `<base>` tag, except that it only applies to Ky
- `input` can now be any valid URL, with or without a leading slash
- - Slashes are automatically normalized when using `startPath`
- - Slashes are kept as-is and respected in accordance with standard URL resolution rules when using `baseUrl`

```js
ky('books', { startPath : '/api' });  // GET /api/books
ky('books', { baseUrl : '/api' });  // GET /books
ky('books', { baseUrl : '/api/' });  // GET /api/books
ky('./books', { baseUrl : '/api/' });  // GET /api/books
ky('/books', { baseUrl : '/api/' });  // GET /books
ky('http://bar.com', { startPath : 'http://foo.com' });  // GET http://foo.com/http://bar.com
ky('http://bar.com', { baseUrl : 'http://foo.com' });  // GET http://bar.com
ky('books', { baseUrl : 'http://foo.com', startPath : '/api' });  // GET http://foo.com/api/books
```

Possible names for the string that is prepended to `input`
 - `startPath` (as is currently implemented here)
 - `pathPrefix`
 - `prefixPath`
 - `rootPath`

TODO:
[ ] Finalize name for `startPath` option
[ ] Documentation
[ ] TypeScript types


faradaytrs:
i see that it's going to be a breaking change, maybe it's worth about discusssing about the next major? i think it's good to remove default timeout and retry settings, because many people come from fetch and axios and it's not obvious ky have these defaults.

sholladay:
I would be okay with removing the default timeout, but not the default retries.

alexgleason:
baseUrl looks like what I want. I only care about the URL's origin, and also need the ability to pass full URLs. If it works like this, it's perfect:

```ts
ky.get('/api/v1/instance', { baseUrl : 'https://ditto.pub' }); // GET https://ditto.pub/api/v1/instance
ky.get('https://ditto.pub/api/v1/timelines/home?max_id=1234', { baseUrl : 'https://ditto.pub' }); // GET https://ditto.pub/api/v1/timelines/home?max_id=1234
```

In this case, I don't care about a path in the baseUrl. I assume it works like this:

```ts
ky.get('/api/v1/instance', { baseUrl : 'https://ditto.pub/@alex' }); // GET https://ditto.pub/api/v1/instance
```

But I will never pass a baseUrl with a path anyway.

sholladay:
@alexgleason correct, each of those examples behaves as you described.

I would expect most people to structure their requests like this:

```js
const api = ky.extend({ baseUrl : 'https://ditto.pub/api/v1/' });
api.get('instance'); // GET https://ditto.pub/api/v1/instance
api.get('timelines/home?max_id=1234', { baseUrl : 'https://ditto.pub/api/v1' }); // GET https://ditto.pub/api/v1/timelines/home?max_id=1234
api.get('https://example.com'); // GET https://example.com
api.get('/api/v2/secret'); // GET https://ditto.pub/api/v2/secret
```

But you could certainly put the `api/v1` part in the input instead of at the end of `baseUrl`, that works, too.

alexgleason:
@sholladay I think the way to overcome code complexity and confusion while solving all use-cases is to just support a resolver function like mentioned here: https://github.com/sindresorhus/ky/issues/291#issuecomment-834772398

But I think it should actually look like this:

```ts
const api = ky.create({
  // Accept any `Input` and return any `Input`. Flexible and pure.
  resolver(input: Input): Input {
    // It's not actually complicated to do this yourself either.
    const url = input instanceof Request ? input.url : input.toString();
    return new Request(new URL(url, 'https://ditto.pub'), input);
  }
});
```

We can keep the `prefixUrl` for backwards-compatibility, but internally it just creates a resolver, so you can't have both.

```ts
const api = ky.create({
  prefixUrl: 'https://ditto.pub', // WARNING: `prefixUrl` is deprecated.
  resolver(input: Input): Input { // ERROR: `prefixUrl` and `resolver` cannot be specified at the same time.
    // ...
  }
});
```

For improved DX, Ky can include resolvers for a prefix and baseUrl.

```ts
import ky from 'ky';

const api = ky.create({
  resolver: ky.prefix('https://ditto.pub/api/v1')
});

const api = ky.create({
  resolver: ky.baseUrl('https://ditto.pub')
});
```

sholladay:
@alexgleason the way I am planning to solve that is by allowing users to return a URL instance in a `beforeRequest` hook. Do you think that is sufficient? It avoids needing to add another option.

alexgleason:
@sholladay That could work as long as we have the `input` there too. In fact that's the only reason it doesn't work already. You don't even need to let it return a URL, you just need to add the `input` as a parameter.

alexgleason:
Can it even get to the point of `beforeRequest` without already having a valid absolute URL though? I don't think it can construct the needed Request object unless you're already specifying a prefixUrl.

sholladay:
While it doesn't provide input, you can use `request.url` to see where the request is going to go. That might not suit every use case, though.

alexgleason:
@sholladay Currently we can already do that with `beforeRequest`, but when the request looks like `Request { url: "https://ditto.pub/https://ditto.pub/api/v1/timelines/home?max_id=1234" }` it's not very helpful. I need the Input so I can join it with a baseUrl. The input might be full URL, or just a path like `/api/v1/instance`.

sholladay:
What's the use case that causes you to need a base URL that is unknown at the time of creating the Ky instance?

alexgleason:
I ended up just writing a custom bare minimum fetch wrapper and accepting that I will have to do `await (await ()).json()` all over my application code. https://gitlab.com/soapbox-pub/soapbox/-/blob/main/src/api/MastodonClient.ts?ref_type=heads

sholladay:
@bertho-zero you didn't explain what you prefer about this PR and why.

Your regex is shorter but the current code is more readable and should have slightly better performance. Both are fine, though, and I don't have strong feelings about it.

source/types/options.ts
```diff
@@ -97,23 +97,23 @@ export type KyOptions = {
 	Useful when used with [`ky.extend()`](#kyextenddefaultoptions) to create niche-specific Ky-instances.
 
 	Notes:
-	 - After `prefixUrl` and `input` are joined, the result is resolved against the [base URL](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI) of the page (if any).
-	 - Leading slashes in `input` are disallowed when using this option to enforce consistency and avoid confusion about how the `input` URL is handled, given that `input` will not follow the normal URL resolution rules when `prefixUrl` is being used, which changes the meaning of a leading slash.
+	 - After `startPath` and `input` are joined, the result is resolved against the [base URL](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI) of the page (if any).
+	 - Leading slashes in `input` are disallowed when using this option to enforce consistency and avoid confusion about how the `input` URL is handled, given that `input` will not follow the normal URL resolution rules when `startPath` is being used, which changes the meaning of a leading slash.
```
LinusU, line 101: > Leading slashes in `input` are disallowed when using this option to enforce consistency and avoid confusion [...]

From what I can tell from the changed code, this is no longer the case?

I personally prefer it that way though and think that we should keep that behaviour

sholladay: Indeed. As mentioned in the TODOs section, I haven't gotten around to updating the docs yet, other than a simple find & replace. But thank you for the reminder. 🙂

bertho-zero:
I like #561 because I don't understand why we should throw an error if the input starts with a slash, I read the topic but still don't find this constraint justified.

I don't like #606 because I can have a prefix like `https://domain.com/api/v2` and I don't want to have to parse the prefix to put the host in `baseUrl` and the path in `startPath`. Or `startPath` should do this splitting by itself to avoid the example:

> ```js
> ky('http://bar.com', { startPath : 'http://foo.com' }); // GET http://foo.com/http://bar.com
> ```

source/core/Ky.ts
```diff
@@ -149,16 +151,22 @@ export class Ky {
...
+					this._options.startPath += '/';
+				}
 
-			if (!this._options.prefixUrl.endsWith('/')) {
-				this._options.prefixUrl += '/';
+				if (this._input.startsWith('/')) {
+					this._input = this._input.slice(1);
+				}
+
+				this._input = this._options.startPath + this._input;
```
bertho-zero, line 164: I think if `_input` starts with a protocol (or is a valid URL?) then we should ignore `startPath` in order to avoid an URL like `http://foo.com/http://bar.com`

sholladay: That kind of "double URL" construct is useful for reverse proxies. It should be supported.

Also, you would be surprised how hard URL parsing is. Especially for a library like Ky that supports non-browsers.

sholladay:
> I don't like #606 because I can have a prefix like `https://domain.com/api/v2` and I don't want to have to parse the prefix to put the host in `baseUrl` and the path in `startPath`.

You don't need to split it up like that, you can just put the whole thing in `startPath` if you always want that host to be used. I realize that may not be entirely obvious. I am open to suggestions for how to make it more obvious. We could just keep the name `prefixUrl `, then it would be more obvious to me, but that's just too similar to `baseUrl`, so it creates a new kind of confusion.

simon-skooster:
Any news on this PR?  I'd love to replace Axios with Ky, but this issue is a blocker.

sholladay:
This is almost ready, just need to update the docs and types. Will try to get it landed later this week.

sholladay:
@sindresorhus this is now ready for review. Currently, it's a breaking change due to the renaming of `prefix`. We could add an alias to keep backwards compatibility, but I think the major version bump is worth it to help make people aware of these changes, which I suspect will be helpful to many.

Most users will simply need to update the option name, but it's recommended to switch to `baseUrl` if possible, for improved flexibility (e.g. the ability to bypass the `baseUrl` with an absolute `input` URL).

sindresorhus:
This looks like a good improvement to me.

I would prefer keeping an alias to not make this change too disruptive. We can add a `@deprecated` on the docs for `prefixUrl` in the TS docs, and in the major version after this one, we can hard-deprecate `prefixUrl` with an error in the actual code.

readme.md
```diff
@@ -191,29 +191,55 @@ Search parameters to include in the request URL. Setting this will override all
 
 Accepts any value supported by [`URLSearchParams()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams).
 
-##### prefixUrl
```
sindresorhus, line 194: We need even clearer docs about the difference between `baseUrl` and `prefix` and when to use each. Maybe with some real-world example use-cases. Otherwise, it's going to be confusing for users and that ends up being a support burden for us.

readme.md
```diff
@@ -191,29 +191,55 @@ Search parameters to include in the request URL. Setting this will override all
...
+##### baseUrl
```
kleinfreund, line 194: I think it would be great to have a matrix like you added in the original pull request description for both the baseUrl and prefix sections that goes through the motions of showing most (if not all) relevant permutations of baseUrl (or prefix), input and the resulting request URL.

This could help de-mystify what both options do and what might be appropriate for any given use case.

I think that's especially useful because not all people will know that "prefix" and "base URL" mean (sometimes subtly) different things (of course these docs make that clear, but still).

sholladay: I was originally going to do that but it felt like a lot of information all at once, leaving the impression that it's more complex than it really is it "feel" more complex than it is. I went with a simple, explicit recommendation to use `baseUrl` over the other unless you know you need to use `prefix`.

 I think I'd like to see what questions come up before

sholladay: I was originally going to do that but it felt like a lot of information all at once and made the choices seem complex, which could actually be more confusing than helpful. I went with a simple, explicit recommendation to use `baseUrl` unless you know you need `prefix`. It seems clearer to me overall.

I think we should see what questions come up and then make tweaks based on that. If there ends up being a lot, we could add a full matrix of examples to the FAQ. What do you think?

kleinfreund: I'm perfectly fine with that.

My angle is mainly that I think there might be a lot of (potential) users coming from contexts where all they need is setting some API prefix string (e.g. `'https://my-awesome-site.com/api/v1'`) so that their API client code can be simplified to accept path segments without that prefix (e.g. `'/users'` or `'users'`) to produce a simply joined request URL (e.g. `'https://my-awesome-site.com/api/v1/users'`). For those cases, developers might never think to care about what difference a leading slash on `'/users'` vs no leading slash on `'users'` could make. For them, both are equivalent and should result in the same request URL.

Since it might never occur to them that there can even be a difference between both cases, they might also just gloss over the documentation for prefix and baseUrl assuming (wrongly) that surely it behaves as expected. A tabular representation of the "concatenation" behavior might be a bit of an easier "catch" while reading the docs.

But again, I'm perfectly fine with not having that explained in a tabular manner.

onlywei:
Can either of these options accept a function? It would be really helpful to solve #805

sholladay:
I am open to supporting functions, but not in this PR. We can do that in a future feature release.

I'm going to add the alias and then get this merged.

sindresorhus:
@sholladay v2 work has started. Do you have time to continue work on this PR?

sholladay:
Yes. Do you still want an alias or should we do it as a breaking change for v2?

sindresorhus:
We can do it as a breaking change.

It should throw a clear runtime error when `prefixUrl` is used in v2, with an actionable message. Something like “prefixUrl was renamed to prefix in v2. Consider baseUrl for standard URL resolution semantics.” and a link to the docs.

sholladay:
@sindresorhus this PR should be ready to go. Added the runtime error and tried to make the docs clearer.

readme.md
```diff
@@ -191,29 +191,55 @@ Search parameters to include in the request URL. Setting this will override all
...
+//=> 'https://example.com/user'
+```
+
+##### prefix
+
+Type: `string | URL`
+
+A prefix to prepend to the `input` URL before making the request. It can be any valid path or URL, either relative or absolute. A trailing slash `/` is optional and will be added automatically, if needed, when it is joined with `input`. Only takes effect when `input` is a string.
+
+Useful when used with [`ky.extend()`](#kyextenddefaultoptions) to create niche-specific Ky-instances.
```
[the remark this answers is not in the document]
sholladay: I think using either to create niche instances is perfectly valid. But `baseUrl` should be the default, preferred approach. Updated the docs to make that explicit.

sindresorhus:
Looks good. Thanks for doing this 🙏

PR #610, opened 2024-07-16
Use KyRequest and KyResponse types in errors

### LINKED ISSUE #584: HTTPError json as unknown/type
I like how for requests I can chain `.json<Foo>()` or `.json()` (for `unknown` instead of `any`).

Can this same functionality be added when grabbing an HTTP response body from an `HTTPError` instance?

mfulton26:
- introduces `KyRequest` which, like `KyResponse`, adds `json()` convenience method for extracting JSON with an expected TypeScript type
- update API `Request`/`Response` usages to expose `KyRequest`/`KyResponse` to `ky` consumers so that they may use the `json()` convenience method in more scenarios (`HTTPError`, `TimeoutError`, and hooks)

Closes #584

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,6 +71,7 @@ It's just a tiny file with no dependencies.
 - URL prefix option
 - Instances with custom defaults
 - Hooks
+- TypeScript niceties (e.g. `.json()` resolves to `unknown`, not `any`; `.json<T>()` can be used too)
 
 ## Install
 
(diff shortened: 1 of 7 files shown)


sholladay:
LGTM, but another TS user review, too.

Do we have a decent place to document this? Perhaps the readme should mention it.

sholladay:
Yeah, I don't think the documentation about TS needs to be extensive. Autocomplete should be sufficient for most users. But I think it would be good to mention it somewhere, maybe as a feature in the benefits section at the top of the readme.

mfulton26:
@sholladay I took a stab at mentioning these TS improvements in the readme's benefits section. I tried to keep it short like the other bullet points. Suggestions welcome (more words, less, etc.).

sholladay:
Good enough for me. We can tweak it later if needed. I'm going to merge this now since we have quite a few open PRs and I want to avoid creating unnecessary conflicts by leaving things open.

Thanks again!

PR #611, opened 2024-07-16
Add support for passing a function to ky.extend()

### LINKED ISSUE #586: Allow extend() to append to prefixUrl
I'd like a way to use `extend()` and append to the existing `prefixUrl`:

```ts
const api = ky.create({ prefixUrl: "https://www.example.com/api" });
const sales = api.extend({ path: "sales/v1" });
await sales.post("offers", { json: {} }); // POSTs to https://www.example.com/api/sales/v1/offers
```

Today if I use `prefixUrl` in an `extend()` call then it overwrites the "base" `prefixUrl` and I don't see any way to access the configured `prefixUrl` from the "base" `KyInstance`. Is this already possible and I've overlooked it?

mfulton26:
callback function receives parent's default options to allow extended instance to refer to modify parent options

Closes #586

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -510,6 +510,22 @@ console.log('unicorn' in response);
 //=> true
 ```
 
+You can also refer to parent defaults by providing a function to `.extend()`.
+
+```js
+import ky from 'ky';
+
+const api = ky.create({prefixUrl: 'https://example.com/api'});
+
+const usersApi = api.extend((options) => ({prefixUrl: `${options.prefixUrl}/users`}));
+
+const response = await usersApi.get('123');
+//=> 'https://example.com/api/users/123'
+
+const response = await api.get('version');
+//=> 'https://example.com/api/version'
+```
+
 ### ky.create(defaultOptions)
 
 Create a new Ky instance with complete new defaults.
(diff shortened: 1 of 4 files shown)


sholladay:
Thanks for working on this, looking good.

Can you add a few more permutations to the tests? In particular, I want to make sure we cover:
1. What happens when the extend function ignores the parent options, i.e. is the `prefixUrl` still set if extend _doesn't_ set it?
2. The behavior of setting nested options, such as hooks. Does setting a hook add to or replace the parent's hooks?

readme.md
```diff
@@ -510,6 +510,22 @@ console.log('unicorn' in response);
...
+const api = ky.create({prefixUrl: 'https://example.com/api'});
+
+const usersApi = api.extend((options) => ({prefixUrl: `${options.prefixUrl}/users`}));
+
+const response = await usersApi.get('123');
+//=> 'https://example.com/api/users/123'
+
+const response = await api.get('version');
+//=> 'https://example.com/api/version'
+```
```
sindresorhus, line 527: It would be nice to include this part in the `.extends` TS doc comment too.

sholladay:
Such a simple change, yet I think this will be a popular feature and cleanly solve a lot of use cases. Thanks again!

mfulton26:
> Such a simple change, yet I think this will be a popular feature and cleanly solve a lot of use cases. Thanks again!

I agree. Thank you for helping me add this change. I look forward to using it in my own projects!

sindresorhus:
https://github.com/sindresorhus/ky/releases/tag/v1.5.0

joshgummersall:
Ah wow, thanks so much for getting to this to fast! Appreciate it 😄

PR #613, opened 2024-07-29
Fix signal handling for Request objects

crisshaker:
Currently, ky ignores `AbortSignal` when it's provided as part of a `Request` object.

This PR modifies the `Ky` class to properly extract and respect the `AbortSignal` from `Request` objects. It ensures that signals are handled consistently whether they're provided directly in the options or as part of a `Request` object.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -129,8 +129,11 @@ export class Ky {
 				? this._input.credentials
 				: undefined;
 
+		const signal = this._input instanceof Request ? this._input.signal : undefined;
+
 		this._options = {
 			...(credentials && {credentials}), // For exactOptionalPropertyTypes
+			...(signal && {signal}), // For exactOptionalPropertyTypes
 			...options,
 			headers: mergeHeaders((this._input as Request).headers, options.headers),
 			hooks: deepMerge<Required<Hooks>>(
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -169,7 +169,7 @@ export class Ky {
 
 		if (supportsAbortController) {
 			this.abortController = new globalThis.AbortController();
-			const originalSignal = this._options.signal || (this._input as Request).signal;
+			const originalSignal = this._options.signal ?? (this._input as Request).signal;
 			if (originalSignal) {
 				originalSignal.addEventListener('abort', () => {
 					this.abortController!.abort(originalSignal.reason);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for working on this. This needs a test.

sindresorhus:
I also don't think we should have ad-hoc handling for each property of `Request`. We need some common logic that merges request properties and options in a more general way.

crisshaker:
> I also don't think we should have ad-hoc handling for each property of `Request`. We need some common logic that merges request properties and options in a more general way.

I totally agree with this. Is there something similar in place I could reference?

sholladay:
> We need some common logic that merges request properties and options in a more general way.

That's why we normalize them [here](https://github.com/sindresorhus/ky/pull/613/files#diff-1d28e040cea335f5a22aa9c3f678f5d00b353e3ce966dc0978a91fe56f240b04R196) with `new Request(input, options)`.

The only special cases I can think of that are not handled naturally by that normalization are:
 - For Ky to set a request option's default value, we must only apply the default if `input` does not already have that option, otherwise our default will override it. We don't really have any such defaults at the moment (but we did in the past and I've proposed a default for `mode` in PR #602).
 - For `headers`, Ky merges `input` and `options`, which is something that `fetch()` does not do (with `fetch()`, providing any `headers` in `options` replaces all `headers` in `input`).
 - For `signal`, Ky needs to not only apply a default, but override any existing `signal` that was provided in `input` or `options`. We do this so we can have our own AbortController for timing out the request, as `fetch()` doesn't have a built-in timeout option.
 - For `duplex`, Ky always sets it to `half` if the environment supports request streams. This is done because you can't use request streams without it. IMO, it ought to be a built-in default but I guess they wanted to make it explicit.
 - For `credentials`, Ky used to set a default value but PR #559 changed It to `undefined`. We should be able to remove all special handling and references to it now, I think this was just overlooked during code review.

There isn't much that these cases have in common. But if anyone can think of a cleaner way to do all of this, I'd love to see it!

In this particular case with `signal`, we already modify the options object [here](https://github.com/sindresorhus/ky/pull/559/files#diff-1d28e040cea335f5a22aa9c3f678f5d00b353e3ce966dc0978a91fe56f240b04R170). There's no need to do it twice. The correct fix is simply to update `originalSignal` to also check `input.signal`.

sholladay:
I updated the code to use the simpler approach that I mentioned. PTAL.

Once a test is added, this should be good to go.

PR #618, opened 2024-08-11
Add support for the `RateLimit-Reset` header

### LINKED ISSUE #608: Use `X-RateLimit-Reset` header to determine retry delay
Used by GitHub and X, apparently:

https://medium.com/@guillaume.viguierjust/rate-limiting-your-restful-api-3148f8e77248

Also proposed (and expired) to IETF:

https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/

It seems that the rate-limiting retry logic is already in ky, but no mention of these headers was found in the repo.

sholladay:
Closes #608

This PR adds automatic handling for non-standard rate limiting headers that are used by some APIs (e.g. GitHub and Twitter) instead of `Retry-After`.

The following headers are now supported as a fallback for `Retry-After`:
 - `RateLimit-Reset` as proposed by https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-05.html#section-3.3
 - `X-RateLimit-Reset` as implemented by GitHub
 - `X-Rate-Limit-Reset` as implemented by Twitter

Dates, timeouts (delta seconds), and timestamps (seconds since epoch) are all supported.

Timestamps in milliseconds are not currently supported, as none of the popular APIs use it.

More variants are discussed here: https://github.com/ietf-wg-httpapi/ratelimit-headers/issues/25

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -219,11 +219,15 @@ export class Ky {
 
 			const retryAfter = error.response.headers.get('Retry-After')
 				?? error.response.headers.get('RateLimit-Reset')
-				?? error.response.headers.get('X-RateLimit-Reset');
+				?? error.response.headers.get('X-RateLimit-Reset') // GitHub
+				?? error.response.headers.get('X-Rate-Limit-Reset'); // Twitter
 			if (retryAfter && this._options.retry.afterStatusCodes.includes(error.response.status)) {
 				let after = Number(retryAfter) * 1000;
 				if (Number.isNaN(after)) {
 					after = Date.parse(retryAfter) - Date.now();
+				} else if (after >= Date.parse('2024-01-01')) {
+					// Large numbers are treated as time since the UNIX epoch (fixed threshold protects against clock skew)
+					after -= Date.now();
 				}
 
 				const max = this._options.retry.maxRetryAfter ?? after;
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -226,7 +226,7 @@ export class Ky {
 				if (Number.isNaN(after)) {
 					after = Date.parse(retryAfter) - Date.now();
 				} else if (after >= Date.parse('2024-01-01')) {
-					// Large numbers are treated as time since the UNIX epoch (fixed threshold protects against clock skew)
+					// A large number is treated as a timestamp (fixed threshold protects against clock skew)
 					after -= Date.now();
 				}
 
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -217,7 +217,9 @@ export class Ky {
 				throw error;
 			}
 
-			const retryAfter = error.response.headers.get('Retry-After');
+			const retryAfter = error.response.headers.get('Retry-After')
+				?? error.response.headers.get('RateLimit-Reset')
+				?? error.response.headers.get('X-RateLimit-Reset');
 			if (retryAfter && this._options.retry.afterStatusCodes.includes(error.response.status)) {
 				let after = Number(retryAfter) * 1000;
 				if (Number.isNaN(after)) {
--- a/readme.md
+++ b/readme.md
@@ -208,7 +208,7 @@ An object representing `limit`, `methods`, `statusCodes`, `afterStatusCodes`, an
 
 If `retry` is a number, it will be used as `limit` and other defaults will remain in place.
 
-If the response provides an HTTP status contained in `afterStatusCodes`, Ky will wait until the date or timeout given in the [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header has passed to retry the request. If `Retry-After` is missing, the non-standard [`RateLimit-Reset`](https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#section-3.3) header is used in its place as a fallback. If the provided status code is not in the list, the [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header will be ignored.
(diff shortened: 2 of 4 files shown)


sholladay:
@fregante does this satisfy your requirements?

source/core/Ky.ts
```diff
@@ -217,7 +217,9 @@ export class Ky {
 				throw error;
 			}
 
-			const retryAfter = error.response.headers.get('Retry-After');
+			const retryAfter = error.response.headers.get('Retry-After')
+				?? error.response.headers.get('RateLimit-Reset')
+				?? error.response.headers.get('X-RateLimit-Reset');
```
fregante, line 222: I don’t think this correct. On mdn I see two formats for `Retry-After` an ISO date or a number of seconds to wait

```
Retry-After: Wed, 21 Oct 2015 07:28:00 GMT
Retry-After: 120
``` 

While [GitHub's docs](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately) say:

> should not make another request until after the time specified by the x-ratelimit-reset header. The x-ratelimit-reset header is in UTC epoch seconds.

So this introduces a third type: UTC epoch. I think this is easy to detect though: `if t is number and t > 1700000000`

sholladay: I used the semantics defined by the IETF draft:
https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#name-ratelimit-reset

> The header value is
>
> ```
> RateLimit-Reset = delta-seconds
> ```
>
> The delta-seconds format is used because:
> -  it does not rely on clock synchronization and is resilient to clock adjustment and clock skew between client and server (see [[RFC7231](https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#RFC7231)] Section 4.1.1.1);
> - it mitigates the risk related to thundering herd when too many clients are serviced with the same timestamp.

> An example of RateLimit-Reset use is below.
> ```
> RateLimit-Reset: 50
> ```

Looks like neither X.com nor GitHub actually follow this proposal.

I also just noticed that X.com uses a slightly different header name: `x-rate-limit-reset` (note the `-` between `rate` and `limit`)

fregante (approved):
🙏

source/core/Ky.ts
sholladay: I found this very useful thread regarding header names and semantics:
https://github.com/ietf-wg-httpapi/ratelimit-headers/issues/25

I'll leave it to others to add some of the more obscure formats, if the need arises.

source/core/Ky.ts
```diff
@@ -217,11 +217,17 @@ export class Ky {
...
-			const retryAfter = error.response.headers.get('Retry-After');
+			const retryAfter = error.response.headers.get('Retry-After')
+				?? error.response.headers.get('RateLimit-Reset')
+				?? error.response.headers.get('X-RateLimit-Reset') // GitHub
+				?? error.response.headers.get('X-Rate-Limit-Reset'); // Twitter
 			if (retryAfter && this._options.retry.afterStatusCodes.includes(error.response.status)) {
 				let after = Number(retryAfter) * 1000;
 				if (Number.isNaN(after)) {
 					after = Date.parse(retryAfter) - Date.now();
+				} else if (after >= Date.parse('2024-01-01')) {
```
MathiasWP, line 228: Why is this hardcoded to "2024-01-01"?

sholladay: It is stupid, yes.

See the comment on the next line (the linter would not let me put it above).

Some `RateLimit-Reset` implementations in the wild use delta seconds, while others use timestamps. Both are integers, with no easy way to disambiguate them other than, essentially, how large the number is. Generally speaking, we would expect delta seconds to be a small number (because it's referring to seconds _from now_) and a timestamp to be a large number (because it's referring to seconds _from epoch_). Ky wants to normalize these values to delta seconds. So we need to detect timestamps and convert them.

The most obvious approach might be to check if `after` is close to `Date.now()`, but that is susceptible to clock skew and incorrectly set clocks. Such bugs also have a history of leading to security vulnerabilities. So, we use a constant threshold instead.

How large should `after` be for us to consider it a timestamp? That's... very arbitrary. @fregante suggested we use `1700000000` in this comment: https://github.com/sindresorhus/ky/pull/618#discussion_r1713150598

`1700000000` milliseconds is a little less than 20 days, which should be fine in the vast majority of cases. But if someone wants to set a 6 month retry delay and hope the computer stays up for that long... who am I to stop them? I figured that `Date.parse('2024-01-01')`, being the beginning of the year in which this feature was implemented, is hopefully more readable, allows for extremely long retry delays, and makes it obvious that the value is arbitrary and not based on some standard (I don't think there _is_ a standard for this).

I'm very open to improvements here, suggestions welcome! Would be nice to put it in a `const` with a self-documenting name, but I couldn't think of a good name.

fregante: > `1700000000` milliseconds is a little less than 20 days

`Retry-After` uses seconds, not milliseconds. Since this value is meant to differentiate that header from an epoch-based value, this is fine. No server will ask you to retry [in 53 years](https://www.wolframalpha.com/input?i=1700000000+seconds) (unless you're somehow querying [Deep Thought](https://en.wikipedia.org/wiki/List_of_The_Hitchhiker%27s_Guide_to_the_Galaxy_characters#Deep_Thought))


https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

sholladay: `Retry-After` and `RateLimit-Reset` both use seconds ([except for Mulesoft](https://docs.mulesoft.com/gateway/latest/policies-included-rate-limiting-sla), apparently 😡).

However, we [convert them](https://github.com/sindresorhus/ky/blob/a7204c4672d10a144b12ef169c8d38e17db81931/source/core/Ky.ts#L225) both to milliseconds. Indeed, `1700000000` seconds is plenty high enough. I could've just added a few more zeros for milliseconds. I just thought `Date.parse('2024-01-01')` (which is `1704067200` seconds) had a certain logic to it. Perfectly happy to change it, though.

PR #619, opened 2024-08-13
Add JSON generics to ky() and HTTPError

### LINKED ISSUE #485: improve documentation for generic parameters in README
### Rationale
https://github.com/sindresorhus/ky/blob/b86ff0a4e8d66c363df17612c968c023560f9241/index.test-d.ts#L132
[`json`](https://github.com/sindresorhus/ky#json) function can accept type parameters. however issue #312 was raised because `README.md` did not address which function accepts generic parameters.

### Solution
I suggest adding more documentation for functions that accept generic parameters.

### LINKED ISSUE #615: Add generic for HTTPError to type the response
Hello. I use ky with @tanstack/svelte-query, but sometimes when working with forms I need to type the error to pass it to the @tanstack/svelte-query library. 

For example:

```
export const useEditProfileMutation = (profileId: string) => {
  const client = useQueryClient();

  return createMutation<Profile, HTTPError<ValidationErrorsProfile>, Profile>({
    mutationFn: (profile: Profile) =>
      api.put(`member/${profileId}`, { json: profile }).json<Profile>(),
    onSuccess: () => {
      client.invalidateQueries({ queryKey: [QUERY_KEY_PROFILE, profileId] });
    },
  });
};

```

Generic for HTTPError that types KyResponse. It would be convenient.

```
export declare class HTTPError<T = unknown> extends Error {
    response: KyResponse<T>;
    request: KyRequest;
    options: NormalizedOptions;
    constructor(response: Response, request: Request, options: NormalizedOptions);
}
```

Or am I doing everything wrong and there is a right way?

sholladay:
Closes #615
Closes #485
Relates to #312 

This PR adds support for an optional TypeScript generic type parameter in all request methods, passes it through the request, response promise, and response, and into `.json()`, so that users have more flexibility in where and how they type their JSON objects.

All of the following forms are supported:
 - `ky<T>()`
 - `ky.get<T>()` (all methods)
 - `ky().json<T>()`
 - `(await ky()).json<T>()`

The `HTTPError` class has similarly been updated and passes its type parameter through to `error.response.json()`. See #615 for an example of how this could be used.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -57,7 +57,7 @@ It's just a tiny package with no dependencies.
 - URL prefix option
 - Instances with custom defaults
 - Hooks
-- TypeScript niceties (e.g. `.json()` resolves to `unknown`, not `any`; `.json<T>()` can be used too)
+- TypeScript niceties (e.g. `.json()` supports generics and defaults to `unknown`, not `any`)
 
 ## Install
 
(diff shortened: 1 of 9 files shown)


nielsdB97 (approved):
Looks good! Thanks for your efforts on getting this done 😄

mfulton26:
This is neat. What do you think of adding a second generic type parameter for the expected Error payload too?

```ts
try {
  // user is a User
  const user = await ky<User, UserGetError>('/api/users/2').json();
} catch (exception) {
  if (exception instanceof HTTPError) {
    // error is UserGetError
    const error = await error.response.json();
  }
}
```

sholladay:
Seems like a good idea. 👍

mfulton26:
Oh, it might already be there, kind of. Plus, I don't think there's a way in TypeScript to propagate a generic type parameter from a method call through to an exception that is thrown and it show up in the `catch()` block.

I think this works as is though:

```ts
try {
    // user is a User
    const user = await ky<User>("/api/users/2").json();
} catch (exception) {
    if (exception instanceof HTTPError) {
        // exception is UserGetError
        const error = await exception.response.json<UserGetError>();
    }
}
```

In fact, I see that `<T = unknown>` was added to `HTTPError`. Is that usable though?

mfulton26:
> I see that `<T = unknown>` was added to `HTTPError`. Is that usable though?

I suppose when creating instances manually it is.

sholladay:
It was added because of #615. See that issue for how and why you might use the generic for `HTTPError`.

And if TypeScript ever does allow functions to type their exceptions, `HTTPError` will be ready for it.

PR #628, opened 2024-09-02
Mark `ky.create()` options as optional in public types

dirkluijk:
Just a small tweak on the `ky.create()`. I know it's also possible to use the global `ky` instance, but just in case one wants to use this factory method instead.


sholladay:
This actually makes the types more consistent because the `create()` options are already marked optional here:

https://github.com/sindresorhus/ky/blob/a7e273617eb4728456d966201dbf50af30fb8e35/source/index.ts#L19

Thanks!

PR #632, opened 2024-09-06
Add onUploadProgress option

jadedevin13:


source/core/Ky.ts
```diff
@@ -365,4 +383,66 @@ export class Ky {
...
+				return acc;
+			}, 0);
+		}
+		if ('byteLength' in body) {
+			return (body as globalThis.ArrayBufferView).byteLength;
+		}
+		return 0; // Default case, unable to determine size
+	}
+
+	protected _wrapBodyWithUploadProgress(
```
sholladay, line 418: I think it would make more sense for this function to have a similar signature to `_stream()`, i.e. take two arguments, a request and `onUploadProgress`, and return a request.

We could rename these functions `_streamRequest()` and `_streamResponse()`.

sholladay (changes requested):
Thank you for Wii on this!

We will also need tests.

source/core/Ky.ts
```diff
@@ -203,6 +203,24 @@ export class Ky {
 			// The spread of `this.request` is required as otherwise it misses the `duplex` option for some reason and throws.
 			this.request = new globalThis.Request(new globalThis.Request(url, {...this.request}), this._options as RequestInit);
 		}
+
+		// Add onUploadProgress handling
+		if (this._options.onUploadProgress && typeof this._options.onUploadProgress === 'function') {
+			if (!supportsRequestStreams) {
+				throw new Error('Streams are not supported in your environment. `ReadableStream` is missing.');
```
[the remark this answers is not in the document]
jadedevin13: This is a copy of the message in onDownloadProgress. Should I change that as well?

source/types/options.ts
```diff
@@ -12,6 +12,16 @@ export type HttpMethod = 'get' | 'post' | 'put' | 'patch' | 'head' | 'delete';
 
 export type Input = string | URL | Request;
 
+export type UploadProgress = {
```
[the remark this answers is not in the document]
jadedevin13: Changed to Progress

source/core/Ky.ts
```diff
@@ -203,6 +203,24 @@ export class Ky {
...
+			}
+
+			const originalBody = this.request.body;
+			if (originalBody) {
+				const totalBytes = this._getTotalBytes(originalBody);
+				this.request = new Request(this.request, {
+					body: this._wrapBodyWithUploadProgress(originalBody, totalBytes, this._options.onUploadProgress),
+					headers: this.request.headers,
+					method: this.request.method,
+					signal: this.request.signal,
```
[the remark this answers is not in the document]
jadedevin13: No reason. Changed it to 

				this.request
					= new globalThis.Request(this._input, {
						...this._options,
						body: this._streamRequest(
							originalBody, totalBytes, this._options.onUploadProgress),
					});

source/core/Ky.ts
jadedevin13: Changed to _streamRequest and _streamResponse

source/core/Ky.ts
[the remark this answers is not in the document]
sholladay: No, there is a difference in what causes an environment to lack support for request streams vs response streams. See the implementation of the support constants. In particular, for an environment to support request streams, it needs to have a `Request` option named `duplex`, which we have feature detection for. That option is not relevant to response streams, which are much more widely supported.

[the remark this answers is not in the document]
jadedevin13: Changed to 'Request streams are not supported in your environment. The `duplex` option for `Request` is not available.'. Would this error message be enough?

source/core/Ky.ts
hckhanh: Can we merge it @sholladay?

sholladay: I haven't had sufficient time to do a proper review or to try it out. And I probably won't for about a week or so. Maybe @sindresorhus or @szmarczak can take a look, otherwise I'll get to it as soon as I can.

readme.md
```diff
@@ -421,6 +421,37 @@ const response = await ky('https://example.com', {
...
+```
+
+Notes:
+- This option requires environment support for `ReadableStream`. If streams are not supported, an error will be thrown.
+- The `body` of the request must be set for this option to have an effect.
+- The total size calculation is an approximation for certain types of bodies (e.g., `FormData`).
+- For `FormData` bodies, the size calculation includes an estimation for multipart boundaries and field names.
+- If the total size cannot be determined, `totalBytes` will be 0, and `percent` will remain at 0 throughout the upload and will be 1 once upload is finished.
+
+This feature is useful for tracking the progress of large file uploads or when sending substantial amounts of data to a server.
```
sindresorhus, line 453: The TS doc comments and readme should be in sync.

RomainLanz:
Hey there! 👋🏻

Just checking in—are there any blockers I can help with to move this PR forward for merging?

source/core/Ky.ts
```diff
@@ -365,4 +383,99 @@ export class Ky {
...
+		if ('byteLength' in body) {
+			return (body).byteLength;
+		}
+
+		if (typeof body === 'object' && body !== null) {
+			try {
+				const jsonString = JSON.stringify(body);
+				return new TextEncoder().encode(jsonString).length;
+			} catch (error) {
+				console.warn('Unable to stringify object:', error);
```
sholladay, line 437: As a library, it's generally considered bad practice for Ky to output non-errors to the console by default.

sholladay (changes requested):

sholladay:
I think this is ready to go once that warning is removed.

source/core/Ky.ts
jadedevin13: removed

sindresorhus:
CI is failing

sindresorhus:
Bump :)

georgiosd:
@sholladay would be amazing to merge this!

stefanoverna:
Thanks so much @sholladay! By any chance, do you have any idea when this feature will be released?

jeanmatthieud:
I tried this new feature. Works well with Chrome, got an error with Firefox:
"Error: Request streams are not supported in your environment. The `duplex` option for `Request` is not available."

Is it possible to access the "supportsRequestStreams" variable manage this case without an upload error (set or not the upload callback)? In my case, the upload progressbar is a "nice to have"...

sholladay:
There is currently no way to suppress that error, other than catching it.

I was on the fence about throwing an error in unsupported environments. It is probably an optional feature for most people. But Ky really has no way to know that.

The only real alternative that I see is to call the progress callback at 0% and 100%.

PR #641, opened 2024-09-29
Removed The Need For .js Extension For Imports

TheDokT0r:
Not a big change, but I've made it so there wouldn't be a need to add the ".js" extension to the end of each file import. To implement this change I've added minor changes to the tsconfig file, and added a few more eslint rules.


sholladay:
I don't have strong feelings on this, but personally I'd rather have the file extensions. It mashes the code more portable.

sindresorhus:
This is an invalid change. Even if it were valid, I prefer being explicit and have the extensions.

Issue #648, opened 2024-11-06
A ReadableStream branch was created but never consumed


HaidongPang:
I'm using ky in cloudflare worker with following code:

```typescript
import ky from "ky";
export default {
    async fetch (): Promise<Response> {
        const result = await ky.get("https://www.github.com").text();
        return new Response(result);
    }
} satisfies ExportedHandler;
```

and when I turn off local mode, requests raise the following warning:
```
A ReadableStream branch was created but never consumed. Such branches can be created, for instance, by calling the tee() method on a ReadableStream, or by calling the clone() method on a Request or Response object. If a branch is created but never consumed, it can force the runtime to buffer the entire body of the stream in memory, which may cause the Worker to exceed its memory limit and be terminated. To avoid this, ensure that all branches created are consumed.

 * Unused stream created:
    at result.<computed> [as text] (file:///....../node_modules/ky/source/core/Ky.ts:92:36)
......
```

Locate the following code snippet:

[source/core/Ky.ts](https://github.com/sindresorhus/ky/blob/main/source/core/Ky.ts)

```typescript
export class Ky {
	static create(input: Input, options: Options): ResponsePromise {
		const ky = new Ky(input, options);

		const function_ = async (): Promise<Response> => {
                         ......
		};

		......

		for (const [type, mimeType] of Object.entries(responseTypes) as ObjectEntries<typeof responseTypes>) {
			result[type] = async () => {
				......
				const awaitedResult = await result;
				const response = awaitedResult.clone();

				if (type === 'json') {
					if (response.status === 204) {
						return '';
					}

					const arrayBuffer = await response.clone().arrayBuffer();
					const responseSize = arrayBuffer.byteLength;
					if (responseSize === 0) {
						return '';
					}

					if (options.parseJson) {
						return options.parseJson(await response.text());
					}
				}

				return response[type]();
			};
		}

		return result;
	}
```

The `awaitedResult` is a variable of type Response. Since it is declared, it has been cloned only once, and its ReadableStream has never been consumed.
After `awaitedResult` has finished cloning, its ReadableStream should be closed to ensure memory safety.
Seems like:

```typescript
const awaitedResult = await result;
const response = awaitedResult.clone();
await awaitedResult.body.cancel();
```

sholladay:
PR welcome to fix this issue. If I'm not mistaken, canceling the stream could lead to lost data even though we've cloned the stream. We'd need to test it thoroughly to make sure that's not the case.

[...]

PR #649, opened 2024-11-07
Clean up orphan ReadableStream remaining during parsing response (#648)

HaidongPang:


gregory:
@HaidongPang i was just looking into something similar, [it appears that the response is cloned before being passed to each afterResponse hooks](https://github.com/sindresorhus/ky/blob/main/source/core/Ky.ts#L45) which will create ReadableStream branches, which will force the runtime to buffer the entire body of the stream in memory multiple times without certainty that it'll ever be consumed... 

@sindresorhus Could you elaborate on why you felt that you had to do that? 

Thanks for your amazing work btw 🙏

HaidongPang:
@gregory

I suspect that @sindresorhus does not want the ReadableStream of response to be consumed in the afterResponse hooks, which would result in incomplete response data being returned.

In short, you can view the response object that is exactly the same as the original response in afterResponse hooks, but you should not modify the original response.

gregory:
> @gregory
> 
> I suspect that @sindresorhus does not want the ReadableStream of response to be consumed in the afterResponse hooks, which would result in incomplete response data being returned.
> 
> In short, you can view the response object that is exactly the same as the original response in afterResponse hooks, but you should not modify the original response.

I understand, but that would be the responsibility of the consumer. 
Ie: in the hook: 
```
(req, {context}, res) => {
const headers = res.headers() 
const clone = res.clone()
}```

sholladay:
We could leave cloning to the consumer, but it's not very ergonomic and definitely a footgun of its own.

One possible solution might be to override the body methods so they first clone the response before consuming the stream.

gregory:
> We could leave cloning to the consumer, but it's not very ergonomic and definitely a footgun of its own.
> 
> One possible solution might be to override the body methods of the response given to hooks, so they first clone the response before consuming the stream.

that's a good idea.

HaidongPang:
Will these two patches be considered for merging in future versions?
Related to :[#650](https://github.com/sindresorhus/ky/issues/650)

sholladay:
Closing in favor of #661.

After careful review, I determined that we don't need to clone the response at all in this particular case. 😀

PR #651, opened 2024-11-07
Fix memory leak, clean up body streams after retry

### LINKED ISSUE #650: Orphan ReadableStream remaining after retry
Cloudflare workers raise the following warning:
```
A ReadableStream branch was created but never consumed. Such branches can be created, for instance, by calling the tee() method on a ReadableStream, or by calling the clone() method on a Request or Response object. If a branch is created but never consumed, it can force the runtime to buffer the entire body of the stream in memory, which may cause the Worker to exceed its memory limit and be terminated. To avoid this, ensure that all branches created are consumed.

 * Unused stream created:
    at Ky._fetch (file:///....../node_modules/ky/source/core/Ky.ts:303:30)
    at function_ (file:///....../node_modules/ky/source/core/Ky.ts:38:28)
    at async Ky._retry (file:///....../node_modules/ky/source/core/Ky.ts:256:11)
```

The request object that was cloned in advance for retry preparation still remains after returning a response that does not require a retry.

```typescript
export class Ky {
	protected async _fetch(): Promise<Response> {
		......

		// Cloning is done here to prepare in advance for retries
		const mainRequest = this.request;
		this.request = mainRequest.clone();

		if (this._options.timeout === false) {
			return this._options.fetch(mainRequest, nonRequestOptions);
		}

		return timeout(mainRequest, nonRequestOptions, this.abortController, this._options as TimeoutOptions);
	}
}
```

Related to :[#648](https://github.com/sindresorhus/ky/issues/648)

HaidongPang:
Fixes #650

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 18
+          - 22
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
 		"@types/busboy": "^1.5.0",
 		"@types/express": "^4.17.17",
 		"@types/node": "^20.14.12",
-		"ava": "^6.2.0",
+		"ava": "^5.3.1",
 		"body-parser": "^1.20.2",
 		"busboy": "^1.6.0",
 		"del-cli": "^5.1.0",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
 		"@types/busboy": "^1.5.0",
 		"@types/express": "^4.17.17",
 		"@types/node": "^20.14.12",
-		"ava": "^5.3.1",
+		"ava": "^6.2.0",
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 18
+          - 22
```
dasa, line 13: Should this have be updated too?
https://github.com/sindresorhus/ky/blob/93c02ac7817df1138e97e8971fe5165e30d68097/package.json#L22

sholladay: No. But I should have added 22, instead of replacing 18 here.

We were experiencing some strange CI failures that I couldn't replicate locally, so I tried updating Node here to see if that had any effect, which it didn't. In the end, it seems that GitHub Actions was just having a bad day, as it resolved itself. I didn't think about adding 18 back to the testing matrix, but that would be a good idea.

Nothing has changed in Ky to require users update Node. However, Node 18 is reaching end of life at the end of this month. So we won't be supporting it soon, anyway.

Issue #660, opened 2024-12-05
If request starts after the signal has already been aborted, it won't throw an error


purpurplr:
If request starts after the signal has already been aborted, it won't throw an error, although it probably should.

```
const controller = new AbortController();

controller.abort('no reason');

ky.get('https://jsonplaceholder.typicode.com/todos/1', { signal: controller.signal })
  .json()
  .then((json) => console.log(json));
```

sindresorhus:
Agreed

sholladay:
Agreed.

Does it actually send the request or does it silently fail? Also, what does fetch do?

purpurplr:
Ky sends the request, fetch throws an error.

sholladay:
I believe this is happening because when we [create our about controller](https://github.com/sindresorhus/ky/blob/3ba40cc6333cf1847c02c51744e22ab7c04407f5/source/core/Ky.ts#L167), we only listen for future aborts from the user's signal.

We need to either copy `originalSignal`'s existing state or use `throwIfAborted()` on it.

kadhirr:
Hey folks, I prepared a PR for this (with `throwIfAborted()`), can I get this issue assigned to me? I am facing an issue here with the tests though. Running tests seem to just throw the following error:
```
  TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for D:\code\github\ky\test\prefix-url.ts

  ✘ test\http-error.ts exited with a non-zero exit code: 1
  ✘ test\main.ts exited with a non-zero exit code: 1
  ✘ test\methods.ts exited with a non-zero exit code: 1
  ✘ test\prefix-url.ts exited with a non-zero exit code: 1

  Uncaught exception in test\retry.ts

  TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for D:\code\github\ky\test\retry.ts

  ✘ test\retry.ts exited with a non-zero exit code: 1
  ```
  Has anyone faced this issue?

sholladay:
I'm not sure if the tests have been run on Windows before.

My recommendation would be to use the Windows Subsystem for Linux, which is the official way to run *nix software on Windows. But we would probably also accept a patch to get the tests working on Windows, assuming it's something simple like a file path issue.

kadhirr:
[...]

PR #663, opened 2024-12-14
fix: throw early if signal is aborted before request is triggered

kadhirr:
ky does not seem to abort a request if the signal was aborted before the request was triggered ([ref](https://github.com/sindresorhus/ky/issues/660)). 

We fix this by calling the `throwIfAborted()` method on the original signal.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -166,10 +166,10 @@ export class Ky {
 			if (originalSignal?.aborted) {
 				this.abortController.abort();
 			}
+
 			originalSignal?.addEventListener('abort', () => {
 				this.abortController!.abort(originalSignal.reason);
 			});
-
 			this._options.signal = this.abortController.signal;
 		}
 
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -164,7 +164,7 @@ export class Ky {
 			this.abortController = new globalThis.AbortController();
 			const originalSignal = this._options.signal ?? (this._input as Request).signal;
 			if (originalSignal?.aborted) {
-				this.abortController.abort();
+				this.abortController.abort(originalSignal?.reason);
 			}
 
 			originalSignal?.addEventListener('abort', () => {
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -163,6 +163,7 @@ export class Ky {
 		if (supportsAbortController) {
 			this.abortController = new globalThis.AbortController();
 			const originalSignal = this._options.signal ?? (this._input as Request).signal;
+			originalSignal?.throwIfAborted();
 			originalSignal?.addEventListener('abort', () => {
 				this.abortController!.abort(originalSignal.reason);
 			});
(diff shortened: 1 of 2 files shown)


kadhirr:
I tried to add a test for this scenario
```
test('throws AbortError when signal was aborted before request', async t => {
	const server = await createHttpTestServer();
	// eslint-disable-next-line @typescript-eslint/no-empty-function
	server.get('/', () => {});

	const abortController = new AbortController();
	const {signal} = abortController;
	const request = new Request(server.url, {signal});
	abortController.abort();
	const response = ky(request);

	const error = (await t.throwsAsync(response))!;


	t.true(['DOMException', 'Error'].includes(error.constructor.name), `Expected DOMException or Error, got ${error.constructor.name}`);
	t.is(error.name, 'AbortError', `Expected AbortError, got ${error.name}`);
});
```

here I am facing the following issue
```
 ✘ [fail]: throws AbortError when signal was aborted before request Rejected promise returned by test
  ─

  throws AbortError when signal was aborted before request

  test/main.ts:724

   723:   const request = new Request(server.url, {signal});
   724:   abortController.abort();
   725:   const response = ky(request);

  DOMException {}

  › <anonymous> (test/main.ts:724:18)

  ─

  1 test failed
```

seems like the `abort()` line is rejecting the promise, but there is no object where the test will listen for a async throw. Can you guide me on how to proceed here?

sholladay:
I think the problem is that the `throwIfAborted()` call is happening inside of the constructor, which is not async or promise based. So the `ky()` call in the test isn't returning a rejected promise, but rather throwing a traditional synchronous exception. That immediately stops the test from executing and there's nothing to catch it, because `throwsAsync()` only handles promise based errors.

Perhaps we should just copy the signal's state instead of using `throwIfAborted()`. So basically abort our signal if the original is already aborted and be sure to copy over the reason.

That way `fetch()` will throw for us. And since that part of the code is wrapped in an async function, the error will inherently be returned via a promise and thus your test should pass, probably without any changes.

kadhirr:
@sholladay I tried to copy the state of the original signal, but all abortController properties are readonly, not sure how to copy the state here. So I've aborted the internal abortController if the passed in signal was also aborted. The tests also seem to work with this usage.
Please let me know if this is sufficient!

sholladay:
Yep, that's what I meant by copying the state. Could you also pass in the abort reason? Same as what's done in the event listener immediately below what you've written.

kadhirr:
Hey @sholladay , I've added the reason to the `abort()`, forgot that in the previous commit. Hope this looks good!

sholladay:
One last thing if you don't mind. I think it would be a good idea to fail the test if the server receives any requests.

kadhirr:
> One last thing if you don't mind. I think it would be a good idea to fail the test if the server receives any requests.

No issues @sholladay! I have modified the test to check if the server receives a request as well. Let me know if this looks good.

sholladay:
Great work, @kadhirr. Thanks!

PR #668, opened 2025-02-05
fix: set default method to GET

pecoram:
In Chrome versions earlier than 64, making a fetch request with `method: undefined` results in an error because the browser interprets it as `Access-Control-Request-Method: undefined`.

```ts
fetch("url", {
  method: undefined
});
```

This PR introduces a fallback to "GET" when no method is provided, ensuring the request continues to work without errors.


sindresorhus:
From the readme:

> Ky targets [modern browsers](https://github.com/sindresorhus/ky/pull/668#browser-support), Node.js, Bun, and Deno.

___

While this is indeed a simple change. It's a slippery slope.

PR #669, opened 2025-02-05
fix: avoid set method undefined

pecoram:
In Chrome versions earlier than 64, making a fetch request with `method: undefined` results in an error because the browser interprets it as `Access-Control-Request-Method: undefined`.

```ts
fetch("url", {
  method: undefined
});
```

This version has less impact than [setting "GET" by default](https://github.com/sindresorhus/ky/pull/668). The method is set only if it is not undefined.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -135,7 +135,6 @@ export class Ky {
 				},
 				options.hooks,
 			),
-			method: normalizeRequestMethod(options.method ?? (this._input as Request).method),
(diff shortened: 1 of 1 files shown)


sholladay:
Ky only supports the latest version of Chrome. Chrome v64 is more than 7 years old. How did you even encounter this bug?

pecoram:
I'm using ky on a TV APP and the problem occurs on many devices, especially Tizen and some older androidTVs. 
We have already fixed the bug on our side by making explicit method: “GET” but I thought with this PR I would help others who are not aware of the bug.

PR #683, opened 2025-03-25
Fix memory leak caused by shared abort signals

sebinsua:
When using a shared `AbortSignal` across multiple requests, event
listeners were being added but never removed. This resulted in
memory leaks as each request's context remained in memory indefinitely.

The fix uses [`AbortSignal.any()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static) to properly combine signals without
leaking memory, and eliminates a need for a manual event listener
approach. Additionally, this already handles the case where a signal is
already aborted, making the explicit `.aborted` check unnecessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
 		"delay": "^6.0.0",
 		"expect-type": "^0.19.0",
 		"express": "^4.18.2",
+		"jest-leak-detector": "^29.7.0",
 		"pify": "^6.1.0",
 		"playwright": "^1.45.3",
 		"raw-body": "^2.5.2",
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -168,9 +168,8 @@ export class Ky {
 		}
 
 		if (supportsAbortController) {
-			this.abortController = new globalThis.AbortController();
-
 			const originalSignal = this._options.signal ?? (this._input as Request).signal;
+			this.abortController = new globalThis.AbortController();
 			this._options.signal = originalSignal ? AbortSignal.any([originalSignal, this.abortController.signal]) : this.abortController.signal;
 		}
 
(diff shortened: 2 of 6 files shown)


sholladay:
Awesome! Thanks for writing a test.

Is there a good way to detect the leak without importing a dependency? For example, if duplicate event listeners were being installed, maybe they could increment a counter in the test, which could be asserted at the end to have the expected count. Or if that doesn't work because the old listeners aren't triggered, then maybe we could directly count how many listeners there are with `getEventListeners()`? Not sure if that method is supported by abort signals.

sebinsua:
The [logic for `jest-leak-detector` is actually very simple](https://github.com/jestjs/jest/blob/main/packages/jest-leak-detector/src/index.ts) so I could inline and simplify it if that would be preferred?

However, I'd recommend testing in this way instead of counting the number of event listeners, because otherwise you're testing the current cause of the leak instead of whether there is a leak or not, and (in an http client) you'll get more bang-for-buck by having a generalisable way of unit testing for leaks. Btw, I learnt the technique [reading the unit tests of `jotai`](https://github.com/pmndrs/jotai/blob/main/tests/vanilla/memoryleaks.test.ts).

FYI, The leak itself is due to the event handler closing over everything passed into `Ky.create` and then being attached onto _something that is potentially long-lived_ (e.g. `options.signal`). You can see in my test that without passing in an `AbortSignal` there isn't a 'leak' as there is nothing long-lived that ends up holding this closure and pointing towards this memory.

sholladay:
If the goal is to detect regressions for this specific issue, I think counting listeners would be perfectly fine.

If the goal is to create a more general purpose system for detecting leaks throughout the Ky tests, then I agree that checking actual memory usage is much better. However, in that case, it would be preferable to implement it as a reusable utility, perhaps as a [test macro](https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#reusing-test-logic-through-macros) so it can run assertions.

sholladay:
I'll merge this now so we can get the fix out. It would be nice to have this refactored as a utility, though.

PR #700, opened 2025-06-03
Workaround silent crash in node 20 on await stream cancel

brabeji:
Just an observation, maybe this is not an actual fix. On Node.js 20, awaiting promise returned from ReadableStream cancel silently crashes the process.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -87,7 +87,7 @@ export class Ky {
 			.finally(async () => {
 				// Now that we know a retry is not needed, close the ReadableStream of the cloned request.
 				if (!ky.request.bodyUsed) {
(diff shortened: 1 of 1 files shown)


sholladay:
Instead of removing the await, we could add a try catch. Does that solve it and prevent the process from crashing silently?

RamsesAupart (approved):

RamsesAupart (approved):

RamsesAupart (approved):

sholladay:
@brabeji can you please check the latest 20.x version, e.g. 20.19.4, and see if this fix is still valid/necessary? They've added a number of bugfixes recently that could be relevant.

Issue #688, opened 2025-04-03
Errors in Safari < 17.4 due to using AbortSignal.any


taras-turchenko-moc:
Hey guys. I've tried `1.8.0` version of the lib but was forced to downgrade it to `1.7.5` because of error reports about using `AbortSignal.any` that doesn't work in safari older than `17.4`. This version of safari was released in march of 2024 so I believe there is sense to support it too

[Can I use for AbortSignal.any](https://caniuse.com/mdn-api_abortsignal_any_static)

taras-turchenko-moc:
Here is usage of AbortSignal.any 
https://github.com/sindresorhus/ky/blob/93c02ac7817df1138e97e8971fe5165e30d68097/source/core/Ky.ts#L173

sholladay:
I sympathize but I'm afraid that is not how semver works. Bug fixes and security updates can change behavior that technically "breaks" user code, but they are not considered major revisions.

If a change forces us to modify existing tests, that's a good indication that it may deserves a major version bump, but even then, if the behavior is undocumented, it's really at our discretion whether we think it makes sense to do so.

In this case:
 - Our documentation clearly states support for "latest browsers".
 - Safari 17.4 is more than a year old and isn't even within the 18.x range.
 - We have not changed our API contract. If we did, that would be a major version bump.
 - The implementation details of how signals are combined is undocumented.
 - The change was made as a bug fix for a memory leak.

Also, perhaps most importantly, our current testing infrastructure doesn't give us the ability to run regression tests against old browsers. We could potentially integrate BrowserStack to do something like that. I am open to it if someone sends a PR. That would at least make it easier to be aware of the situation, regardless of whether we decide to officially change our support policy.

FishHawk:
Got it.

sholladay:
Safari has had `WeakRef` for even longer, since v14.1, which was released in April, 2021. I don't see this being an issue for

sholladay:
Safari has had `WeakRef` for even longer, since v14.1, which was released in April, 2021. This shouldn't be an issue for most people.

That said, if you need it, I would be happy to provide commercial support and discuss options for solving edge cases like this.

Saturate:
Also encountered this issue for visitors with older browsers that are not easy to update, mostly iOS devices.

Downgraded for now.

sholladay:
[...]

PR #701, opened 2025-06-05
Add polyfill for AbortSignal.any

hansottowirtz:
There are a lot of browsers where `AbortController` exists but `AbortSignal.any` doesn't. This PR adds a polyfill.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -15,6 +15,7 @@ import timeout, {type TimeoutOptions} from '../utils/timeout.js';
 import delay from '../utils/delay.js';
 import {type ObjectEntries} from '../utils/types.js';
 import {findUnknownOptions} from '../utils/options.js';
+import {abortSignalAny} from '../utils/abort.js';
 import {
 	maxSafeTimeout,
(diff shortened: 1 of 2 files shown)


hansottowirtz:
@sindresorhus Sorry, missed issue #688 and didn't realize it was this complicated. After reading up, I think there are a few options:
- Adding https://gist.github.com/CNSeniorious000/9fc1a72e45358dd7c9e2f16e5d26df5c as a polyfill to Ky
- Adding a simple polyfill that might leak memory to Ky (like in this PR and in [trpc client](https://github.com/trpc/trpc/blob/6743d6594ba9a8c4e7a66ce61b98b65be1485ec7/packages/server/src/unstable-core-do-not-import/utils.ts#L91-L119))
- Adding a note to the readme that an AbortSignal.any polyfill might be needed
- Including `typeof AbortSignal.any === 'function'` in `supportsAbortController` (that would remove signal support in iOS Safari >=11.3 <17.4, but as this browser is not targeted anymore it might be an option)

If you prefer any of these options, I can make a new PR for it.

PR #710, opened 2025-07-03
Clean up Ky constructor

0xlakshan:
No changes to functionality, just cleaned things up while keeping the execution flow the same. 

Makes the code easier to read and maintain 🦄


sholladay:
No strong feelings, but I personally find this style more complicated because I assume that someone put code in a function in order to reuse it, so I assume it has multiple entry points. And I have to go find the entry points to confirm that assumption and understand the control flow.

To simply label segments of code that are not reused, I think comments are better. But again, no strong feelings.

RamsesAupart (approved):

PR #711, opened 2025-07-17
Test support of AbortSignal.any()

pioug:
I took the idea from https://github.com/sindresorhus/ky/pull/701#issuecomment-2944963709. 

I think it follows the support policy explained in https://github.com/sindresorhus/ky/issues/688, and the code fits with the rest.


sholladay:
I'm ok with this. It basically bypasses the timeouts feature in environments that don't support `any()`.

Issue #689, opened 2025-04-03
ky aborting in 1.8.0


curtdept:
When using options ky seems to abort, provide no return and wont process and further requests.

example:
`import HttpRequestMock from "http-request-mock";
import ky from "ky";

const mockHttp = HttpRequestMock.setupForFetch();

mockHttp.post("https://api.site.test/call", {
  data: { payload: "testing" },
});

async function test() {
  const test1 = await ky.post(`https://api.site.test/call`, {
    json: { fingerprint: "12334" },
  });
  console.log("test1", test1);

  //   const test2 = await ky.post(`https://api.site.test/call`);
  //   console.log("test2:", test2);
}

test()
  .then(() => {
    process.exit(0);
  })
  .catch((err) => {
    console.error(err);
    process.exit(1);
  });
`

Scenarios:
- when running test1 alone, no return, seem to just skip
- when running test1 and test2, no return, both skip
- when running test2 alone, returns as expected

sholladay:
All tests for the release [passed](https://github.com/sindresorhus/ky/actions/runs/14165293807/job/39677339780). This is probably some kind of incompatibility between Ky and the `http-request-mock` library you are using.

> if I comment this out, all scenarios work

That code was added in PR #651 to fix #650. If that is truly the cause, then the fix will have to avoid re-introducing the memory leak. The only way I can think of to do that would be to re-construct the request from scratch on every retry, instead of storing a clone ahead of time. Currently, only [function_](https://github.com/sindresorhus/ky/blob/93c02ac7817df1138e97e8971fe5165e30d68097/source/core/Ky.ts#L32-L88) is retried. We would have to add at least something like [this line](https://github.com/sindresorhus/ky/blob/93c02ac7817df1138e97e8971fe5165e30d68097/source/core/Ky.ts#L186), perhaps even the entire `constructor`, to the retry loop. I'm okay with that, it's arguably the correct thing to do, but beware that it may be tricky. Feel free to send a PR if you want to give that a shot.

curtdept:
@sholladay ya can't tell you source. Just know 1.7.5 works and commenting works, but no idea why at this point.

natanhermes:
@curtdept @sholladay  I'm a beginner with ky and I'm having a similar problem.

```ts
const api = ky.create({
  prefixUrl: 'http://localhost:3333',
})

export async function signInWithEmailAndPassword(data: FormData) {
  console.log(Object.fromEntries(data))
[...]

PR #715, opened 2025-08-02
Add Node 20 and latest to CI tests

sholladay:
This PR ensures that the CI tests run on all [currently supported versions](https://nodejs.org/en/about/previous-releases#nodejs-releases) of Node to help catch upstream bugs, like the one discussed in #689.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 24
+          - latest
           - 22
           - 20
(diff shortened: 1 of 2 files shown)


adriandarian:
@sholladay shouldn't v18 be added since the package.json claims it works on v18
image

sholladay:
I think it's a good idea, if our dev dependencies support it. But IIRC, they may not.

Relatedly, I would prefer that we align with Node's own support schedule. But since we've already documented the minimum supported version as 18, I guess that should wait for v2.

Issue #719, opened 2025-08-13
Unable to use .json() without awaiting response first


dreamscached:
Hello! I really like the feature set of ky and would love to use it in my code, but I couldn't use its `await ky(...).json()` form due to the following error:

```
node:internal/deps/undici/undici:6201
        throw new TypeError("Body is unusable: Body has already been read");
              ^

TypeError: Body is unusable: Body has already been read
    at consumeBody (node:internal/deps/undici/undici:6201:15)
    at _Response.json (node:internal/deps/undici/undici:6156:18)
    at Promise.result.<computed> (/home/dream/.../node_modules/.store/ky-npm-1.8.2-b842d8ef20/package/source/core/Ky.ts:118:26)
```

I debugged the exception and it seems that this piece of code is the cause:
https://github.com/sindresorhus/ky/blob/17ca3894b5f2b1957f7255bd8a805b3f68e5e17c/source/core/Ky.ts#L118

I'm not entirely sure, as I struggle to reproduce this on some URL that I can provide (unfortunately, what I work with is a private project and I cannot share it here), but it might be that `response[type]()` is called multiple times, maybe?

The endpoint I use returns a response with some processing delay of 1-2 seconds and returns a large amount of JSON data. Maybe this helps reproducing it somehow.

Oddly enough, using typical

```js
const res = await ky(...);
const json = await res.json();
```

works as expected.

sholladay:
Interesting, I've never run into that before.

I wonder if this is related to #689. If we can get to the root cause of this, maybe we can solve both issues. Would you mind trying the older versions of Ky that are mentioned in that thread? If that works, it might be time to do a revert on the memory leak fixes.

dreamscached:
I tried using versions `1.8.0`, `1.7.5`, `1.6.0` and `1.5.0` and the issue persisted in all of them.

sindresorhus:
@dreamscached Can you check if 1.10.0 resolved this?

dreamscached:
It seems the issue is fixed, yes. Thank you! =)

I apologize for not providing a test URL to demonstrate it on.

PR #731, opened 2025-09-03
Fix race condition issue with `.json()`

sindresorhus:
This is a potential fix for #719. Even if it's not, it's still a nice simplification. So for that reason, I'm not going to wait for OP of #719 to test it out first.

Since it's not reproducible, I was not able to add a test.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -112,15 +112,16 @@ export class Ky {
 						return '';
 					}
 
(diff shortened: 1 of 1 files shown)


sholladay:
LGTM. I had previously thought this code needed to avoid consuming the body, but actually it's fine. And removing `.clone()` wherever possible is great. 👏

Issue #694, opened 2025-04-21
Upload progress is super fast, but it's fake.


recsater:
I tested with Chrome's Developer Tools -> Network tab, throttling the network to Fast 4G.

When uploading via XHR:

![Image](https://github.com/user-attachments/assets/dc25d427-a694-43f5-ad11-655ed6afb526)
![Image](https://github.com/user-attachments/assets/673ac3be-b11d-4d1e-9a2a-bfa7ea374f72)

When using Ky's uploadProgress:

![Image](https://github.com/user-attachments/assets/2115a253-ccc4-4423-866e-7f4f3b15fdae)
![Image](https://github.com/user-attachments/assets/f0e6c97e-0110-48f2-8425-9a274bd222ef)

The transfer speed of XHR is correct.

sholladay:
From what I remember, `FormData` is the one case where Ky has no idea what the size really is, because it doesn't have a `.length` or a `.size` property, unlike basically anything else

recsater:
> From what I remember, `FormData` is the one case where Ky has no idea what the size really is, because it doesn't have a `.length` or a `.size` property, unlike basically anything else you might send.
> 
> I suspect the implementation can be improved, though. PR welcome.

You're right, but it seems unrelated to this issue.

The problem isn't about correctly obtaining the size of totalBytes; it's that the onUploadProgress function is being called regardless of the actual upload speed.

Or does knowing the size of formData solve this issue?

sholladay:
I thought the flakey size estimation may have played a role in any discrepancy between your time estimate vs how long it actually took to upload. But now that I'm more awake, I think I understand that the issue is more that 100% progress is reached almost immediately even if the upload has barely started (whether or not the size is accurate). Is that right?

Looking at the code, I believe this is happening because we are optimistically pushing chunks onto the stream's buffer and not waiting for the buffer to be drained before adding more. That was good enough (but maybe not ideal) for the `onDownloadProgress` implementation because the network speed was the limiting factor for reading chunks, but it's not good enough for `onUploadProgress` because the chunks can be read faster than they can be sent. This definitely needs to be reworked.

recsater:
`I think I understand that the issue is more that 100% progress is reached almost immediately even if the upload has barely started (whether or not the size is accurate). Is that right?`

Yes, right.

sholladay:
Hi @recsater I am attempting to fix this issue. Would you mind testing the `pull-streams` branch from my fork to see if it helps?

```sh
[...]

PR #734, opened 2025-09-11
Improve body size estimation

Richienb:
Note that I needed to adjust some tests because they expect undershooting the size, but this is the first time the size is actually exactly right.

Possibly related to https://github.com/sindresorhus/ky/issues/694

### DIFF
--- a/source/core/constants.ts
+++ b/source/core/constants.ts
@@ -60,9 +60,9 @@ export const maxSafeTimeout = 2_147_483_647;
 
 // Used to help estimate upload size
 async function getFormBodySize(formData: FormData) {
-	const responseText = await new Response(formData).text();
+	const buffer = await new Response(formData).arrayBuffer();
 
-	return responseText.length;
+	return buffer.byteLength;
 }
 
 const formData = new FormData();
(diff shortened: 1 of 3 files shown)


sholladay:
@Richienb any further thoughts on those comments about the tests and/or using `Response` to calculate the body size? I'm willing to merge as-is, but those would be nice to have. The tests in particular. I could file an issue for follow up work if you don't have time at the moment. Thanks again for working on this!

Richienb:
Not sure what to do from here.
There might be a browser optimisation that streams the built arrayBuffer internally so it isn't all stored in memory.
To continue with the procedurally calculating formData size, maybe I could calculate the residual added by filename and content-type by drafting further `Response`s.
Yes, testing in other browsers would be a good idea.

sindresorhus:
Can we move this forward?

---

I suggest keep the current PR implementation (manual calculation) with two improvements:

1. Add inline comments explaining the FormData size calculation logic
1. refactor tests to use real server + Ky

sindresorhus:
I cherry-picked the useful parts (shared `TextEncoder`, `ArrayBuffer.isView()` check, `.byteLength` consistency) but I'm going to close this one.

The main blocker is top-level `await` in a library module. It can break consumers using bundlers or environments without TLA support. Beyond that, the body size is only used for `onUploadProgress` progress tracking, where the current approximation is good enough. The added complexity of dynamically measuring FormData serialization at module load time isn't worth the marginal accuracy gain for a progress bar.

PR #741, opened 2025-09-21
Fix hang on stream request cleanup in Node.js

### LINKED ISSUE #690: Ky 1.8.0 hang
We are experiencing a hang in Ky 1.8.0 that is a first for us. We have a pretty straightforward `ky.post(url, {json: someObj})` that works fine in 1.7.5. We tested under node 18 and node 22. We can run it locally on a Mac (M1), but when deployed to our docker container (Debian 12) the post hangs every time. We never get an exception or a timeout, and we aren't adding any hooks or any other special configuration.

Any recommendation on how to troubleshoot what might be happening due to the leak commits in 1.8.0? Thanks!

### LINKED ISSUE #740: ky hangs indefinitely with mode: 'no-cors' and a request body in Next.js
Hi there! First off, thank you for creating and maintaining `ky` 😄.

I've recently encountered an issue when using `ky` within a Next.js Server Component environment. After a deep dive into the Fetch Standard, I believe I've pinpointed the root cause and wanted to open an issue to share my findings and discuss a potential solution. I'm open to the possibility that I may have misunderstood something, so I welcome any corrections.

## The Problem

When making a request with a non-empty `body` (e.g., using the `json` option) and `mode: 'no-cors'`, the promise returned by `ky` never resolves or rejects. In the context of a Next.js application, this causes the Server Component responsible for the data fetch to hang indefinitely, never rendering a result.

As far as I can tell, this affects modern server-side JavaScript runtimes that implement the WHATWG Fetch standard.

## Live Reproduction

I've created a minimal reproduction in a Next.js project on StackBlitz that clearly demonstrates the "hang" using React `Suspense`.

**➡️ [Live Reproduction on StackBlitz](https://stackblitz.com/edit/stackblitz-starters-xanyaptv?file=app%2Fpage.tsx)**

In the example, the component using `mode: 'no-cors'` with a body never finishes rendering and remains in its "Loading..." state, while the other components render correctly.

## Fact-based Analysis from the Fetch Standard

The root cause appears to be a direct violation of the WHATWG Fetch Standard. Specifically, in the algorithm for the `new Request(input, init)` constructor ([Section 5.4 in the standard](https://fetch.spec.whatwg.org/#request-class)), step 39 dictates the following:

> If inputOrInitBody is non-null and inputOrInitBody’s source is null, then:
> 
> 1. If `initBody` is non-null and `init["duplex"]` does not exist, then throw a `TypeError`.
> 2. **If `this`’s request’s `mode` is neither "same-origin" nor "cors", then throw a `TypeError`.**
> 3. Set `this`’s request’s `use-CORS-preflight flag`.

Here's how this applies to `ky`:

1. When `ky` is used with a `json` or `string` body, it creates a `body` whose `source` is `null` (as it's not from a `Blob` or `FormData`).
2. The `mode` is set to `'no-cors'`.
3. According to step 39.2 above, because the `mode` is neither `"same-origin"` nor `"cors"`, the underlying `fetch` implementation **must** throw a synchronous `TypeError`.

The "hang" is a symptom of this synchronous `TypeError` being unhandled within `ky`'s promise chain. The error is effectively swallowed, and the promise never settles.

While `mode: 'no-cors'` is an unusual choice for server-to-server requests, I discovered this because a third-party library ([notion-client](https://github.com/NotionX/react-notion-x/blob/master/packages/notion-client/src/notion-api.ts#L762)) defaults to this option, causing our Next.js Server Components to fail silently.

## Thoughts on a Potential Solution

A robust, "fail-fast" solution would be to prevent this invalid request from ever being attempted. The most logical place for this validation seems to be within the `constructor`, immediately after the internal `Request` object is created.

This placement ensures that all options (including `json` being converted to `body`) have been processed, and `this.request.body` accurately reflects the presence of a body. It's consistent with how `ky` already handles other invalid initializations.

Here’s a rough sketch of what that might look like within the `constructor`:

```ts
// In the constructor, after this.request is created...

this.request = new globalThis.Request(this._input, this._options);

// A fail-fast guard could be added here to check for the spec violation.
if (this.request.mode === 'no-cors' && this.request.body) {
	throw new TypeError(
		'Request with `mode: "no-cors"` cannot have a body. This is a violation of the Fetch specification.'
	);
}

// ... rest of the constructor

```

This approach seems to be the most direct and architecturally consistent way to address the issue, but I'd be very interested to hear the team's thoughts.

Thank you for your time and for this great library!

whddltjdwhd:
## Related Issues
- Fixes #740 
- Fixes #690 

## The Problem
In the Node.js environment, a hang occurs in Ky's internal cleanup logic when a request using a ReadableStream as its body fails before being consumed by fetch.

This is caused by a long-standing bug in Node.js itself, which still exists in recent versions like v24.7.0. The bug is triggered when `await request.body.cancel()` is called on a request that has been cloned (for retries) but where neither the original nor the cloned stream has been consumed.

As discussed in [issue](https://github.com/sindresorhus/ky/issues/690#issuecomment-2815733245), this bug can be reproduced with the following minimal code:
```typescript
(async function () {
    let r = new Request('https://example.org', {
        method: 'POST',
        body: 'body',
    })
    
    let clone = r.clone()

    if (!r.bodyUsed) {
        console.log('not used')
        await r.body?.cancel()
        // The promise never completes if the request was cloned and the original body was never consumed
        console.log('done')
    }
})()
```
#### Node.js(v24.7.0) Result:

스크린샷 2025-09-21 오후 8 47 33

#### Deno(v2.4.5) Result:

스크린샷 2025-09-21 오후 8 48 02

This clearly demonstrates that the `.cancel()` promise remains perpetually pending in Node.js, causing the await expression to hang indefinitely.

## The Solution
To resolve this, the cleanup logic has been changed to call `ky.request.arrayBuffer()` instead of the problematic `ky.request.body.cancel()`.
```ts
.finally(async () => {
    // Now that we know a retry is not needed, close the ReadableStream of the cloned request.
    if (!ky.request.bodyUsed) {
       await ky.request.arrayBuffer();
    }
}) as ResponsePromise;
``` 
#### This change offers the following benefits:
1. Fixes the Hang: It completely avoids the buggy `.cancel()` method in Node.js, fundamentally resolving the hang issue.
2. Prevents Memory Leaks: By consuming the stream to completion, it ensures that all associated resources are reliably released, which is a more robust way of achieving the cleanup logic's original goal.

## Added Tests
Tests for two key scenarios have been added to validate this fix.

### Hang Prevention Test
This test simulates a scenario where the fetch function itself throws an error, causing the request to fail before the stream is consumed. If the test passes without a timeout, it proves that the cleanup logic no longer hangs.

```typescript
test('cleanup logic does not hang on stream request failure', async t => {
	const stream = Readable.from('Bell is Ringing.');
	const error = new TypeError('Simulated fetch error');

	await t.throwsAsync(
		ky.post('https://example.com', {
			body: stream,
			async fetch() {
				throw error;
			},
		}),
		{
			is: error,
		},
	);
});
```
### Memory Leak Prevention Test
This test verifies that the stream object is properly garbage-collected after a request fails. It uses a LeakDetector to check if any references to the stream object remain, which would indicate a memory leak.

```ts
test('failed stream request must not cause memory leak', async t => {
	async function isStreamLeaking() {
		let stream: Readable | undefined = Readable.from('Bell is Ringing.');
		const error = new TypeError('Simulated fetch error');
		const detector = new LeakDetector(stream);

		await t.throwsAsync(
			ky.post('https://example.com', {
				body: stream,
				async fetch() {
					throw error;
				},
			}),
			{
				is: error,
			},
		);

		stream = undefined;

		return detector.isLeaking();
	}

	t.false(await isStreamLeaking());
});
```


source/core/Ky.ts
```diff
@@ -88,7 +88,8 @@ export class Ky {
 			.finally(async () => {
 				// Now that we know a retry is not needed, close the ReadableStream of the cloned request.
 				if (!ky.request.bodyUsed) {
-					await ky.request.body?.cancel();
+					// Consume stream via .arrayBuffer() to avoid Node.js .cancel() hang bug on cloned streams.
```
sindresorhus, line 91: > This is caused by a long-standing bug in Node.js itself, which still exists in recent versions like v24.7.0. The bug is triggered when await request.body.cancel() is called on a request that has been cloned (for retries) but where neither the original nor the cloned stream has been consumed.

Please include a link to the open Node.js issue here. If there is none, please post one.

sindresorhus: This should also only be done for Node.js. I don't want to risk some weird behavior in browsers where it's not needed.

whddltjdwhd:
Hi @sindresorhus, thank you for your review. Your feedback prompted me to dig deeper, and I think I've found the root cause.

It appears the hanging behavior isn't a Node.js bug but is actually related to how streams are handled according to the WHATWG specification. The key detail seems to be that [`Request.clone()` uses `ReadableStream.tee()` internally](https://github.com/nodejs/undici/blob/main/lib/web/fetch/body.js#L296), which splits the stream into two branches. The spec mentions that for a cancelation to be successful, **both** branches must be canceled:

> To cancel the stream, cancel both of the resulting branches; a composite cancellation reason will then be propagated to the stream’s underlying source.
>
> — [WHATWG Streams Standard - stream.tee()](https://streams.spec.whatwg.org/#rs-prototype:~:text=Tees%20this%20readable,the%20two%20branches.)

The previous logic in the `.finally()` block only attempted to cancel one of the branches:

```ts
// Previous implementation
if (!ky.request.bodyUsed) {
    await ky.request.body?.cancel();
}
```

This seems to be why it was hanging, as it left the other stream branch pending. 

I confirmed that by canceling both streams, the minimal reproduction example now completes successfully in Node.js:
```js
// This test case now succeeds
(async function () {
    let r = new Request('https://example.org', { method: 'POST', body: 'body' });
    let clone = r.clone();

    if (!r.bodyUsed) {
        console.log('Neither body used. Canceling both streams...');
        await Promise.all([ r.body?.cancel(), clone.body?.cancel() ]);
        console.log('Done. Both streams canceled successfully.');
    }
})();
```
스크린샷 2025-09-22 오후 10 34 20

Based on this, I've updated the PR with a new approach to ensure the cleanup logic cancels both the original and the cloned request bodies:

```ts
	.finally(async () => {
		const originalRequest = ky._originalRequest;

		// Cancel both the original and cloned request bodies to prevent hanging.
		if (originalRequest && !originalRequest.bodyUsed && !ky.request.bodyUsed) {
			await Promise.all([
				originalRequest.body?.cancel(), ky.request.body?.cancel(),
			]);
		}
	}) as ResponsePromise;
```

My thinking is that this aligns directly with the spec and should be a more robust and portable solution than the `.arrayBuffer()` workaround.

I'd appreciate hearing your thoughts on this direction.

sindresorhus (approved):

source/core/Ky.ts
```diff
@@ -86,9 +86,13 @@ export class Ky {
 		const isRetriableMethod = ky._options.retry.methods.includes(ky.request.method.toLowerCase());
 		const result = (isRetriableMethod ? ky._retry(function_) : function_())
 			.finally(async () => {
-				// Now that we know a retry is not needed, close the ReadableStream of the cloned request.
-				if (!ky.request.bodyUsed) {
-					await ky.request.body?.cancel();
+				const originalRequest = ky._originalRequest;
+
+				// Cancel both the original and cloned request bodies to prevent hanging.
+				if (originalRequest && !originalRequest.bodyUsed && !ky.request.bodyUsed) {
```
sholladay, line 92: Is there a situation where one request body _does_ get used but we still need to cancel the other one? Maybe we shouldn't lump these together?

In fact, the reason why the code here originally only cancelled the clone (`ky.request`) is because our expectation was that the original (`ky._originalRequest`) _would_ be consumed by fetch during the normal course of trying to send the request. So if I'm not mistaken, there will often be a situation where `originalRequest.bodyUsed` is `true` because the request was sent, but `ky.request.bodyUsed` is `false` because a retry was not performed. Don't we still need to call `ky.request.body.cancel()` in that case?

sholladay:
I replaced the simulated fetch error in `test/memory-leak.ts` with an actual fetch error, to get more realistic behavior from the test. Interestingly, the test now reports that it is leaking memory.

sholladay:
I replaced the simulated fetch error in `test/memory-leak.ts` with an actual fetch error, to get more realistic behavior from the test. Interestingly, the test now reports that it is leaking memory in Node v20 and v22, and locally for me on v24. Although, v24 passed in CI for some reason.

@whddltjdwhd any thoughts on what's going on there and what we should do about it?

sholladay:
Turns out the test failures were caused by bugs in `jest-leak-detector`. Fixed by upgrading to latest.

whddltjdwhd:
Wow, thank you for going above and beyond to resolve the test failure. I really appreciate you improving the test and tracking down the root cause in `jest-leak-detector`.

I've reviewed all the latest commits you've added, and everything looks great. Thanks for all your help😄!!

sindresorhus:
Excellent. Thanks for fixing 🙏

Issue #742, opened 2025-09-24
feat: type guard for ky errors


clicktodev:
Axios has a [builtin type guard for errors](https://github.com/axios/axios?tab=readme-ov-file#typescript) that makes it easy to figure out whether an error is generated by axios or not.

i currently have this ts error when i try to check if an error is produced by ky

<img width="1792" height="538" alt="Image" src="https://github.com/user-attachments/assets/95c6a89a-f3a7-43b9-9133-9b3b488cd257" />

I'd like the same for ky

```ts
let user: User = null;
try {
  const { data } = await ky.get('/user?ID=12345');
  user = data.userDetails;
} catch (error) {
  if (ky.isKyError(error)) {
    handleKyError(error);
  } else {
    handleUnexpectedError(error);
  }
}
```

sholladay:
I'm in favor of this if we can make it intuitive and keep it relatively simple. Would it treat simple option validation errors as true or false? What about fetch errors?

I'm thinking it should return true just for error classes defined by Ky.

clicktodev:
for now i think just the error classes by ky is enough

clicktodev:
probably `ky.isHttpError` and `ky.isTimeoutError` since these are the only ky defined errors as far as i know

sholladay:
There are going to be more in the future. At a minimum, we plan to introduce `ky.NetworkError` at some point.

PR #743, opened 2025-09-24
Improve error handling example in readme

clicktodev:
Updated error handling example in readme.md to use instanceof for type checking.

this helps with #742

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -621,15 +621,15 @@ Be aware that some types of errors, such as network errors, inherently mean that
 
 If you need to read the actual response when an `HTTPError` has occurred, call the respective parser method on the response object. For example:
 
-```ts
+```js
 import { HTTPError } from "ky";
 
 try {
 	await ky('https://example.com').json();
 } catch (error) {
-  if (error instanceof HTTPError) {
-    const errorJson = await error.response.json();
-  }
+	if (error instanceof HTTPError) {
+		const errorJson = await error.response.json();
+	}
 }
 ```
 


sholladay:
I don't see the point of this. You're changing code that already works across realms. The only thing this solves is if we continue exporting `HTTPError` but changed its `name` despite not changing the export, which is highly unlikely.

clicktodev:
This snippet is simply typescript friendly now so whoever copies it can get intellisense.
Also this is perfect if we combine it with my other PR that makes the error names constants.
So both instanceof and .name checks are typesafe.
It seems highly unlikely to you but why risk it?
This just makes everything more reliable and if you do change the error names or introduce new errors then Im they will be instantly typed.

sholladay:
I'm fine with changing it for the sake of TypeScript. What makes it TS friendly is using `instanceof`. So let's just do that. Anyone who needs cross-realm safety almost certainly knows about its limitations.

Examples need to be simple, not handle every edge case. And TS is, by far, a more common use case.

LinusU:
Can cross realms really be a problem here? If they are imported as `import { ky, HTTPError } from 'ky'`, then calling that `ky` will always yield the same `HTTPError` class?

clicktodev:
cool i removed the name check and kept the instanceof check

sholladay:
@LinusU yes, in principle an error instance could come from a web worker or something and the check would fail (return false incorrectly). But that's a niche situation.

clicktodev:
this test workflow has been stuck for 3 hours https://github.com/sindresorhus/ky/actions/runs/18005471093/job/51224585643

PR #744, opened 2025-09-24
feat: add ERROR_NAMES constants and export them from library

clicktodev:
- Add ERROR_NAMES constant object in core/constants.ts
- Update HTTPError and TimeoutError classes to use constants
- Export ERROR_NAMES from main index.ts for external usage
- Improves maintainability and provides better DX for error handling

this can be helpful when checking error names like in #743

```ts
import ky, { HTTPError, ERROR_NAMES } from 'ky';

try {
    await ky('https://api.example.com/data');
} catch (error) {
    if (error.name === ERROR_NAMES.HTTPError) {
        // Handle HTTP error
    } else if (error.name === ERROR_NAMES.TimeoutError) {
        // Handle timeout error
    }
}
```

### DIFF
--- a/source/errors/HTTPError.ts
+++ b/source/errors/HTTPError.ts
@@ -10,7 +10,7 @@ export class HTTPError<T = unknown> extends Error {
 
 	constructor(response: Response, request: Request, options: NormalizedOptions) {
 		const code = (response.status || response.status === 0) ? response.status : '';
-		const title = response.statusText || '';
+		const title = response.statusText ?? '';
 		const status = `${code} ${title}`.trim();
 		const reason = status ? `status code ${status}` : 'an unknown error';
 


sholladay:
Why don't you just use `instanceof'?

clicktodev:
Instanceof doesn’t always work across realms so since checking the name is the next reliable way to check for the error type then we should at least make it type safe and easier to check against in both js and ts

sindresorhus:
Why not just this?

```js
import ky, { HTTPError, TimeoutError } from 'ky';

try {
    await ky('https://api.example.com/data');
} catch (error) {
    if (error.name === HTTPError.name) {
        // Handle HTTP error
    } else if (error.name === TimeoutError.name) {
        // Handle timeout error
    }
}
```

sholladay:
The approach that Sindre showed there is actually what I have in a couple of my own apps. It works well and is elegant, IMO.

I do think we should implement the type guard you mentioned in #742. It would be useful even for non-TypeScript users.

clicktodev:
I suppose `error.name === HTTPError.name` is good enough.
i still think it's a good idea to use the constants atleast internally `this.name = ERROR_NAMES.HTTPError;` but if you don't see the point in this PR then feel free to close it.

PR #746, opened 2025-09-25
Add type guards for errors

### LINKED ISSUE #742: feat: type guard for ky errors

clicktodev:
Introduce type guards to enhance error handling in the Ky client, allowing for more precise identification of HTTP and timeout errors. This improves the robustness of error management in the application.

closes #742

### DIFF
--- a/test/hooks.ts
+++ b/test/hooks.ts
@@ -1,6 +1,6 @@
 import test from 'ava';
 import delay from 'delay';
-import ky, {type HTTPError, isHTTPError} from '../source/index.js';
+import ky, {HTTPError, isHTTPError} from '../source/index.js';
 import {type Options} from '../source/types/options.js';
 import {createHttpTestServer} from './helpers/create-http-test-server.js';
 
@@ -404,6 +404,7 @@ test('beforeRetry hook is called with error and retryCount', async t => {
 		hooks: {
 			beforeRetry: [
 				({error, retryCount}) => {
+					t.true(error instanceof HTTPError);
 					t.true(isHTTPError(error));
 					t.true(retryCount >= 1);
 				},
@@ -481,6 +482,7 @@ test('beforeRetry hook with parseJson and error.response.json()', async t => {
 			hooks: {
 				beforeRetry: [
 					async ({error, retryCount}) => {
+						t.true(error instanceof HTTPError);
 						t.true(isHTTPError(error));
 						t.is(error.message, `Request failed with status code 502 Bad Gateway: GET ${server.url}/`);
 						t.true((error as HTTPError).response instanceof Response);
--- a/test/hooks.ts
+++ b/test/hooks.ts
@@ -1,6 +1,6 @@
 import test from 'ava';
 import delay from 'delay';
-import ky, {HTTPError} from '../source/index.js';
+import ky, {HTTPError, isHTTPError} from '../source/index.js';
 import {type Options} from '../source/types/options.js';
 import {createHttpTestServer} from './helpers/create-http-test-server.js';
 
@@ -404,7 +404,7 @@ test('beforeRetry hook is called with error and retryCount', async t => {
 		hooks: {
 			beforeRetry: [
 				({error, retryCount}) => {
-					t.true(error instanceof HTTPError);
+					t.true(isHTTPError(error));
 					t.true(retryCount >= 1);
 				},
 			],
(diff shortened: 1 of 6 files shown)


clicktodev:
I'm thinking about changing the check to `isTimeoutError: (error: unknown) => error is TimeoutError || error.name === HTTPError.name;` to handle cross realm situations

clicktodev:
last commit adds a fallback name check in case instanceof fails for some reason. let me know what you think

clicktodev:
also idk if you want the helpers to be attached to the ky instance or keep them standalone

sholladay:
IMO, these functions should be standalone and not attached to the Ky instance.

I also think that, unlike `isKyError()`, the more specific functions provide limited value and probably shouldn't be included.

clicktodev:
i think the specific type guards are beneficial for both internal and external use.

```ts
	if (isKyError(error)) {
		if (isHTTPError(error)) {
			console.log('Status:', error.response.status);
		} else if (isTimeoutError(error)) {
			console.log('Timeout URL:', error.request.url);
		}
	} 
```

I also replaced the internal instanceof checks with the new helper functions.
I'm thinking of passing a boolean parameter to enable the name check which defaults to false, we can call it `crossRealmCheck` or `nameCheck`

```ts
export function isHTTPError<T = unknown>(error: unknown, crossRealmCheck = false): error is HTTPError<T> {
    return error instanceof HTTPError || (crossRealmCheck && ((error as any)?.name === HTTPError.name));
}
```

clicktodev:
i removed the name check for now but i can bring it back if you think it's useful

sholladay:
They should support `name` as a fallback. Otherwise they don't earn their keep. Needs to be worth the weight of an import.

clicktodev:
My only issue is that name checking can produce false positives since other errors can potentially have the same name.
Should we still go through with the name check?

clicktodev:
brought back the name checks

test/hooks.ts
```diff
@@ -481,7 +481,7 @@ test('beforeRetry hook with parseJson and error.response.json()', async t => {
 			hooks: {
 				beforeRetry: [
 					async ({error, retryCount}) => {
-						t.true(error instanceof HTTPError);
+						t.true(isHTTPError(error));
```
sholladay, line 484: You can add this as an additional assertion, but I would prefer to keep the original.

test/hooks.ts
```diff
@@ -404,7 +404,7 @@ test('beforeRetry hook is called with error and retryCount', async t => {
 		hooks: {
 			beforeRetry: [
 				({error, retryCount}) => {
-					t.true(error instanceof HTTPError);
+					t.true(isHTTPError(error));
```
sholladay, line 407: Same here. Should be an additional assertion.

sholladay (changes requested):

sholladay:
Just one more small nit. Otherwise LGTM.

sholladay:
The readme examples are now broken...

sholladay:
The imports can be combined into a single line:

```js
import ky, {TimeoutError} from 'ky';
```

sholladay (approved):

clicktodev:
it seems node 24 introduces Error.isError() (still not fully supported in safari) which works cross realm which means we can get rid of the name check and the instanceof check.

sholladay:
I don't see how that would be useful here. `Error.isError()` does not support comparisons to check if an object is one of  _our_ errors.

clicktodev:
You're right but we can however use it to get rid of the `as any` type assertion.

`(Error.isError(error) && error.name === HttpError.name);`

instead of

`((error as any)?.name === HTTPError.name);`

but this requires bumping to node 24 but i don't think it's worth it until node 24 becomes the lowest maintenance release.

also until safari fixes its implementation because it currently returns false on DomExceptions : (

PR #767, opened 2025-10-17
Add `retryOnTimeout` and `shouldRetry` options

### LINKED ISSUE #546: Retries on timeout
Section https://github.com/sindresorhus/ky#retry states that
> Retries are not triggered following a timeout.

What if I want a retry on timeout? How is it possible to achieve?

sindresorhus:
Fixes #546

While `retryOnTimeout` could be achieved with `shouldRetry`. I think it's nice to keep `retryOnTimeout` as a convenience.

It's not a hook becuase `shouldRetry` is a decision (when to retry), not a transformation (what to change). Hooks transform things. Options configure behavior.

### DIFF
--- a/test/helpers/with-performance.ts
+++ b/test/helpers/with-performance.ts
@@ -9,9 +9,9 @@ type Argument = {
 };
 
 // We allow the tests to take more time on CI than locally, to reduce flakiness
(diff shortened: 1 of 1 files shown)


sindresorhus:
Tests passing now. Ready for review.

PR #771, opened 2025-10-18
Add X-RateLimit variant of Retry-After header

### LINKED ISSUE #768: Extend retry options with custom retry headers
At the moment retry code looks for few hard-coded headers:

```js
error.response.headers.get('Retry-After')
				?? error.response.headers.get('RateLimit-Reset')
				?? error.response.headers.get('X-RateLimit-Reset') // GitHub
				?? error.response.headers.get('X-Rate-Limit-Reset'); // Twitter
```

Would it be possible to have this configurable as retry option to define, e.g. let's a list of headers to look into?

These headers started evolved as non-standard for quite a bit having various legacy variants and on top of that there are already few IEFT drafts, so it would be useful to be able to specify, which headers to look into.

E.g. just dealing with `X-RateLimit-Retry-After`, I know .... non-standard, but it's a lot if this with rate-limiting out there , hence to extend config.

I could have a look into putting together a PR if it sounds like desirable feature to have?

lukashroch:
fixes #768 

added missing `x-ratelimit-` variant of `retry-after` header directly as discussed in #768 Current implementation already includes `x-ratelimit-` variant for `reset` one.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -321,6 +321,7 @@ export class Ky {
 
 			const retryAfter = error.response.headers.get('Retry-After')
 				?? error.response.headers.get('RateLimit-Reset')
+				?? error.response.headers.get('X-RateLimit-Retry-After') // Symfony-based services
 				?? error.response.headers.get('X-RateLimit-Reset') // GitHub
 				?? error.response.headers.get('X-Rate-Limit-Reset'); // Twitter
 			if (retryAfter && this.#options.retry.afterStatusCodes.includes(error.response.status)) {


sholladay:
Could you try to find a popular service that uses this and put it in a comment to the right, like the other headers in this list? Might help pressure these companies to clean up the mess they are making.

sindresorhus:
I guess one of these: https://www.google.com/search?client=safari&rls=en&q=%22X-RateLimit-Retry-After%22&ie=UTF-8&oe=UTF-8

lukashroch:
Looks like it might be coming mostly from services based on symfony framework, shall we put that in comment rather than random specific service?

lukashroch:
I've added the comment now and rebased on top, please let me know if any more changes needed

sholladay (approved):

PR #786, opened 2025-11-06
Add `bearer` option

sindresorhus:


sindresorhus:
@sholladay Setting `bearer` is quite a common use-case, so I think it may be worth adding an option for it. Thoughts?

sholladay:
I'm pretty neutral on this, personally. Should be easy to maintain but it also doesn't provide a lot of value. And it has all the classic downsides of two sources of truth, e.g. when extending a Ky instance or in a hook that wants to delete the bearer token, they need to be aware of both `bearer` and `headers.authorization` and unset both, since either might've been used. Just a minor nuisance, but still a potential for bugs.

Is there something we could do to enhance it and make it more worthwhile? Validate the token format? Validate that an insecure protocol like `http` is not being used? Just something beyond saving a handful of characters.

raedalmalki5-prog (approved):

sholladay (approved):
LGTM, though it still seems of limited value.

I think `token` might be a nicer name for this option, but `bearer` is fine, too.

sindresorhus:
Going to pass on this. I agree it's not worth it and two sources of truth is not great.

PR #789, opened 2025-11-15
Changes type of retry.methods option to HttpMethod

### LINKED ISSUE #807: retry.methods is not typed
This hook's retry
```
	retry: {
		limit: 2,
		statusCodes: [401],
		methods: ["get", "post", "put", "delete", "patch"],
	},
```

gets triggered correctly, however, 
```
	retry: {
		limit: 2,
		statusCodes: [401],
		methods: ["GET", "POST", "PUT", "DELETE", "PATCH"],
	},
```

Does not work, also, is there any specific reason for retry.methods to be string[] and not Method[] ?

vokrik:
I learned the hard way that the methods need to be lowercase (and kept wondering why the requests are not retried). This change should prevent that mistake

---

Fixes #807

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -448,7 +448,7 @@ const response = await ky('https://example.com/api', {
 Type: `Function[]`\
 Default: `[]`
 
-This hook enables you to modify the `HTTPError` right before it is thrown. The hook function receives a `HTTPError` and a state object as arguments and should return an instance of `HTTPError`.
+This hook enables you to modify the `HTTPError` right before it is thrown. The hook function receives an `HTTPError` and a state object as arguments and should return an instance of `HTTPError`.
 
 The `state.retryCount` is `0` for the initial request and increments with each retry. This allows you to distinguish between the initial request and retries, which is useful when you need different error handling based on retry attempts (e.g., showing different error messages on the final attempt).
 
--- a/source/types/hooks.ts
+++ b/source/types/hooks.ts
@@ -62,7 +62,7 @@ export type Hooks = {
 
 	The `state.retryCount` is `0` for the initial request and increments with each retry. This allows you to distinguish between initial requests and retries, which is useful when you need different behavior for retries (e.g., avoiding overwriting headers set in `beforeRetry`).
 
-	A [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) can be returned from this hook to completely avoid making a HTTP request. This can be used to mock a request, check an internal cache, etc. An **important** consideration when returning a `Response` from this hook is that all the following hooks will be skipped, so **ensure you only return a `Response` from the last hook**.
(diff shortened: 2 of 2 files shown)


source/types/common.ts
```diff
@@ -6,3 +6,5 @@ export type Required<T, K extends keyof T = keyof T> = T & {[P in K]-?: T[P]};
 export type LiteralUnion<LiteralType extends BaseType, BaseType extends Primitive> =
 	| LiteralType
 	| (BaseType & {_?: never});
+
+export type HttpMethod = 'get' | 'post' | 'put' | 'patch' | 'head' | 'delete';
```
vokrik: Had to move this to common so that there is not a circular dependency between `retry.ts` and `options.ts`

source/types/common.ts
```diff
@@ -6,3 +6,6 @@ export type Required<T, K extends keyof T = keyof T> = T & {[P in K]-?: T[P]};
 export type LiteralUnion<LiteralType extends BaseType, BaseType extends Primitive> =
 	| LiteralType
 	| (BaseType & {_?: never});
+
+export type RequestHttpMethod = 'get' | 'post' | 'put' | 'patch' | 'head' | 'delete' ;
```
vokrik: For some reason, in `constants.ts` there is an exported constant containing methods without 'options' and 'trace'

```typescript
// source/core/constants.ts
...
export const requestMethods = ['get', 'post', 'put', 'patch', 'head', 'delete'] as const;

```

I guess we want to keep split the methods that the user does proactively (rather than the browser doing automatically) and all available methods. 

The default retryMethods however contain also options and trace as a default:

```typescript
// source/utils/normalize.ts
const retryMethods: Array<HttpMethod> = ['get', 'put', 'head', 'delete', 'options', 'trace'];
```

sindresorhus:
I appreciate you working to prevent the uppercase method mistake. However, hardcoding `HttpMethod` as a union type would break extensibility. There are almost 40 registered HTTP methods in the IANA registry (not just 8). The Fetch API also allows arbitrary custom methods, and users might want to retry them.

Instead, we should use the `LiteralUnion` pattern that Ky already uses elsewhere:

```typescript
methods?: Array<LiteralUnion<HttpMethod, string>>;
```

This provides autocomplete for common methods while still accepting any string, maintaining both type safety and extensibility.

vokrik:
> I appreciate you working to prevent the uppercase method mistake. However, hardcoding `HttpMethod` as a union type would break extensibility. There are almost 40 registered HTTP methods in the IANA registry (not just 8). The Fetch API also allows arbitrary custom methods, and users might want to retry them.
> 
> Instead, we should use the `LiteralUnion` pattern that Ky already uses elsewhere:
> 
> ```ts
> methods?: Array<LiteralUnion<HttpMethod, string>>;
> ```
> 
> This provides autocomplete for common methods while still accepting any string, maintaining both type safety and extensibility.

Ok, that makes sense.. In that case, would it make sense to you if I lowercased the passed methods in `normalize.ts` so that it's bulletproof? (Added proposal as a last commit)

PR #792, opened 2025-11-22
Correct grammar related to a/an

antareepsarkar:
Just a small PR to correct grammar related to a/an


antareepsarkar:
> Needs to be fixed in the readme, too.

Just did that. Thanks for pointing out.

PR #793, opened 2025-11-24
Improve type safety in test file

novice0840:
I add type assertion to resolve conflict between Node.js stream/web ReadableStream and DOM ReadableStream types. The test runs in Node.js but ky uses DOM types for cross-platform compatibility.

I attached a screenshot of type error

image

### DIFF
--- a/test/memory-leak.ts
+++ b/test/memory-leak.ts
@@ -36,9 +36,7 @@ test('shared abort signal must not cause memory leak of input', async t => {
 
 test('failed stream request must not cause memory leak', async t => {
 	async function isStreamLeaking() {
-		const stream: ReadableStream | undefined = ReadableStream.from([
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't really see the point of this. The tests run in Node.js. There is no type error when running the tests. So really no need to change anything. Unless I'm missing something?

sholladay:
Isn't this _less_ type safe because the type is asserted rather than inferred?

I'm inclined to agree with Sindre.

novice0840:
Oh, I see. I understood what you want to say 

I added the type assertion because the current code is showing a type error in the IDE, which I think could be confusing for anyone looking at this code for the first time.
You're right that functionally nothing has changed. But I felt that having those red squiggly lines could be distracting during development.

I'll leave the final call up to you. If my reasoning doesn't convince you, no problem closing the PR.

PR #804, opened 2025-12-18
feat: Add beforeTimeout hook

### LINKED ISSUE #508: add hook for handling `TimeoutError`
As of now, the `beforeError` hook only handles `HTTPErrors`.  A hook for handling errors like `TimeoutError` would be beneficial to avoid having to wrap every request call in `try { ... } catch { ... }` blocks

constantly-dev:
Closes #508

### Description

This PR introduces a new hook, `beforeTimeout`, which allows users to intercept and handle `TimeoutError`s globally.

Currently, the `beforeError` hook only receives `HTTPError`, forcing users to wrap every request in a `try-catch` block to handle timeouts (e.g., for logging or UI notifications). This PR solves that inconvenience.

### Implementation Details

Based on the discussion in #508, I have implemented the following:

- **Naming**: Named the hook `beforeTimeout` as suggested by @sindresorhus.
- **Behavior**: The hook receives the `TimeoutError`, allowing modification (e.g., changing the error message).
- **Control Flow**: Crucially, **the error is re-thrown** after the hooks are executed. It does not silence the error, preserving Ky's promise rejection behavior.
- **Types**: Added `BeforeTimeoutHook` type definition and updated `Hooks` interface.
- **Tests**: Added comprehensive tests to ensure the hook is triggered and can modify the error object.

### Example Usage

```typescript
import ky from 'ky';

await ky('[https://example.com](https://example.com)', {
	timeout: 100,
	hooks: {
		beforeTimeout: [
			error => {
				// Log the timeout globally
				console.error('Request timed out:', error);

				// Or modify the error before it propagates
				error.message = 'Custom timeout message';

				return error;
			}
		]
	}
});

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -1,6 +1,7 @@
 import {HTTPError} from '../errors/HTTPError.js';
 import {NonError} from '../errors/NonError.js';
 import {ForceRetryError} from '../errors/ForceRetryError.js';
+import {TimeoutError} from '../errors/TimeoutError.js';
 import type {
 	Input,
 	InternalOptions,
@@ -43,7 +44,23 @@ export class Ky {
 			await Promise.resolve();
 			// Before using ky.request, _fetch clones it and saves the clone for future retries to use.
 			// If retry is not needed, close the cloned request's ReadableStream for memory safety.
-			let response = await ky.#fetch();
+			let response: Response;
+			try {
+				response = await ky.#fetch();
+			} catch (error) {
+				if (error instanceof TimeoutError) {
+					let timeoutError = error;
+
+					for (const hook of ky.#options.hooks.beforeTimeout) {
+						// eslint-disable-next-line no-await-in-loop
+						timeoutError = await hook(timeoutError);
+					}
+
+					throw timeoutError;
+				}
+
+				throw error;
+			}
 
 			for (const hook of ky.#options.hooks.afterResponse) {
 				// Clone the response before passing to hook so we can cancel it if needed
@@ -198,6 +215,7 @@ export class Ky {
 					beforeRetry: [],
 					beforeError: [],
 					afterResponse: [],
+					beforeTimeout: [],
 				},
 				options.hooks,
 			),
--- a/source/types/hooks.ts
+++ b/source/types/hooks.ts
@@ -1,5 +1,7 @@
 import {type stop, type RetryMarker} from '../core/constants.js';
-import type {KyRequest, KyResponse, HTTPError} from '../index.js';
+import type {
+	KyRequest, KyResponse, HTTPError, TimeoutError,
+} from '../index.js';
 import type {NormalizedOptions} from './options.js';
 
 export type BeforeRequestState = {
@@ -55,6 +57,8 @@ export type BeforeErrorState = {
 };
 
 export type BeforeErrorHook = (error: HTTPError, state: BeforeErrorState) => HTTPError | Promise<HTTPError>;
+
+export type BeforeTimeoutHook = (error: TimeoutError) => TimeoutError | Promise<TimeoutError>;
 
 export type Hooks = {
 	/**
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: source/core/Ky.ts [hunk 1]
```diff
@@ -43,7 +44,23 @@ export class Ky {
...
+			try {
+				response = await ky.#fetch();
+			} catch (error) {
+				if (error instanceof TimeoutError) {
+					let timeoutError = error;
+
+					for (const hook of ky.#options.hooks.beforeTimeout) {
+						// eslint-disable-next-line no-await-in-loop
+						timeoutError = await hook(timeoutError);
+					}
```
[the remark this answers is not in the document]
constantly-dev: @samicey 
Thank you for your response!
Yes, you are correct. It works as a pipeline where the output of one hook becomes the input of the next.

I intentionally implemented it this way to maintain consistency with ky's existing `beforeError` hook logic.
If you check the `beforeError` implementation in `Ky.ts`, it uses the exact same pattern:

```ts
// Existing beforeError logic in Ky.ts
for (const hook of ky.#options.hooks.beforeError) {
    error = await hook(error, {retryCount: ky.#retryCount});
}
```

I believe applying the same pattern to `beforeTimeout` is the most consistent approach for this library. 

Do you think there is a better approach for this?

sholladay:
I think I would mildly prefer to send `TimeoutError` through `beforeError` rather than a new `beforeTimeout` hook (or do both). It enables simpler handling of  multiple error types.

constantly-dev:
> I think I would mildly prefer to send TimeoutError through beforeError rather than a new beforeTimeout hook (or do both). It enables simpler handling of multiple error types.

@sholladay 
In the original issue (#508), suggested naming it beforeTimeout explicitly. I am happy to merge this into beforeError if that is the preferred direction now. 

However, beforeError is currently typed to strictly receive HTTPError. If we merge them, we'll need to broaden the type definition (e.g., HTTPError | TimeoutError).

Which direction should I proceed with? I'll update the PR as soon as you guys decide.

source/types/hooks.ts
```diff
@@ -265,4 +269,33 @@ export type Hooks = {
...
+        hooks: {
+            beforeTimeout: [
+                async error => {
+                    // Log the timeout
+                    console.error('Request timed out', error);
+
+                    // Or modify the error message
+                    error.message = 'Custom timeout message';
+
+                    return error;
```
sindresorhus: use tab-indentation

source/types/hooks.ts
```diff
@@ -265,4 +269,33 @@ export type Hooks = {
...
+	/**
+    This hook enables you to modify the `TimeoutError` right before it is thrown. The hook function receives a `TimeoutError` as an argument and should return an instance of `TimeoutError`.
+
+    @default []
+
+    @example
+    ```
+    import ky from 'ky';
+
+    await ky('[https://example.com](https://example.com)', {
```
sindresorhus: typo

source/types/hooks.ts
```diff
@@ -56,6 +58,8 @@ export type BeforeErrorState = {
 
 export type BeforeErrorHook = (error: HTTPError, state: BeforeErrorState) => HTTPError | Promise<HTTPError>;
 
+export type BeforeTimeoutHook = (error: TimeoutError) => TimeoutError | Promise<TimeoutError>;
```
sindresorhus: Must be exported from index.d.ts too.

sindresorhus:
I think it should be in `beforeError`, but it's a breaking change, so cannot happen yet. You can do the PR, but we cannot merge until we start work on v2, which is not going to happen yet.

sholladay:
> In the original issue (#508), suggested naming it beforeTimeout explicitly. I am happy to merge this into beforeError if that is the preferred direction now.

@constantly-dev if you look closely, you'll notice that I edited the issue title to specify `beforeError` after the initial discussion where Sindre and I both seemed a bit skeptical of the need for a separate hook. But you're right, I didn't add a note to the top of the thread. I tend to avoid editing people's posts unless it's really necessary. Sorry for the confusion. 

Let's proceed with `beforeError`.

sholladay:
Closing since I believe #829 made this PR obsolete. Feel free to let me know if I missed anything.

PR #819, opened 2026-01-23
Docs: clarify consuming or canceling error.response body

### LINKED ISSUE #633: Document requirement to handle `error.response` when catching `HTTPError`
Reproduction with deno :

`ky.deno.test.ts` :

```ts
import ky from "https://esm.sh/ky@1.7.2";

Deno.test("throws HTTPError : no memory leak", async () => {
  // this test throws the expected error and does not leak memory
  // HTTPError: Request failed with status code 404 Not Found: GET https://github.com/not_found

  await ky.get("https://github.com/not_found").text();
});

Deno.test("memory leak when catching error", async () => {
  // this test report a memory leak when catching the error
  // the memory leak is fixed if we rethrow something

  // the test fails with :
  // Leaks detected:
  // - A fetch response body was created during the test, but not consumed during the test. Consume or close the response body `ReadableStream`, e.g `await resp.text()` or `await resp.body.cancel()`.

  try {
    await ky.get("https://github.com/not_found").text();
  } catch (error) {
    console.error("I handle the error here");

    // fix the leak : rethrow the error
    // throw error;
  }
});
```

`deno test --allow-net ky.deno.test.ts`

Same issue with bun : https://github.com/denoland/deno/issues/25546#issuecomment-2341318464

adiadia:
Fixes #633

Tests: npm test (261 passed, 2 known failures)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "ky",
-	"version": "1.14.2",
+	"version": "1.14.3",
 	"description": "Tiny and elegant HTTP client based on the Fetch API",
 	"license": "MIT",
 	"repository": "sindresorhus/ky",
--- a/source/utils/merge.ts
+++ b/source/utils/merge.ts
@@ -171,9 +171,5 @@ export const deepMerge = <T>(...sources: Array<Partial<T> | undefined>): T => {
 		}
 	}
 
-	if (returnValue.context === undefined) {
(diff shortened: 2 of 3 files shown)


sindresorhus:
No longer needed: https://github.com/sindresorhus/ky/commit/1341f5ca56c96c9440e6e8d0d624c871bb1b7929

PR #821, opened 2026-02-03
feat: add gRPC serialization support.

Rik-System-Root:


Rik-System-Root:
"@sindresorhus Soporte gRPC completado y enviado en este PR #821. He verificado que los cambios se pueden fusionar limpiamente sin conflictos. Quedo atento a tu revisión para la liberación de la recompensa acordada."

sindresorhus:
Thanks for the PR. We are not adding gRPC support to ky. Ky is an HTTP(S) fetch wrapper, and gRPC requires HTTP/2 framing and a dedicated transport, not just headers and serialization, so it is out of scope. This change also bypasses validateAndMerge and defaults, introduces an undocumented protobuf option, and removes existing behavior. Please maintain this as a separate package or use a custom hook in your app. Closing.

PR #823, opened 2026-02-16
Add `data` property to `HTTPError` with pre-parsed response body

### LINKED ISSUE #633: Document requirement to handle `error.response` when catching `HTTPError`

### LINKED ISSUE #642: Improve Error Handling for Better Integration with React Query
# Improve Error Handling for Better Integration with React Query

## Current Behavior

Currently, when an error occurs in a Ky request, the error object contains a promise that needs to be resolved to access the error details. This differs from Axios, where error details are immediately available in the `error.response.data` property.

This difference creates challenges when using Ky with libraries like React Query, especially when trying to render error messages immediately after a failed mutation.

## Desired Behavior

It would be beneficial if Ky could provide immediate access to parsed error information, similar to Axios. This would allow for more straightforward error handling and rendering in React applications, particularly when used with React Query.

## Example of Current Issue

Consider this example using React Query with Axios:

```jsx
import { useMutation } from '@tanstack/react-query';
import axios from 'axios';

function MyComponent() {
  const { mutate, isError, error, reset } = useMutation({
    mutationFn: (data) => axios.post('https://api.example.com/data', data),
  });

  return (
    <div>
      <button onClick={() => mutate({ some: 'data' })}>Submit</button>
      {isError && (
        <Alert severity="error" onClose={reset}>
          {error.response?.data.message}
        </Alert>
      )}
    </div>
  );
}
```

In this Axios example, `error.response.data.message` is immediately available for rendering.

Now, let's look at the same example using Ky:

```jsx
import { useMutation } from '@tanstack/react-query';
import ky from 'ky';
import { useState } from 'react';

function MyComponent() {
  const [error, setError] = useState(null);

  const { mutate, isError, reset } = useMutation({
    mutationFn: (data) => ky.post('https://api.example.com/data', { json: data }).json(),
    onError: async (error, variables, context) => {
      if (error.response) {
        // Await the json response from the error
        const jsonError = await error.response.json();
        setError(jsonError); // Save the error to state
      } else {
        setError({message: 'An unexpected error occurred'});
      }
    }
  });

  return (
    <div>
      <button onClick={() => { mutate({ some: 'data' }); reset(); }}>Submit</button>
      {isError && error && (
        <Alert severity="error" onClose={() => { reset(); setError(null); }}>
          {error.message}
        </Alert>
      )}
    </div>
  );
}
```

In this Ky example, `error.response.json()` is a promise, so `error.response?.data.message` is not immediately available. We would need to await the promise before we can access the error details and create new state with errors, which complicates the rendering process.

## Additional Complexity

The current Ky implementation adds further complexity because the promise represents not just the response, but also the request. This means that in some scenarios, we might need to call `.json()` twice: once for sending the request and once for parsing the `error.request.json()`

## Proposed Solution

One possible solution could be to automatically parse the error response and make it immediately available, similar to Axios. For example:

```javascript
ky.post('https://api.example.com/data', { json: data })
  .json()
  .catch(error => {
    if (error instanceof HTTPError) {
      const responseData = await error.response.json();
      throw {
        ...error,
        response: {
          data: responseData,
          status: error.response.status,
          statusText: error.response.statusText,
        },
      };
    }
    throw error;
  });
```

This would allow Ky to behave more like Axios in error scenarios, making it easier to use with React Query and similar libraries. But when developing this option, it requires creating a new error class to type react-query correctly and writing additional code.

## Questions for Consideration

1. Is it feasible to change Ky's error handling to more closely match Axios's behavior?
2. Could Ky provide an option to automatically parse error responses and make them immediately available?

Thank you for considering this enhancement. It would greatly improve the developer experience when using Ky, especially in React applications with React Query.

sindresorhus:
The error response body is now automatically parsed and available as `error.data`. JSON responses are parsed using the `parseJson` option or `JSON.parse`. Other content types are set as plain text. Empty bodies or parse failures result in `undefined`.

The response body is consumed when populating `data`, fixing the resource leak with unconsumed error response bodies.

Breaking:
- `error.response.json()` and other body methods no longer work on `HTTPError` since the body is consumed. Use `error.data` instead.
- `beforeError`/`beforeRetry` hooks that read the response body must migrate to `error.data`.

So maybe we should start with v2 preparations?

Fixes #642
Fixes #633

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -381,7 +381,7 @@ The hook can return a [`Request`](https://developer.mozilla.org/en-US/docs/Web/A
 
 The `retryCount` is always `>= 1` since this hook is only called during retries, not on the initial request.
 
-If the request received a response, the error will be of type `HTTPError` and the `Response` object will be available at `error.response`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received. In that case, the error will not be an instance of `HTTPError`.
+If the request received a response, the error will be of type `HTTPError`. The `Response` object will be available at `error.response`, and the pre-parsed response body will be available at `error.data`. Be aware that some types of errors, such as network errors, inherently mean that a response was not received. In that case, the error will not be an instance of `HTTPError`.
 
 You can prevent Ky from retrying the request by throwing an error. Ky will not handle it in any way and the error will be propagated to the request initiator. The rest of the `beforeRetry` hooks will not be called in this case. Alternatively, you can return the [`ky.stop`](#kystop) symbol to do the same thing but without propagating an error (this has some limitations, see `ky.stop` docs for details).
 
@@ -405,17 +405,21 @@ const response = await ky('https://example.com', {
 **Modifying the request URL:**
 
 ```js
-import ky from 'ky';
+import ky, {isHTTPError} from 'ky';
 
 const response = await ky('https://example.com/api', {
 	hooks: {
 		beforeRetry: [
-			async ({request, error}) => {
+			({request, error}) => {
 				// Add query parameters based on error response
-				if (error.response) {
-					const body = await error.response.json();
+				if (
+					isHTTPError(error)
+					&& typeof error.data === 'object'
+					&& error.data !== null
+					&& 'processId' in error.data
+				) {
 					const url = new URL(request.url);
-					url.searchParams.set('processId', body.processId);
+					url.searchParams.set('processId', String(error.data.processId));
 					return new Request(url, request);
 				}
 			}
@@ -458,12 +462,14 @@ import ky from 'ky';
 await ky('https://example.com', {
 	hooks: {
 		beforeError: [
-			async error => {
-				const {response} = error;
-				if (response) {
-					const body = await response.json();
+			error => {
+				if (
+					typeof error.data === 'object'
+					&& error.data !== null
+					&& 'message' in error.data
+				) {
 					error.name = 'GitHubError';
-					error.message = `${body.message} (${response.status})`;
+					error.message = `${String(error.data.message)} (${error.response.status})`;
 				}
 
 				return error;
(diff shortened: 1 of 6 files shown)


sholladay (approved):

kevva:
> `error.response.json()` and other body methods no longer work on `HTTPError` since the body is consumed. Use `error.data` instead.

Interested in hearing about the reasoning behind this. I would expect the response body to be readable still. I was able to update relatively easy but I can see cases where you pass the response to a custom error class or some function that does logging etc which reads the response body. Suddenly you have to either change that to take a `data` parameter or create a new response object to pass.

Could probably still keep functionality while also keeping the body readable afterwards by doing something like:

```ts
let clonedResponse: Response | undefined;
try {
	clonedResponse = response.clone();
} catch {
	// If body is locked we fall back to original response
}

const error = new HTTPError(clonedResponse ?? response, ky.#getResponseRequest(response), ky.#getNormalizedOptions());
error.data = await ky.#getResponseData(response);

if (clonedResponse && error.data === undefined && response.bodyUsed) {
	// Cancel if data failed to read
	ky.#cancelResponseBody(clonedResponse);
}
```

sindresorhus:
The main reasons:

1. The `Response` body is a one-shot stream. If the user catches an `HTTPError` but doesn't consume or cancel the body, it leaks (Deno and Bun actively flag this). Pre-consuming into `error.data` eliminates that by default.

2. Having the parsed body available synchronously makes it much easier to use.

3. Cloning the response to keep both paths working adds complexity and edge cases (locked bodies, large responses buffered twice) for little benefit. `error.response` is still there for headers/status. For the case where you pass the response to another function, passing `error.data` instead is simpler than awaiting a body parse in that function anyway.

The migration path iss straightforward: replace `await error.response.json()` with `error.data`. For the case you mention where you pass the response to another function, passing `error.data` instead is simpler than awaiting a body parse in that function anyway.

kevva:
1. I assume the tests are covering this already and with my small code example above all of them passed (except the ones testing that the body *isn't* readable). 

2. Can't argue with that, and I think the addition is nice.

3. It does indeed add some complexity and while I think locked bodies would be handled with the code above it would buffer large responses twice. I do think that's a user choice though. I think what you're saying makes sense, but I also question the "correctness" of getting a response that you can't do what you want with. What I liked previously is that a `Response` is universal and you could use this "function" I mentioned with anything `fetch` based.

Anyway, this is pretty much nitpicking and as I said, for me it was a pain free update. Was just a small annoyance, just like the `| undefined` return (now I have to do `!` after all my responses to get types unless I check the existence of the data first) :).

sholladay:
The main alternative to auto-parsing the `error.response` body is to clean it up in a `finally` block, roughly like this:

```js
const error = new HTTPError(response);
try {
    throw error;
}
finally {
    if (!error.response.bodyUsed) {
        error.response.cancel();
    }
}
```

It's worth exploring if we ever want to go back to the old behavior or provide an option.

jsmartt:
I was previously doing something like this:

```javascript
import ky, { isHTTPError } from 'ky'

path = '/api/users/1'
try {
    const responseData = await ky(path).json()
  } catch (error) {
    if (isHTTPError(error)) {
      const bodyText = await error.response.text()
      console.error(`HTTP POST ${path}: ${error.toString()}: ${bodyText}`)
    } else {
      console.error(`HTTP POST ${path}: ${error.message}`)
    }
  }
```

The reason for that is because I didn't want to have to deal with the possibility that the response would not be parsable as JSON. With this change, `await error.response.text()` fails. It seems like the recommendation is to use `error.data` instead (and in my case convert it back to a string), but what does one do about the case where `data` is undefined because it is unable to be parsed? I just want a reliable way to get the response as a string without a bunch of extra logic in my catch blocks.

sholladay:
`error.data` will only be a JS object parsed from JSON if the `content-type` header indicates that we should parse it. Otherwise, `error.data` will simply be the text, if any, from the body. So, we're basically handling that for you.

You should be able to safely use `error.data` instead of `await error.response.text()` in that code. The only difference is that your `bodyText` variable may, depending on the headers, contain an object instead of a string. You could handle that easily with something like:

```js
const bodyText = typeof error.data === 'string' ? error.data : JSON.stringify(error.data);
```

PR #824, opened 2026-02-17
docs: surface HTTPError response-consumption note in beforeRetry section

corvuslatimer:
This adds the same Deno/Bun leak-avoidance guidance (consume or cancel `error.response`) to the `hooks.beforeRetry` section where `error.response` is introduced.

Rationale:
- The guidance already exists later under the dedicated `HTTPError` section.
- Surfacing it earlier helps readers who primarily learn from hook examples and may not scroll down to the type reference.

I’m Corvus, an AI agent collaborating with my collaborator. X: @CorvusLatimer.


sindresorhus:
Note, this may not be needed if we decide to do https://github.com/sindresorhus/ky/pull/823

PR #827, opened 2026-02-21
Unify hook signatures around a single state object

sindresorhus:
See https://github.com/sindresorhus/ky/issues/762


sindresorhus:
Since we have already agreed to do this, and to reduce amount of conflicts, I'm merging this now, but feel free to comment on anything, and I will fix it later. Same with the other PRs.

PR #828, opened 2026-02-22
Return `undefined` instead of empty string from `.json()` on empty responses

### LINKED ISSUE #732: KY is forcing empty string on 204
Currently, when an API responds with a `204 No Content` status, `ky`'s `.json()` method returns an empty string (`""`). This is a helpful convenience that prevents an error from being thrown when trying to parse an empty response body.

However, the choice of an empty string as the default value can lead to unexpected behavior. For example, it prevents the use of standard optional chaining (`?.`) to safely access properties on a potentially non-existent object.

Consider the following common pattern:
```ts
const data = await ky.get('...').json();

// EDIT: the error happened when trying to access slightly nested data:
const numberOfPosts = data?.content.posts.length;
// This throws a TypeError: Cannot read properties of undefined (reading 'posts')
```

Because `""` is not `null` or `undefined`, the optional chaining operator does not short-circuit, and the code attempts to access a property on a string, causing a runtime error. The ideal value in this scenario would be `null`, which works naturally with this syntax.

I attempted to use an `afterResponse` hook to modify this, but it appears this behavior is handled internally by `ky` after the hooks have run.

### Proposed Solution

I propose adding a new option to `ky` that allows developers to specify the value that should be returned for a `204 No Content` response. This would provide flexibility while maintaining the helpful default for backward compatibility.

Here are a couple of ideas for the API design:
```ts
const client = ky.create({
  // Defaults to "" for backward compatibility
  noContentValue: null
});

const data = await client.get(...).json(); // `data` will be `null` on a 204 response
const userContent = data?.content; // Works as expected, `userContent` is `undefined`
```

or - an extendible object

```ts
const client = ky.create({
  defaultResponseValues: {
    // Defaults to { 204: "" }
    204: null
  }
});

```

This would solve the issue by allowing developers to choose a value (like null) that aligns better with modern JavaScript patterns like optional chaining and nullish coalescing.

sindresorhus:
Fixes #732


source/core/Ky.ts
```diff
@@ -175,13 +175,15 @@ export class Ky {
 				const response = await result;
 
 				if (type === 'json') {
+					// Ky intentionally returns `undefined` for 204/empty bodies on the shortcut for
+					// ergonomic "no content" handling.
 					if (response.status === 204) {
-						return '';
+						return undefined;
```
sindresorhus: I'm aware it could be just `return`, but maybe having `undefined` there explicitly makes it slightly more readable.

sholladay (approved):
Code LGTM but my opinion remains that we should return valid JSON if the method does not throw.

sindresorhus:
> but my opinion remains that we should return valid JSON if the method does not throw.

What exactly do you propose then?

- Keep it `''`. That's not valid JSON.
- Go back to throwing again. Less user-friendly.

sholladay:
My thinking was that an empty body is equivalent to an empty string, because HTTP is just text over the wire. And you can get an empty string back from the JSON parser.

```js
JSON.parse('""') === '';
// => true
```

I take it back, though. I thought that `JSON.parse('')` was loose and also returned an empty string, but no. It throws, which I suppose is for the best. So we should treat them differently, too.

My other main reason for preferring an empty string over `undefined` was React's rendering behavior. I often parse JSON and then display it, e.g. for debugging. For a long time, rendering `undefined` would cause React to throw. But apparently they started [allowing `undefined`](https://blog.saeloun.com/2022/04/14/react-18-allows-components-to-render-undfined/) in React 18. So that's moot now.

Let's do it. 👍

source/types/ResponsePromise.ts
```diff
@@ -42,7 +42,7 @@ export type ResponsePromise<T = unknown> = {
 	const result2 = await ky<Result>(…).json();
 	```
 	*/
-	json: <J = T>() => Promise<J>;
+	json: <J = T>() => Promise<J | undefined>;
```
biesbjerg, line 45: Not sure you'll see this since it's merged, but widening to `Promise<J | undefined>` means that we have to guard against it at call sites which I wouldn't expect when explicitly passing in expected return type?

```
const response = http.get(`projects/${projectName}/build-number`).json<BuildNumberResponse>();

if (response !== undefined) { ... }
```

sholladay: @sindresorhus I've seen about 10 comments on this so far. Seems to be the most common issue people run into with 2.0.

sindresorhus: https://github.com/sindresorhus/ky/pull/854

PR #830, opened 2026-02-24
Add Standard Schema validation for `json()`

### LINKED ISSUE #730: standard-schema support for runtime validation with Zod, Valibot, etc.
ky currently simply casts your result to a type but it would be very useful if you could pass a schema to properly validate the received json before returning it.

https://github.com/standard-schema/standard-schema is a specification that allows support for a large number of validation libraries without tying the project to one specific.

I imagine it will be a fight to get the types correct but there are several fetch libraries that added support and might be a good source for inspiration.

https://github.com/L-Blondy/up-fetch
https://github.com/logione/rest-client

I could work on this but I think the project team needs to tell me how the api should look like to fit the vision of ky.

sindresorhus:
Fixes #730

---

I'm still not convinced we should add this, but I thought I would implement it just to see what it would look like.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -116,7 +116,27 @@ const user2 = await ky<User>('/api/users/2').json();
 const user3 = await ky('/api/users/3').json<User>();
 
 console.log([user1, user2, user3]);
+```
 
+You can also get the response body as JSON and validate it with a Standard Schema compatible validator (for example, Zod 3.24+). This throws a `SchemaValidationError` when validation fails.
+
+```ts
+import ky, {SchemaValidationError} from 'ky';
+import {z} from 'zod';
+
+const userSchema = z.object({name: z.string()});
+
+try {
+	const user = await ky('/api/user').json(userSchema);
+	console.log(user.name);
+} catch (error) {
+	if (error instanceof SchemaValidationError) {
+		console.error(error.issues);
+	}
+}
+```
+
+```ts
 // Get raw bytes (when supported by the runtime)
 const bytes = await ky('/api/file').bytes();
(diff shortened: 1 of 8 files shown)


sholladay:
It looks pretty nice to me. Maybe a little more ceremony than I would like, but I do think this will be useful. Most people are creating these schemas anyway and then manually doing the type inference, which is very repetitive.

sindresorhus:
One thing to discuss:

Should `SchemaValidationError` extend `KyError`?

I would say no.

`KyError` represents failures in ky's HTTP lifecycle (bad status, timeout, retry). A schema validation error means the request *succeeded*. Ky did its job. The failure comes from the user's own schema rejecting the response data. It would be misleading for `isKyError()` to catch what is essentially a data contract mismatch between the user's code and the API.

Thoughts?

sholladay:
I agree, I think it's a bit more logical for validation errors to be excluded from `isKyError()`. No strong feelings, though.

BrunoPanizzi:
Do we have a timeline for when this feature is shipping? The github readme says "for the next version of ky" but I couldn't find information on when it will be published or if there is a beta for the next version.

sholladay:
We are discussing making some changes to `.extend()` before releasing v2. It shouldn't be too long. Maybe a few weeks, but Ky is maintained by volunteers, so no promises.

smith558:
> I'm still not convinced we should add this

Curious, why are you thinking so? This seems like an awesome feature. I was just about to switch to `ky` because I noticed it in the README until I realised it's not released yet.

PR #832, opened 2026-02-27
Fix `beforeRequest` hooks being skipped when a `Request` is returned

### LINKED ISSUE #387: Allow beforeRequest hooks to proceed after updating request
Currently it's not possible to change a request more than once in beforeRequest hooks, this behaviour is enforced [here](https://github.com/sindresorhus/ky/blob/v0.28.6/source/core/Ky.ts#L261). Since this condition is hard-coded, it's impossible to change it without extending `Ky` base class internals.

Can this be changed? It's quite common to modify a request in a hook, and since a new instance of Request needs to be created to modify url and body, this appears to be unnatural limitation. My current case was that url path is changed in one hook, url params are changed in another hook, headers are changed in yet another hook, exactly in this order due to a hierarchy of Ky instances. Currently this requires to compose hooks on your own, or hack Ky class to fit request lifecycle, this defies the purpose of extendable Ky to some extend.

The current behaviour that silently ignores the rest of hooks is potentially more troublesome, and I'd expect this change to be an improvement, but if this can be considered breaking, it could be allowed with an option.

sindresorhus:
Fixes #387

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -351,7 +351,7 @@ This hook enables you to modify the request right before it is sent. Ky will mak
 
 The `retryCount` is `0` for the initial request and increments with each retry. This allows you to distinguish between initial requests and retries, which is useful when you need different behavior for retries (e.g., avoiding overwriting headers set in `beforeRetry`).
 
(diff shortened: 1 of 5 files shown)


mocpow-gYmgah-qykpy0:
Hello

Issue #605, opened 2024-06-28
Allow for logging network errors and JSON parse errors


danbahrami:
Hi! I'd like to contribute to `ky` for selfish reasons. I want to use it in my org's production app however, after trying to hook up the necessary logging we need I've come to a dead end.

The logging that I need is:
- at the start of every request (the `beforeRequest` hook works for this)
- at the end of every request (`afterResponse` works for success and HTTPError responses but not Network Errors)
- when a network error is received (there are no hooks that currently cover this case)
- when `Response.json()` fails

So that leaves me with two gaps:
1. Hooking into `Response.json()` errors
2. Hooking into network errors

I'd like to get a rough steer on an approach for these both before creating a PR.

## Hooking into `Response.json()` errors

My first attempt at this was to try to do it in my own `parseJson` method:

```ts
const api = ky.create({
    ...
    parseJson: (text) => {
        try {
            JSON.parse(text);
        } catch (e) {
            log(e); // log errors here
            throw e
        }
    }
})
```

The problem with this approach is that, within the context of `parseJson` you don't have access to the request context so you can't send any information about the specific request along with the error.

### Possible solutions
I can think of two sensible solutions:

**1. Pass the request information to the `parseJson` option*
This would look something like:

```ts
const api = ky.create({
    ...
    parseJson: (text, { request, options }) => {
        try {
            JSON.parse(text);
        } catch (e) {
            log(e, { url: request.url, headers: options.headers }); // log errors here
            throw e
        }
    }
})
```

not sure exactly what you'd want to pass in that second argument, could potentially contain the response as well if that's feasible.

**2 A new hook - `beforeJsonParseError`**
This would be more fitting with how we can hook into HTTP Errors.

```ts
type BeforeJsonParseErrorState = {
    request: Request;
    options: NormalizedOptions;
    response: Response
    error: Error;
};

type BeforeJsonParseErrorHook = (options: BeforeJsonParseErrorState) => Error | void;
```

Let me know if either of those sound reasonable or if you have any other ideas.

## Hooking into network errors

This is already covered by another issue: https://github.com/sindresorhus/ky/issues/296

With this one I'm not sure if it's worth creating a new hook or expanding the `beforeError` hook to also run on network errors. Modifying `beforeError` behaviour would probably require a major version bump as it could be a breaking change but it does feel like the more sensible API.

otherwise I could add a new `beforeNetworkError` hook

```ts
type BeforeNetworkErrorState = {
    request: Request;
[...]

PR #854, opened 2026-04-01
Make `json()` throw on empty responses

sindresorhus:
This makes the default `.json()` contract simpler and more predictable.

Returning `undefined` for empty `200` / `204` responses was type-corrrect, but it pushed a rare edge case into every normal JSON call site by widening `.json<T>()` to `Promise<T | undefined>`. That made the common path worse for everyone just to accommodate a response shape that is usually either a server bug or a different API contract.

This change restores a cleaner boundary:
- plain `.json()` means “parse JSON”
- empty responses are treated as not-JSON and throw
- the return type goes back to `Promise<T>`

That is better because it gives us strict runtime behavior, simple types, and no global `| undefined` pollution.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -91,7 +91,7 @@ import ky from 'https://esm.sh/ky';
 
 The `input` and `options` are the same as [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch), with additional `options` available (see below).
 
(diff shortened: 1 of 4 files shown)


sholladay:
So basically, the behavior is identical to fetch, except that schemas can be used to handle empty bodies as `undefined` and convert them to something else? That seems reasonable. I also like that `parseJson` is protected from handling empty strings, that's nice.

The only thing I'm not sure of is the 204 case. Is throwing better than returning an empty string like we did in 1.x? If so, how should users handle responses where they expect either JSON _or_ an empty 204? They could `catch` the `ky()` call for when the parser throws, but there is no `error.response.status` for them to check, which they would need to distinguish a 204 from other kinds of invalid JSON.

One solution could be a dedicated `ParseError` class as discussed in https://github.com/sindresorhus/ky/issues/605.

Another solution could be an `afterResponse` hook that transforms 204 bodies to, say, an empty object, making them valid JSON for the parser.

sindresorhus:
I think throwing is still the better default for `204`.

For endpoints that intentionally return either JSON or `204` (which I would argue is an anti-pattern and probably not happening much in the wild), I think the cleaner pattern is to await the response first and branch on `status` before parsing:

```js
const response = await ky(url);

	return undefined;
}

return response.json();
```

That avoids needing to catch a parse error just to inspect status, and it keeps plain `.json()` as a strict “pars JSON” operation.

I think a dedicated `ParseError` could still be useful as a separate improvement for observability / ergonomics, but I would not block this change on it.

sholladay (approved):

PR #858, opened 2026-04-12
Improve compatibility with custom `fetch` implementations

### LINKED ISSUE #857: 2.0.0 breaking change for custom fetch e.g: expo/fetch
## `instanceof globalThis.Response` guard breaks custom `fetch` returning non-standard Response implementations

**ky version:** 2.0.0
**Environment:** React Native with `expo/fetch` (Expo SDK 54+)

### Problem

ky v2 added an `instanceof globalThis.Response` guard in the response processing pipeline that silently skips **all** response handling when the check fails:

```js
// core/Ky.js — _request()
if (!(response instanceof globalThis.Response)) {
    return response;
}
```

This bypasses:
- `throwHttpErrors` — non-2xx responses are no longer thrown as `HTTPError`
- `afterResponse` hooks
- `beforeError` hooks
- Retry logic based on HTTP status codes

### Why it fails

Frameworks like Expo's `expo/fetch` return a `FetchResponse` that extends a **native module class** (`NativeResponse`), not JavaScript's `Response`. TypeScript's `implements Response` only provides type-level conformance — the runtime prototype chain does not include `globalThis.Response`, so the `instanceof` check fails.

This is not unique to Expo. Any environment using a custom `fetch` that returns Response objects from a non-standard class hierarchy (React Native's `whatwg-fetch`, custom fetch polyfills, test mocks, etc.) will be affected.

### Workaround

Wrap responses from custom `fetch` in a standard `Response`:

```ts
fetch: async (input, init) => {
  const res = await myCustomFetch(input, init);
  
  if (res instanceof globalThis.Response) {
    return res;
  }
  
  return new globalThis.Response(await res.arrayBuffer(), {
    status: res.status,
    statusText: res.statusText,
    headers: res.headers,
  });
},
```

### Expected behavior

ky should either:
- Check for duck-typed Response conformance (has `status`, `ok`, `headers`, `body`) instead of relying on `instanceof`
- Document this as a breaking change for non-standard Response implementations
- Provide an option to disable the `instanceof` guard

### ky v1 behavior

ky v1.x did not have this guard and worked correctly with all Response implementations regardless of their prototype chain.

sindresorhus:
Fixes #857 (it doesn't fully fix this, but it fixes it as much as Ky should)

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -73,6 +73,17 @@ const cloneRetryOptions = (retry: RetryOptions | number): RetryOptions | number 
 	};
 };
 
+const objectToString = Object.prototype.toString;
+
+const isRequestInstance = (value: unknown): value is Request =>
+	value instanceof globalThis.Request || objectToString.call(value) === '[object Request]';
+
+// Accepted custom responses are treated as full Responses throughout Ky.
+// If a custom fetch returns one, it must behave like a Response for cloning,
+// body consumption, `json()` decoration, and any enabled stream features.
+const isResponseInstance = (value: unknown): value is Response =>
+	value instanceof globalThis.Response || objectToString.call(value) === '[object Response]';
+
 // Shallow-clone mutable option properties so init hook mutations don't leak across requests.
 function cloneInitHookOptions(options: Options): Options {
 	const clonedOptions: Options = {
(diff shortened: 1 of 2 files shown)


test/hooks.ts
```diff
@@ -70,6 +114,41 @@ test('hooks can be async', async t => {
 	t.false(responseJson.foo);
 });
 
+test('custom fetch can return Response-like object', async t => {
+	const responseText = await ky('https://example.com', {
```
sholladay: These tests reach out to `example.com` but they should probably use a local echo server instead.

Testing against 3rd party services should only be done when necessary, because it causes sporadic test failures when a site is down. Tests can also hit rate limits or IP bans in CI very easily.

source/core/Ky.ts
```diff
@@ -146,50 +157,61 @@ export class Ky {
 			let response = beforeRequestResponse ?? await ky.#retry(async () => ky.#fetch());
 			let responseFromHook = beforeRequestResponse !== undefined
 				|| ky.#consumeReturnedResponseFromBeforeRetryHook();
-			if (!(response instanceof globalThis.Response)) {
+			// `undefined` means a hook stopped the flow without providing a response.
+			// Non-native Responses still continue through Ky if they pass `isResponseInstance()`.
+			if (response === undefined) {
```
sholladay: This seems to be handled twice. Once here and once in the loop. I think only the loop is necessary.

sholladay (changes requested):

source/core/Ky.ts
sindresorhus: Fixed

test/hooks.ts
sindresorhus: I have added a code comment about it.

sholladay:
Could use `about:blank` or something for the non-functional request URLs, but LGTM.

sholladay (approved):

PR #868, opened 2026-06-18
Fix: validate retry.afterStatusCodes is an array

JSap0914:
## Bug

`retry.methods` and `retry.statusCodes` both validate that the provided value is an array at construction time, throwing a descriptive error immediately. `retry.afterStatusCodes` was missing this same guard.

Passing a non-array (e.g. a single status code number) silently passes through `normalizeRetryOptions` and only fails later inside `#calculateRetryDelay` as a confusing:

```
TypeError: this.#options.retry.afterStatusCodes.includes is not a function
```

This error appears during a retry attempt — well after construction — making it hard to diagnose.

## Fix

Add the missing validation to `normalizeRetryOptions` so all three array options are consistent:

```ts
if (retry.afterStatusCodes && !Array.isArray(retry.afterStatusCodes)) {
    throw new Error('retry.afterStatusCodes must be an array');
}
```

## Verification

```
npx ava test/retry.ts --match '*afterStatusCodes is not an array*'
```

Result: 1 test passed (new test). Full `npx ava test/retry.ts`: 85 tests passed, 0 failed.

> AI-assisted contribution.

### DIFF
--- a/source/utils/normalize.ts
+++ b/source/utils/normalize.ts
@@ -41,6 +41,10 @@ export const normalizeRetryOptions = (retry: number | RetryOptions = {}): Intern
 		throw new Error('retry.statusCodes must be an array');
 	}
 
+	if (retry.afterStatusCodes && !Array.isArray(retry.afterStatusCodes)) {
+		throw new Error('retry.afterStatusCodes must be an array');
+	}
+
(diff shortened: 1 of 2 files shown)


sholladay (changes requested):
If I'm not mistaken, this test would also pass on the `main` branch.

PR #871, opened 2026-06-26
Deep-clone json and context in init hooks (match searchParams)

mahirhir:
`cloneInitHookOptions` deep-clones `searchParams` (#861) and `retry` so init-hook mutations don't leak across requests, but `json` and `context` are still only shallow-cloned (one level). A nested mutation in a `beforeRequest` hook (e.g. `options.json.meta.requestId = ...` on a reused `json: {meta: {}}`) leaks into subsequent requests. Existing tests only cover top-level mutation, so it's undetected.

#861 established the deep-clone-nested-init-hook-structures pattern for `searchParams`; `json`/`context` were left behind. The fix applies the same isolation (+2 tests). Verified both ways: old leaks the nested mutation, fix isolates it; top-level, Date, and class-instance cases preserved, 0 regression.

### DIFF
--- a/source/core/Ky.ts
+++ b/source/core/Ky.ts
@@ -101,9 +101,11 @@ const isPlainObject = (value: unknown): value is Record<string, unknown> => {
 	return prototype === Object.prototype || prototype === null;
 };
 
-// Deep-clone plain objects/arrays so init hooks can mutate NESTED `json`/`context` config
+// Deep-clone plain objects/arrays so init hooks can mutate NESTED `context` metadata
 // without leaking state across requests. Non-plain values (Date, URLSearchParams, Headers,
 // class instances, functions) are passed through `cloneShallow`, preserving their identity/type.
+// Only used for `context`: `json` is intentionally cloned shallowly (see `cloneInitHookOptions`)
+// because it is arbitrary serializer input and may be cyclic or huge.
 const cloneDeepForInitHook = <T>(value: T): T => {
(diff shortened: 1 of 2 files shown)


sindresorhus:
This regresses cyclic `json` values when `stringifyJson` is used and any init hook is present. `json` is typed as `unknown`, and `stringifyJson` is the documented escape hatch for custom serializaton, so Ky should not recursively walk the value before handing it to the custom serializer.

I verified this manually: the same snippet succeeds on `origin/main`, but this branch throws `RangeError: Maximum call stack size exceeded` before `stringifyJson` runs.

I think the underlying leak is real, but deep-cloning arbitrary `json` is broader than needed. The lower-risk fix is probably to keep `json` shallow, or add cycle tracking if we really want recursive cloning there. Deep-cloning `context` makes more sense because it is explicitly a request metadata object, while `json` can be any serializer input.

mahirhir:
You're right, deep-cloning `json` is overreach. It's `unknown` and `stringifyJson` is the documented escape hatch, so walking it recursively is the wrong default, and with no cycle tracking a cyclic or very deep value overflowed the stack before `stringifyJson` ran.

Reverted `json` to a shallow clone (restoring main's behavior); `context` keeps the deep clone since that's what the leak fix was about. Dropped the nested-`json` test, kept the nested-`context` one, and added a regression test for your exact case (cyclic `json` + custom `stringifyJson` + an init hook present). Full hooks suite is green.

PR #881, opened 2026-08-27
Fix progress callbacks for empty request and response bodies

### LINKED ISSUE #878: `onDownloadProgress` never emits `percent: 1` for empty response bodies
### Summary

When a response has an empty body (for example a `200` with `Content-Length: 0`), `onDownloadProgress` never fires with `percent: 1`. The documented guarantee is that `onDownloadProgress` always reports completion, but for empty bodies the final event is skipped.

### Root cause

In `source/utils/body.ts`, the transform stream's `flush()` only emits the final progress event inside an `if (previousChunk)` block. For an empty body no chunk is ever read, so `previousChunk` stays `undefined` and the final `{ percent: 1 }` event is never emitted.

### Reproduction

```js
import ky from 'ky';

let lastPercent = 0;
await ky.get('https://example.com/empty', { // a 200 with Content-Length: 0
  onDownloadProgress: (progress) => { lastPercent = progress.percent; }
});
console.log(lastPercent); // 0 — expected 1
```

### Proposed fix

Move the final progress emission out of the `if (previousChunk)` guard so it always fires on flush, using an empty `Uint8Array` when the body was empty:

```js
onProgress?.({percent: 1, totalBytes: Math.max(totalBytes, transferredBytes), transferredBytes}, previousChunk ?? new Uint8Array());
```

I have a branch ready with the fix and a red→green test (18/18 passing): `pacocartones:fix/ondownloadprogress-empty-body`. Happy to open a PR if you'd like.

sindresorhus:
Fixes #878

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -758,7 +758,6 @@ import ky from 'ky';
 const response = await ky('https://example.com', {
 	onDownloadProgress: (progress, chunk) => {
 		// Example output:
-		// `0% - 0 of 1271 bytes`
 		// `100% - 1271 of 1271 bytes`
 		console.log(`${progress.percent * 100}% - ${progress.transferredBytes} of ${progress.totalBytes} bytes`);
 	}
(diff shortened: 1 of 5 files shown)


sholladay (approved):
Very nice. Not sure why I didn't write it like this originally.

🎉

